PR #22, opened 2015-03-17
DROOLS-742 DecisionTableXLSToDecisionTableGuidedConverterTest fix.

baldimir:
See JIRA: https://issues.jboss.org/browse/DROOLS-742


manstis:
Hello,

Could you please re-instate the test that checks the Model is correct after conversion? Whilst your changes for the "conditions" tests validate the data; they no longer validate the resulting Model classes. I think this is important.

Also, it'd be nice if you could use the code-style formatting we try to adhere to: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/tree/master/ide-configuration.

Thanks,

Mike

baldimir:
Hi, I made formatting changes. But it looks, that the formatter config which you linked is not actual. I use Idea 14 and i.e. the spaces after and before parenthesis are not there. 

Which test do you mean that I should reinstate? I didn't erase any test. I simply joined two tests that were the same . The only difference was in XLS file used and data rows that were checked at the end of the test. Everything that was tested there is still in the tests.

manstis:
http://github.com/droolsjbpm/drools-wb/commit/dc97bf1bf

manstis:
Sorry, my mistake. I looked at the code after your change and the test is fine.

Commits squashed and merged into master.

Thanks for the PR.

PR #23, opened 2015-03-20
BZ-1203696: fix for contract change

porcelli:


manstis:
http://github.com/droolsjbpm/drools-wb/commit/397cda7db

PR #25, opened 2015-04-24
missing configs of GWT 2.7 upgrade

porcelli:


manstis:
http://github.com/droolsjbpm/drools-wb/commit/6d6023b14

PR #26, opened 2015-04-27
cleanup of webapp dirs

porcelli:


drools-wb-webapp/pom.xml
```diff
@@ -932,7 +932,12 @@
             <fileset>
               <directory>${basedir}</directory>
               <includes>
-                <include>.errai/**</include>
+                <include>src/main/webapp/org.drools.workbench.DroolsWorkbench/</include>
+                <include>src/main/webapp/WEB-INF/deploy/</include>
+                <include>src/main/webapp/WEB-INF/classes/</include>
+                <include>src/main/webapp/WEB-INF/lib/</include>
+                <include>**/gwt-unitCache/**</include>
+                <include>.errai/</include>
```
psiroky: I am most probably out of context here, but do we need to create this stuff under src/main/webapp? Would it be maybe possible to copy the stuff from src/main/webapp to target/ and then fiddle with that?

porcelli: we may try it later... but this has been the way to do this promoted by errai. at this point I'm looking for a working solution that later can be improved.

psiroky: Ah, OK, I wasn't aware of that. I agree with making it work like this and possibly improve later.

manstis:
http://github.com/droolsjbpm/drools-wb/commit/368efeeb8

PR #29, opened 2015-05-12

wmedvede:


drools-wb-webapp/pom.xml
```diff
@@ -792,7 +797,9 @@
         <configuration>
           <deploy>${project.build.directory}/gwt-symbols-deploy</deploy>
           <localWorkers>1</localWorkers>
-          <extraJvmArgs>-Xmx2048m -XX:MaxPermSize=256m -Xms1024m -XX:PermSize=128m -Xss1M -XX:CompileThreshold=7000 -Derrai.jboss.home=${errai.jboss.home}</extraJvmArgs>
+          <extraJvmArgs>-Xmx2048m -XX:MaxPermSize=256m -Xms1024m -XX:PermSize=128m -Xss1M -XX:CompileThreshold=7000 -Derrai.jboss.home=/home/wmedvede/development/projects/droolsjbpm/drools-wb/drools-wb-webapp/target/wildfly-8.1.0.Final</extraJvmArgs>
```
manstis: Tssk tssk.. I don't want to create this path on my laptop ;)

manstis: I've replaced with ${errai.jboss.home} in the merge.

manstis:
http://github.com/droolsjbpm/drools-wb/commit/98f8e709f

drools-wb-webapp/pom.xml
Rikkola: I actually did the same change. Based on Porcelli's instructions.
But once I got everything working and tried with the variable. There actually is no need to change this line in order to get the super dev mode working. It is enough to set this variable in the Idea's run configurations.

PR #32, opened 2015-05-26
GUVNOR-2188: Converting XLS Decision Tables fails a row doesn't contain all necessary cells (i.e. POI reports a cell as non-existent)

manstis:
This fix ensures all columns have the same number of rows when converting.

"Empty" cells are caught early in the ExcelParser and not passed to the conversion code. The problem arises when POI "thinks" some cells contain content (i.e. the user has simply deleted the content rather than deleted the cell).


Rikkola:
https://github.com/droolsjbpm/drools-wb/commit/6a9ef2f1e

One thing about the code. This ends up being repeated a lot.
```
if ( this.values.size() < maxRowCount ) {
 for ( int iRow = this.values.size(); iRow < maxRowCount; iRow++ ) {
 this.values.add( new DTCellValue52( "" ) );
 }
}
```
We could make Column and Row classes that extend ArrayList<DTCellValue52> and add this functionality into that class/those classes. It is also easy to see what the list is about once it is named. I don't think we need to do this now, probably better to do it while working with the dtable UI task.

manstis:
@Rikkola The problem with this bug was that the Lists did not have enough elements in the first place; so extending List for row and column would not have solved this issue; as we'd still need code to add the missing elements.

Rikkola:
Sure, the comment was not for how to do the fix. It was more for how the code would have less redundancy and would be cleaner. The functionality that fills the Column/Row class would be inside the classes.

Rikkola:
@manstis So the class would be something like this. It would auto fill the Column and probably have some other methods that execute changes on the actual list.
```
class Column extend ArrayList<DTCellValue52> {
  
   public Column(int size, Object defaultValue){
   super(size);
     if ( this.size() < maxRowCount ) {
       for ( int iRow = this.size(); iRow < maxRowCount; iRow++ ) {
         this.add( new DTCellValue52( defaultValue ) );
       }
     }
   }

}
```

Just throwing out ideas :)

PR #33, opened 2015-05-28
DROOLS-798: Guided Score Card: Issues re-loading previously saved files

manstis:
See https://issues.jboss.org/browse/DROOLS-798


Rikkola:
https://github.com/droolsjbpm/drools-wb/commit/81059bdae

PR #34, opened 2015-06-02
GUVNOR-2187 : Real time verification and validation for decision tables

Rikkola:
Depends on this PR https://github.com/droolsjbpm/kie-wb-common/pull/36
And this: droolsjbpm/drools#445
Documentation is here. droolsjbpm/kie-wb-distributions#58

The goal of this commit is to get the real time feature in. The validation and verification checks are still more or less work in progress. Redundancy, subsumption and conflict checks do work, but they all still have blind spots, with enums for example. 

Issues that I spotted while looking at the changes myself:
* Some check classes have commented out code that I need to revisit.
* Numeric*ConditionInspector will most likely be killed an all the numbers will use NumericConditionInspector.
* EnumConditionInspector needs to be made to work. 
None of the missing bits cause the V&V to report false negatives so they will just not make it for milestone1.


drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-backend/src/main/java/org/drools/workbench/screens/guided/rule/backend/server/GuidedRuleEditorServiceImpl.java
```diff
@@ -161,6 +161,7 @@ public RuleModel load( final Path path ) {
             return ruleModel;
 
         } catch ( Exception e ) {
+            e.printStackTrace();
```
Rikkola: AHA! This should not be here

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/resources/rebel.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<application xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.zeroturnaround.com" xsi:schemaLocation="http://www.zeroturnaround.com http://www.zeroturnaround.com/alderaan/rebel-2_0.xsd">
+
+	<classpath>
+		<dir name="/home/rikkola/Work/Codes/master/drools-wb/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/target/classes">
```
yurloc: Absolute path. Could it be avoided by using `${project.build.outputDirectory}` property coming from [Super POM](http://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Super_POM) and enabling [resource filtering](http://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html)?

manstis:
@yurloc This PR wasn't merged. https://github.com/droolsjbpm/drools-wb/pull/35 replaced it (and did not contain the JRebel configuration).

PR #35, opened 2015-06-02
GUVNOR-2187 : Real time verification and validation for decision tables

Rikkola:
Depends on this PR https://github.com/droolsjbpm/kie-wb-common/pull/36
And this: droolsjbpm/drools#445
Documentation is here. droolsjbpm/kie-wb-distributions#58

The goal of this commit is to get the real time feature in. The validation and verification checks are still more or less work in progress. Redundancy, subsumption and conflict checks do work, but they all still have blind spots, with enums for example. 

Issues that I spotted while looking at the changes myself:
* Some check classes have commented out code that I need to revisit.
* Numeric*ConditionInspector will most likely be killed an all the numbers will use NumericConditionInspector.
* EnumConditionInspector needs to be made to work. 
None of the missing bits cause the V&V to report false negatives so they will just not make it for milestone1.


manstis:
http://github.com/droolsjbpm/drools-wb/commit/07c293f85

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/DecisionTableAnalyzer.java
```diff
@@ -16,226 +16,224 @@
...
+                   UpdateColumnDataEvent.Handler,
+                   AppendRowEvent.Handler,
+                   InsertRowEvent.Handler,
+                   AfterColumnInserted.Handler {
+
+    private final RowInspectorCache cache;
+    private final GuidedDecisionTable52 model;
+    private final EventBus eventBus;
+    private final Checks checks = new Checks();
+    private final EvenManager eventManager = new EvenManager();
```
manstis: Probably want to call this Even*t*Manager

Rikkola: Good poin.

PR #36, opened 2015-06-03
GUVNOR-2190 : Ace editor DRL syntax hilighting for the source tab.

Rikkola:
https://github.com/droolsjbpm/kie-wb-common/pull/37


manstis:
http://github.com/droolsjbpm/drools-wb/commit/6c8df52d9

PR #37, opened 2015-06-04

wmedvede:


drools-wb-webapp/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -0,0 +1,27 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<jboss-deployment-structure xmlns="urn:jboss:deployment-structure:1.0">
+  <deployment>
+    <dependencies>
+
+      <!-- add JAXB processing -->
+      <module name="com.sun.xml.bind"/>
```
psiroky: `com.sun.xml.bind` is a EAP/WildFly private module and the application should not, in _ideal_ case, depend on those. It would be better to bundle this dependency directly into WEB-INF/lib (unless there is a other reason not to, which I am not aware of). In this case, it is not that important as we don't support `drools-wb-webapp` for user deployments.

manstis:
@wmedvede I'll have to defer to @psiroky's knowledge about EAP/Wildfly modules. Whilst ```drools-wb-webapp``` isn't a supported deployment; I assume we'd need something similar to ```kie-wb-distributions```?

drools-wb-webapp/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
wmedvede: this is a showcase that we always run with Widlfy

wmedvede:
Since JAXB is also used in other xml parsings around the platform (e.g. jbpm runtime, rest services) I can guess the different different distributions has already the proper configuration

PR #38, opened 2015-06-15
GUVNOR-2192 : Guided Decision Table V&V - Finding Deficiency

Rikkola:
Deficiency checks.
Several bug fixes with improved test coverage.

Added one to many check type. For situations where an issue on a single row depends on all the other rows. Not sure if this will be needed for anything else than defiency, but making the check allows the Checks class to automatically update the checks as cell values change.


manstis:
http://github.com/droolsjbpm/drools-wb/commit/71e104d15

PR #39, opened 2015-06-16
GUVNOR-2195 - Guided Decision Table V&V -Redundant Conditions/Actions

Rikkola:
I merged InsertFactActionInspectorKey and SetFieldColActionInspectorKey. Could not think of a good enough reason to have the info about inserting/editing in the key. Better to just have a simple key and then the information is stored in the actual ConditionInspector.

Again, threw in some bug fixes with unit tests. Conflicts between rows were a bit too eager to report conflicts. If the StringConditionInspector has value set to "" we do not generate source for it. The checks were picking up that the "" value is not conflicting with what ever was set for the field in another column.


manstis:
http://github.com/droolsjbpm/drools-wb/commit/a06401fbf

PR #43, opened 2015-06-25

wmedvede:
Small changes due to some house cleaning at kie-wb-common


manstis:
http://github.com/droolsjbpm/drools-wb/commit/930652c9f

PR #44, opened 2015-07-02
Fix M2RepoPerspective for using new UploadForm

Hasys:
Addition for droolsjbpm/guvnor#135 commit. Have to be merged together.


manstis:
http://github.com/droolsjbpm/drools-wb/commit/ec3ee27c8

PR #45, opened 2015-07-06
GUVNOR-2205: Syntax highlighting for WID and DSL

manstis:
See https://issues.jboss.org/browse/GUVNOR-2205


manstis:
http://github.com/droolsjbpm/drools-wb/commit/a109166ae

PR #46, opened 2015-07-15
GUVNOR-2194 : Guided Decision Table V&V - Improved Reporting

Rikkola:
* Removed the analysis column from the dtable and added an analysis panel that shows up on the right side of the workbench if the V&V picks up any issues
* Removed any TODO code that was in. I'll add them back in when we actually need to figure out ranges. The goal is to keep master as clean as it can be
* Removed the condition inspectors classes for each type expect integer, integer is special
* Made a lot of method parameters final. @manstis likes finals. I'm starting to like them too now that they have saved me from making bugs

This PR depends on this https://github.com/droolsjbpm/drools/pull/460

pr


manstis:
http://github.com/droolsjbpm/drools-wb/commit/28de70536

PR #47, opened 2015-07-20
GUVNOR-2176 : Guided decision table verification and validation

Rikkola:
More bug fixes. Few that affected enums in the dtable. I also added a way to add tests against gdst files.


manstis:
http://github.com/droolsjbpm/drools-wb/commit/50b4d3c24

PR #49, opened 2015-07-22
BZ1241414: Ability to set default locale of Business Central

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1241414


porcelli:
https://github.com/droolsjbpm/drools-wb/commit/25503d511

PR #50, opened 2015-07-22
BPMSPL-100: removed categories editor

pefernan:


Rikkola:
http://github.com/droolsjbpm/drools-wb/commit/3647320de

PR #51, opened 2015-07-24
GUVNOR-2226: Wires: Fixes following upgrade of Lienzo version

manstis:


manstis:
http://github.com/droolsjbpm/drools-wb/commit/1caba8613

PR #52, opened 2015-07-24
GUVNOR-2225: Wires: Replace overloaded Layer.draw() with calls to Layer.batch()

manstis:
See https://issues.jboss.org/browse/GUVNOR-2225


manstis:
http://github.com/droolsjbpm/drools-wb/commit/d5bb3192d

PR #57, opened 2015-08-21
DROOLS-684 : Drools Work bench not allowing in .XLSX to upload

Rikkola:


manstis:
(master) http://github.com/droolsjbpm/drools-wb/commit/658705911

(6.3.x) http://github.com/droolsjbpm/drools-wb/commit/5be636bc3

PR #59, opened 2015-09-02
BZ1256623: Not possible to convert XLS table to GDT with blank cells

manstis:
Test for captioned BZ.. the test passes (due to a previous fix for GUVNOR-2188).


Rikkola:
master:
https://github.com/droolsjbpm/drools-wb/commit/508ff85c7

6.3.x:
https://github.com/droolsjbpm/drools-wb/commit/e16964ef9

PR #60, opened 2015-09-02
BZ1256625: Conversion of XLS to GDT fails when date-effective and date-expires are included

manstis:


drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/main/java/org/drools/workbench/screens/dtablexls/backend/server/conversion/builders/GuidedDecisionTableDateEffectiveBuilder.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright 2012 JBoss Inc
```
psiroky: It is already 2015 :)

manstis:
@psiroky 2015 already? Really? Corrected in PR...

Rikkola:
Same for 6.3.x https://github.com/droolsjbpm/drools-wb/commit/9aa4f5825

PR #68, opened 2015-09-15
Styling tweaks following BS3 upgrade.

manstis:
@cristianonicolai Hi, I've made the changes you recommended. 

In addition to using the proposed widgets I also now only ```setStyleName()``` on vanilla GWT ```ScrollPanel```s otherwise I ```addStyleNames()``` (but the occurrence of this has also been kept to an absolute minimum; most of the time I'm now using vanilla PF\BS3 widgets).

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/pages/ActionInsertFactFieldsPageViewImpl.ui.xml
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/pages/ActionInsertFactFieldsPageViewImpl.ui.xml
@@ -18,7 +18,8 @@
 <ui:UiBinder xmlns:ui='urn:ui:com.google.gwt.uibinder'
              xmlns:gwt='urn:import:com.google.gwt.user.client.ui'
              xmlns:guvnor='urn:import:org.drools.workbench.screens.guided.rule.client.editor'
-             xmlns:b="urn:import:org.gwtbootstrap3.client.ui">
+             xmlns:b="urn:import:org.gwtbootstrap3.client.ui"
+             xmlns:pf="urn:import:org.uberfire.client.views.pfly.widgets">
 
   <ui:with field="i18n" type="org.drools.workbench.screens.guided.dtable.client.resources.i18n.GuidedDecisionTableConstants"/>
   <ui:with field="ufImages" type="org.uberfire.ext.widgets.common.client.resources.CommonImages"/>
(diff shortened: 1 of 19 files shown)


cristianonicolai:
@manstis thanks! looking really good, glad to help ;)

I also added a few more comments of things we could improve/consider for future versions.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/pages/ActionInsertFactFieldsPageViewImpl.ui.xml
```diff
@@ -26,90 +27,99 @@
   <ui:with field="res" type="org.uberfire.ext.widgets.core.client.resources.WizardResources"/>
 
   <gwt:VerticalPanel ui:field="container" width="100%">
```
cristianonicolai: I would try to avoid using Vertical and Horizontal panels as possible. Using the Bootstrap grid system should be enough for composing  our widgets, with the plus of also being responsive. Have a look in this simple example here: https://github.com/cristianonicolai/kie-wb-common/blob/f5554d0037327c4d3806d1288958df2f284e4aff/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/popups/about/AboutPopup.ui.xml

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/pages/ActionInsertFactFieldsPageViewImpl.ui.xml
```diff
@@ -26,90 +27,99 @@
...
-    <gwt:VerticalPanel>
-      <b:Label text="{i18n.DecisionTableWizardDescriptionActionInsertFactFieldsPage}" styleName="{res.css.wizardDTableCaption}"/>
+    <gwt:VerticalPanel width="100%">
+      <b:Label text="{i18n.DecisionTableWizardDescriptionActionInsertFactFieldsPage}"/>
```
cristianonicolai: I think you could use <b:Legend> in here, if I recall correctly, it already adds a border/hr below it.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/pages/ActionInsertFactFieldsPageViewImpl.ui.xml
```diff
@@ -26,90 +27,99 @@
...
+          <b:FormGroup ui:field="bindingContainer">
+            <pf:FormLabelHelp for="txtBinding" addStyleNames="col-lg-4" text="{i18n.Binding}:" helpTitle="{i18n.BindingFact}" helpContent="{i18n.BindingDescription}"/>
+            <gwt:FlowPanel addStyleNames="col-lg-7">
+              <guvnor:BindingTextBox ui:field="txtBinding" enabled="false"/>
+              <b:HelpBlock ui:field="txtBindingHelp" text="{i18n.DecisionTableWizardDuplicateBindings}" visible="false"/>
+            </gwt:FlowPanel>
+          </b:FormGroup>
+          <b:FormGroup>
+            <pf:FormLabelHelp for="chkLogicalInsert" addStyleNames="col-lg-4" text="{i18n.LogicallyInsert}:" helpTitle="{i18n.LogicallyInsert}" helpContent="{i18n.LogicallyAssertAFactTheFactWillBeDeletedWhenTheSupportingEvidenceIsRemoved}"/>
+            <gwt:FlowPanel addStyleNames="col-lg-7">
```
cristianonicolai: For cases like this, would be preferable to use <b:Column size="LG_7">. I still have doubts which size is actually the best for us MD or LG, but that is a different topic.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/pages/ActionInsertFactFieldsPageViewImpl.ui.xml
manstis: Fair enough.. this PR was just to fix the Guided Decision Table Wizard - that was broken (style-wise) after the BS3/PF migration. We have many, many uses of GWT's "layout" widgets (VerticalPanel, HorizontalPanel, SimplePanel etc) throughout the workbench (and Uberfire)... replacing them all would be a sizeable undertaking!

PR #70, opened 2015-09-18
GUVNOR-2269: Work Item Definition editor does not open following migration to BS3-PF

manstis:
See https://issues.jboss.org/browse/GUVNOR-2269


porcelli:
https://github.com/droolsjbpm/drools-wb/commit/45ad7218de2ac4e5522d044bee34595996452595

PR #72, opened 2015-09-25
GUVNOR-2271: Post-BS3 migration: XLS Decision Table screen stopped working

manstis:
See https://issues.jboss.org/browse/GUVNOR-2271


manstis:
This only needs to be applied to master.

PR #75, opened 2015-10-02
Death to the rogue errai.marshalling.serializableTypes entries...

manstis:


manstis:
*drools-wb*
(master) http://github.com/droolsjbpm/drools-wb/commit/c33ea56d9

manstis:
Thanks @manstis!

PR #76, opened 2015-10-14
GUVNOR-2268 : Test Scenarios: Switch to using uberfire-extensions DatePicker

Rikkola:
Part uno. Can be merged in as it is. 
Part dos will remove the legacy date picker from kie-wb-commons and needs to be done after this. 

datepicker


Rikkola:
Ok, found another location where the date picker is used. Please wait with the PR merge.

Rikkola:
Ok, at first I only did the given section. Had to also do expect section.

manstis:
*drools-wb*
(master) http://github.com/droolsjbpm/drools-wb/commit/c067b2379

manstis:
Thanks @Rikkola !

PR #77, opened 2015-10-15
BZ1269366: Indexing DRL related files logs null error instead of real cause

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1269366


manstis:
Related PRs:

https://github.com/uberfire/uberfire-extensions/pull/89
https://github.com/droolsjbpm/kie-wb-common/pull/109

manstis:
master and 6.3.x please!

drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/src/test/java/org/drools/workbench/screens/drltext/backend/server/indexing/IndexRuleInvalidDrlTest.java
```diff
@@ -0,0 +1,119 @@
...
+    @Test
+    public void testIndexRuleInvalidDrl() throws IOException, InterruptedException {
+        //Setup logging
+        final Logger root = (Logger) LoggerFactory.getLogger( Logger.ROOT_LOGGER_NAME );
+        final Appender<ILoggingEvent> mockAppender = mock( Appender.class );
+        when( mockAppender.getName() ).thenReturn( "MOCK" );
+        root.addAppender( mockAppender );
+
+        //Add test files
+        final Path path = basePath.resolve( "bz1269366.drl" );
```
Rikkola: Good idea for a file name.

Rikkola:
I dare say, that the test IndexRuleInvalidDrlTest.testIndexRuleInvalidDrl is failing.
This happened after building uberfire-extensions and kie-wb-common with your changes.

manstis:
What does it fail with?

I suspect something like match was greater than expected? - but it'd be
great if this is not the case!!

manstis:
If my assumption is correct can you help please (a) check the code you have
contains the IndexersFactory.clear() called in BaseIndexer, and (b)
IndexersFactory only contains one Indexer when the test runs. Use of the
singleton leads to multiple Indexers being registered for subsequent tests.

If the failure is caused by something else please let me know!

manstis:
OK, worked out the problem :)

Originally I logged as errors; but changed - just before committing to
warnings - however logback was not configured to record warnings; so the
assertions failed.

I'm correcting the PR by adding logback-test.xml configuration files and
confirmed the tests pass locally (for all of drools-wb).

drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/pom.xml
```diff
@@ -143,6 +143,17 @@
...
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
```
psiroky: It is enough to add dependency on `logback-classic` as it depends transitively on `logback-core`. (this is not really important, just fyi).

manstis: @psiroky Thanks Petr! To be honest I could really do with an explanation as to our logging configuration works - I still see some [log4j](https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-webapp/src/main/resources/log4j.xml) configuratrion files floating around...

psiroky: There are few rule and some exceptions. From what I was able to observe there are three main rules to follow:
 * always use SLF4J API (slf4j-api jar) for the logging calls (not log4j, Java Util Logging, or Commons Logging)
 * use Logback as logging implementation (logback-classic) where it makes sense (testing, end-user apps/bundles)
 * exclude other logging implementations coming transitively. These are usually commons-logging and log4j. Lot's of 3rd party deps are dragging logging implementations with them. That is generally bad and should be fixed on the library side. But until that happens we need to exclude them ourselves. Of course nothing is that easy, so whenever you exclude the logging impl, you should make sure that there is a logging bridge defined so that the logging actually works for those 3rd party libraries. See http://www.slf4j.org/legacy.html which nicely explains these bridges. So for example whenever you exclude log4j, you should add dependency on log4j-over-slf4j (unless it's already there).

Only "end-user" applications, like jcr2vfs migration tool or kie-config-cli tool should actually depend on the logging implementation. One exception are the WAR files which should _not_ ideally contain any logging implementation and we should just rely on the logging impl bundled within the container. Other modules should only use it with test scope, so it is not included in the transitive dependency list. 

We should remove any dependency on log4j and any log4j.xml files. 

Huh, I am not sure the explanation above is clear enough as it is hard to describe some of these rules and exceptions. Hopefully I haven't made this more complicated than it already was :)

Rikkola:
uberfire-extensions
-------------------
(master) http://github.com/uberfire/uberfire-extensions/commit/65a421036

(0.7.x) http://github.com/uberfire/uberfire-extensions/commit/a27b7bdfb

kie-wb-common
-------------
(master) http://github.com/droolsjbpm/kie-wb-common/commit/ffb49c85a

(6.3.x) http://github.com/droolsjbpm/kie-wb-common/commit/bd96322ac

drools-wb
---------
(master) http://github.com/droolsjbpm/drools-wb/commit/72c17a21d

(6.3.x) http://github.com/droolsjbpm/drools-wb/commit/1cc32060e

drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/pom.xml
yurloc: @psiroky that is a great elaboration! It is perfectly clear, it's just difficult to keep it all in mind all the time. It would be best if you turn this into a wiki page in droolsjbpm-build-bootstrap, wdyt?

Just a single note: it is a good idea to have `src/test/resources/logback-test.xml` so that logging works during test execution.

PR #80, opened 2015-10-21
Copy 'Full Text' search implementation from kie-wb-distributions.

manstis:
Copy back ```kie-wb-distributions``` implementation of "Full Text Search" to support debugging issues with said feature in ```drools-wb```. This is only required on 6.3.x as master does not support a top-level widget bar following the BS3 migration.

### DIFF
--- a/drools-wb-webapp/pom.xml
+++ b/drools-wb-webapp/pom.xml
@@ -142,6 +142,10 @@
       <artifactId>uberfire-widgets-commons</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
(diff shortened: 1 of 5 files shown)


Rikkola:
+1 
There are no tests, but this bit is just for drools-webapp that is our dev enviroment. Lighter to run than the full kie-drools-webapp in kie-wb-distributions.

csadilek:
Thanks!

6.3.x: 3bc964df507aae924542e36b6257a38f958feefd and 4468a23fe5827518caa13cd501ea7022ce8af0a4

PR #85, opened 2015-11-02
BZ1274349: Asset Management: Branch selector partially hidden. Setup drools-wb-webapp to support Asset Management.

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1274349

This adds support for Asset Management to drools-wb-webapp; to support local development without needing to run kie-wb. There's no tests, there's no documentation. None of which is needed for drools-wb-webapp.

### DIFF
--- a/drools-wb-webapp/src/main/java/org/drools/workbench/backend/server/PicketLinkDefaultUsers.java
+++ b/drools-wb-webapp/src/main/java/org/drools/workbench/backend/server/PicketLinkDefaultUsers.java
@@ -3,7 +3,7 @@
  * Copyright 2014, Red Hat, Inc. and/or its affiliates, and individual
  * contributors by the @authors tag. See the copyright.txt in the
  * distribution for a full listing of individual contributors.
- *
+ * <p/>
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Salaboy:
@manstis: is this missing in kie-wb and kie-drools-wb as well (6.3.x branch)? Because nothing seems to be working related with asset management

manstis:
@Salaboy This is purely a cosmetic tweak to prevent the list of branches being clipped when it overflows Project Explorer's container width. It has nothing, unfortunately, to do with what I saw you and @pefernan discussing on IRC :(

Salaboy:
@manstis But this is also adding some roles and users in the PicketLink configuration, is that also added in the other projects? because it might be the problem

manstis:
@Salaboy The additional configuration is to only enable Asset Management features in drools-wb-webapp; e.g. New Repository Wizard -> support for Managed Repositories and Project Explorer -> Branch Selector. The Picketlink configuration is not needed in kie-(drools)-wb as these handle authentication differently. The workbench-policies.properties file also already exists in kie-(drools)-wb.

csadilek:
+1

6.3.x: 9afbdee1830a2afba6746ab34c9a25e0c26209f0

PR #88, opened 2015-11-05
webapp: fix GWT SDM in IDEA

psiroky:
* IDEA does not honor maven resource filtering,
   see https://youtrack.jetbrains.com/issue/IDEA-96854.
   This change workarounds it by inheriting the GWT app
   modules the other way around. Now IDEA will only use
   .gwt.xml which does need any filtering, and the filtering
   will still be done in Maven buils. This change should not
   have any influence on how the resulting WAR will look like

@manstis, @Rikkola could you please try this locally to make sure I haven't missed something?


manstis:
@psiroky @Rikkola I can confirm this works well in IDEA :)

My only comment would be to add a whacking big comment in the ```gwt.xml``` files explaining why they are set up as they are - it'll be all too tempting for a "do gooder" to think "oh, this should surely be in the non-Fast configuration file...."

psiroky:
I was even thinking about renaming the files to something like `DroolsWorkbench.gwt.xml` and `FullDroolsWorkbench.gwt.xml` (or AllPermutationsDroolsWorkbench.gwt.xml`), but that would break the current setup for everyone using the FastCompiler*. 

But I do agree with adding the comment, so will do that.

psiroky:
I've removed the "full" GWT webapp module as it seems it is not being used at all. After talking to @manstis and @Rikkola this seems like the best option. Please take a look.

The webapp currently contains only `DroolsWorkbench.gwt.xml`, so you will need to update your IDEA config (as it was using `FastCompiledDroolsWorkbench.gwt.xml`). Or I could rename the file as well, so we would just have the `FastCompiled*` one. Let me know if this is what would you prefer.

manstis:
+1 to merge.

csadilek:
Thanks.

master: 3c5dee7b873843c6e627da6c7a735513d271ecff

PR #92, opened 2015-11-24
Backports - Clone some changes made to kie-wb-distributions to drools-wb

manstis:


manstis:
ok to test

manstis:
ok to test

psiroky:
@manstis just FYI: you don't need to write "ok to test" if you update the PR. Jenkins will pickup the change automatically and run the build again (it will notice the HEAD of the PR branch is different)

manstis:
@psiroky Thanks for the info.. in this case the first "ok to test" was because Jenkins failed to connect to github; but the second "ok to test" was following an additional commit fixing a missing transient dependency.

psiroky:
@manstis I see. In the first case you can use phrase "jen-kins retest this" (without the hyphen, I am adding it here just to avoid triggering another build). It is effectively the same as "ok to test", but it adds some additional info, e.g. you are explicitly requesting rebuild due to (likely) random failure. "ok to test" is usually used to approve PR build for contributors which are not white-listed. Really just a detail though....

Rikkola:
+1
Just drools-webapp changes, only used by me an manstis

csadilek:
Thanks.

master: ad3e4ae858be144a19176ebf4a79590f423d1d98 70ff7e73c48fd8c27c8e248d3490f9258a3fb088 8c1d473d8a13432b404e66aa69195adb70cb4ce6

PR #96, opened 2015-11-26
UF-244: (Work-around) Default DEV environment to Wildfly 10.0.0(CR4)

manstis:
See https://issues.jboss.org/browse/UF-244

This PR provides a workaround to the issue (of ```Async``` EJBs not running on Wildfly 8.1 - which is our Hosted Mode environment). Support for System Property ```org.uberfire.async.executor.safemode``` to force use of ```Executors.newCachedThreadPool``` has been added.

This PR sets the said System Property to "true" to ensure Hosted Mode (which uses Wildfly 8.1) works without further ado with ```mvn gwt:run``` or from your favourite IDE.


csadilek:
+1

master: 0419787f4ea23a5e7dbe2c7b10d0f7e5b525faca

PR #104, opened 2016-01-04
GUVNOR-2327: Ace Editor layout doesn't respect asset editor bounds

manstis:
See https://issues.jboss.org/browse/GUVNOR-2327

View implementation changes only... hence no tests.

The "rinky dink" size constants are really a manifestation of issues with the BS3 TabPanel.. widgets getting their parent size would expect this to exclude the TabPanel header; however it appears it includes the header height and hence when setting the height of "child" Widgets we need to take the TabPanel header height into account :(


manstis:
See https://issues.jboss.org/browse/GUVNOR-2328 for the generic issue of Widget.getParent().getOffsetHeight() *including* the ```TabPanel``` header height...

csadilek:
+1

csadilek:
master: e294514d6c0af28f36cfeb249327948c240c2975

manstis:
Thanks, I'm working on the referenced JIRA too at the moment (both are on
Edson's "QE" list.

PR #105, opened 2016-01-05
BZ-1197649 - Updating addAtribute return type to FormStyleItem instea…

paulovmr:
…d of int.


paulovmr:
Depends on:
https://github.com/uberfire/uberfire-extensions/pull/158

manstis:
+1 to merge (when the ```uf-ext``` PR is merged).

psiroky:
The CI build failed with compilation error. Is that related to these changes or is it false negative? (in which case I need to fix it to avoid similar issues in future).

manstis:
@psiroky I'm assuming it's because the change needed for this PR is in ```uberfire-extensions``` - IDK if the "automated chain" has picked up https://github.com/uberfire/uberfire-extensions/pull/158

psiroky:
@manstis the changes in uf-ext were (should have been) picked up (based on the same branch name). But looking at the failure again it is actually in guvnor repo. Will try to run this again to see if we can reproduce.

paulovmr:
@psiroky I think this is a false negative, because the build is ok on my machine. But looking at the error, it seems related to a PR that was closed yesterday (https://github.com/droolsjbpm/guvnor/pull/209). I think its changes weren't considered on this build.

psiroky:
Hmm, it indeed looks like the build picked incompatible versions of uf (or uf-ext) and guvnor. Some changes landed in uf (uf-ext), then this CI build took all the repos and only after that rest of the changes were pushed into guvnor.

This is, however, something which is almost impossible to detect and should happen only rarely.

psiroky:
Well, I still haven't figured out why the build is failing. The same branch is being built as part of Guvnor, where it works as expected. https://github.com/droolsjbpm/guvnor/pull/210

paulovmr:
jenkins retest this

csadilek:
master: dc4f0b900719d08bf2b16b87947ff2147b85fc11

PR #121, opened 2016-02-08
GUVNOR-2364: Invalid enumeration is validated without error

manstis:
See https://issues.jboss.org/browse/GUVNOR-2364

Move conversion of ```String``` to UI model to Presenter, new ```Cell``` to disable editing for rows with incomplete definition and tests.

### DIFF
--- a/drools-wb-screens/drools-wb-enum-editor/drools-wb-enum-editor-client/src/test/java/org/drools/workbench/screens/enums/client/editor/EnumEditorPresenterTests.java
+++ b/drools-wb-screens/drools-wb-enum-editor/drools-wb-enum-editor-client/src/test/java/org/drools/workbench/screens/enums/client/editor/EnumEditorPresenterTests.java
@@ -1,0 +1,217 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.workbench.screens.enums.client.editor;
+
+import java.util.ArrayList;
+import java.util.List;
+import javax.enterprise.event.Event;
+
+import com.google.gwt.core.client.GWT;
+import com.google.gwtmockito.GwtMock;
+import com.google.gwtmockito.GwtMockitoTestRunner;
+import org.drools.workbench.screens.enums.client.type.EnumResourceType;
+import org.drools.workbench.screens.enums.model.EnumModel;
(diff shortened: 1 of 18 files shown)


manstis:
See related PR: https://github.com/droolsjbpm/kie-wb-common/pull/241

drools-wb-screens/drools-wb-enum-editor/drools-wb-enum-editor-client/src/test/java/org/drools/workbench/screens/enums/client/editor/EnumEditorPresenterTests.java
```diff
@@ -0,0 +1,217 @@
...
+    @Mock
+    private PlaceRequest place;
+
+    @Mock
+    private Overview overview;
+
+    @GwtMock
+    private ViewDRLSourceWidget sourceWidget;
+
+    @Captor
```
Rikkola: Maybe I should read the Mockito docs again. Didn't realize I could do this. This is a lot nicer way.

drools-wb-screens/drools-wb-enum-editor/drools-wb-enum-editor-client/src/test/java/org/drools/workbench/screens/enums/client/editor/EnumEditorPresenterTests.java
```diff
@@ -0,0 +1,217 @@
...
+                              "['a', 'b']" ) );
+        }} );
+
+        this.enumServiceCaller = new CallerMock<EnumService>( enumService );
+
+        this.presenter = new EnumEditorPresenter( view,
+                                                  enumServiceCaller,
+                                                  type ) {
+            {
+                //Yuck, yuck, yuck... the class hierarchy is really a mess
```
Rikkola: The presenters are not really easy to test. We should make them more testable or make the setup easier with writing some helper classes to guvnor-test-utils.

Rikkola:
+1

drools-wb-screens/drools-wb-enum-editor/drools-wb-enum-editor-client/src/test/java/org/drools/workbench/screens/enums/client/editor/EnumEditorPresenterTests.java
manstis: @Rikkola I agree (this is the second time I've had to set ```@Injected'ed``` fields in a static initializer block. Ensuring ```BaseEditor``` and ```KieEditor``` use Constructor injection would help for sure.. as we'd need to pass objects through parent constructors. This would fix another issue I found where ```BaseEditor``` and ```KieEditor``` both ```@Inject Event<Notification>``` leading to the super and subclasses both having instances of the event with the same field name... which screwed up tests (I've removed the duplicate in a previous PR).

csadilek:
master: c07a5d6955213d1e5bef41d9e1a857cedda645a3

drools-wb-screens/drools-wb-enum-editor/drools-wb-enum-editor-client/src/test/java/org/drools/workbench/screens/enums/client/editor/EnumEditorPresenterTests.java
Rikkola: For what it is worth, I made a ticket for this.
https://issues.jboss.org/browse/GUVNOR-2383

PR #127, opened 2016-02-24
BZ1301051: Column values are deleted after changing the operator in a Guided Decision Table

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1301051

This PR splits the values only of ```STRING``` cells (which are the default for "Value Lists" and "Enumerations"). Before https://github.com/droolsjbpm/drools/pull/656 it was possible a ```DTCellValue52``` was a ```NUMERIC``` but had a ```valueString==""```.


manstis:
I'll look at test failures later...

drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/main/java/org/drools/workbench/screens/dtablexls/backend/server/conversion/builders/GuidedDecisionTableSalienceBuilder.java
```diff
@@ -78,9 +79,10 @@ public void addCellValue( final int row,
             value = value.substring( 1,
                                      value.lastIndexOf( ")" ) - 1 );
         }
-        final DTCellValue52 dcv = new DTCellValue52();
+        final DTCellValue52 dcv = new DTCellValue52( 0 );
         try {
             dcv.setNumericValue( Integer.valueOf( value ) );
+            /**/
```
Rikkola: Best to say nothing if you have nothing to say.

Rikkola:
+1

PR #129, opened 2016-02-25
[6.3.x] BZ1301055: [GSS](6.2.z) Column values are deleted after changing the operator in a Guided Decision Table

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1301055

This is the corollary of https://github.com/droolsjbpm/drools-wb/commit/9058ac86d for 6.3.x

NOTE 6.3.x has a different BZ number but the fix is the same.

(cherry picked from commit 9058ac86dcc15f50fac7bb8edf915cef3b1df606)


manstis:
See https://github.com/droolsjbpm/drools/pull/659 too.

Rikkola:
+1

mswiderski:
merged into 6.3.x

PR #130, opened 2016-02-25
BZ1310208: When converting .XLS to GDT, Globals defined with short Class Names are converted to invalid Globals file

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1310208

The "Globals" Editor in the Workbench creates "Globals" with Fully Qualified Class Names; however the conversion of XLS Decision Table to Guided Decision Table creates "Globals" with whatever was defined in the XLS. 

This is not a problem for XLS->DRL generation as a single DRL ```String``` containing ```import``` and ```global``` is generated and hence short-names can be resolved. However since Guided Decision Table->DRL generation uses a separate "Globals" DRL definition file that cannot contain ```imports``` short-class names fail to resolve. This PR ensures generated "Global" definitions are FQCNs.


manstis:
There are tests - you'll need to click on the "View" button. I renamed the existing test class to better reflect the tests it performed; and re-purposed the old existing test class name for the new tests.

Rikkola:
+1

manstis:
**INVESTIGATING TEST FAILURES ON 6.4.x PR - PLEASE DO NOT MERGE**

manstis:
@psiroky Please merge at your leisure!

PR #141, opened 2016-03-09
GUVNOR-2446: NPE in SafeSessionInfo.getIdentity() leading to 'Anonymous' user being used

manstis:
See https://issues.jboss.org/browse/GUVNOR-2446

The ```@Inject```ed ```SessionInfo``` is produced by ```UberFireGeneralFactory``` which is tied to Errai's ```RpcContext```. XLS Decision Tables and XLS Score Cards are uploaded through a ```Servlet``` and not Errai's ```RPC``` therefore the produced failed leading to ```null``` ```SessionInfo```.


csadilek:
+1

csadilek:
master: b0e8e8f69dae4a2344084ea06a5a604be1379672

PR #148, opened 2016-03-22
GUVNOR-2030: Guided Decision Table: Empty field results in all Conditions being cancelled

manstis:
See https://issues.jboss.org/browse/GUVNOR-2030

- This commit fixes handling of Template Keys in non-String fields.


drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/main/java/org/drools/workbench/screens/dtablexls/backend/server/conversion/builders/ParameterUtilities.java
```diff
@@ -60,12 +80,20 @@ public String convertSingleParameterToTemplateKey(String xlsTemplate) {
         }
         parameterCounter++;
         matcherSingleParameter.appendTail( result );
+
+        if ( part == ParameterizedValueBuilder.Part.LHS ) {
```
csadilek: This block seems to be the same in ```convertIndexedParametersToTemplateKeys``` and ```convertSingleParameterToTemplateKey```. Prob. better to create a method and reuse?

drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/test/java/org/drools/workbench/screens/dtablexls/backend/server/conversion/GuidedDecisionTableGeneratorListenerTest.java
```diff
@@ -1729,6 +1729,165 @@ public void testMissingTemplateKeyValues() {
 
     @Test
     //https://issues.jboss.org/browse/GUVNOR-2030
+    public void testMissingTemplateKeyValues2() {
```
csadilek: What's the difference between ```testMissingTemplateKeyValues2``` and ```testMissingTemplateKeyValues```. A more descriptive test name would help here. I find it hard to understand what these tests do. I don't know much about the feature and used APIs so that contributes to that, of course ;).

manstis:
@csadilek Updates made.. thanks for reviewing.

csadilek:
master: 8a1b444f22deac69369d31b783280e587cec6660

PR #150, opened 2016-03-28
[6.4.x] [BZ-1319568] jcr2vfs: fix missing rule names + other fixes/tweaks

psiroky:
@manstis sorry to bring you into this, but I think you are the only one who dealt with some parts of jcr2vfs as well. Could you please give this a quick look?


manstis:
@psiroky Good stuff Petr! 

Regarding the TODO, David is working on an ACL rewrite for 7.x that package permissions would/should/could integrate with - you'd need to liaise with him about the details (I don't think they're finalised yet anyway). Also "Categories" became tags in 6.3 (possibly 6.2) so there's no need to migrate the categories.xml file - just tag any migrated assets with the old categories.

+1 to merge.

psiroky:
The TODO is there for quite some time (probably since the rewrite of the tool to two different modules - export + import), I will need to revisit that as well. 

Thanks for the tip with the tags, that is definitely something that would be useful I think.

PR #153, opened 2016-03-30
[BZ-1322045] jcr2vfs: don't fail fast after encountering error during migration

psiroky:
* in case there is an error/exception migrating one of the assets/modules,
   the migration tool should try to continue migrating the rest of the assets
   since they are independent and there is a high chance they will be
   migrated successfully. The tool now just logs the error and continues.
   For users it is better if the end result contains as many assets/modules
   as possible.

@manstis could you please quickly review?


manstis:
@psiroky Is it impossible to test this (changes look OK, but I wonder if we should have a test)

psiroky:
Good question. It should be possible to write a test using mocks, will take look at that.

psiroky:
I tried to come up with some meaningful and easy to understand test, but I failed. It seems it is very hard to write simple unit tests for the current code. We would need major refactoring to fix that.

manstis:
Jenkins please retest this.

manstis:
+1 to merge.

manstis:
master - https://github.com/droolsjbpm/drools-wb/commit/400941433b16f295fb6c7286ebea79e61427455d

PR #154, opened 2016-03-31
[6.4.x][BZ-1322045] jcr2vfs: don't fail fast after encountering error during migration

psiroky:
* in case there is an error/exception migrating one of the assets/modules,
   the migration tool should try to continue migrating the rest of the assets
   since they are independent and there is a high chance they will be
   migrated successfully. The tool now just logs the error and continues.
   For users it is better if the end result contains as many assets/modules
   as possible.

Backport of #153


manstis:
+1 to merge.

manstis:
6.4.x - https://github.com/droolsjbpm/drools-wb/commit/3808d0daee653e2fc8764247d14c91007575f063

PR #161, opened 2016-04-13
changed character in GuidedDecisionTableConstants_zh_TW.properties fo…

mbiarnes:
this change is for kie-wb-distribution to compile.
Michael Anstis fixed this on Zanata server but it is not yet "available" doing zanata:pull-module.


manstis:
+1 to merge.

I've also updated in Zanata: https://translate.jboss.org/webtrans/translate?project=drools-wb&iteration=7.0.0&localeId=zh-Hant&locale=en#view:doc;doc:org.drools/drools-wb-guided-dtable-editor-client/org/drools/workbench/screens/guided/dtable/client/resources/i18n/GuidedDecisionTableConstants (page 6)

PR #164, opened 2016-04-14
fix compilation failure due to the use of String.format(..)

zkrejcov:
@manstis this should help, right? ;-)


manstis:
@zkrejcov Right!

I'll let Jenkins turn green before merging.

manstis:
+1 to merge.

PR #170, opened 2016-04-29
BZ-1329121: [fr_FR] Ned to translate a string on Authoring -> Project Authoring -> New Item->guided rule page.

paulovmr:


csadilek:
+1

PR #171, opened 2016-05-04
latest Zanata translation changes in drools-wb from 2016-05-04

kiereleaseuser:


manstis:
There should probably be a space after "Inserted column"{0}"...." but since this came from Zanata perhaps you could email Angela? 

Otherwise +1 to merge (it'll be changing again for sure... since it's not translated anyway)

mbiarnes:
me:
Hi Angela,
Doing today zanataull-module in drools-wb (https://translate.jboss.org/iteration/view/drools-wb/7.0.0)
I saw when doing a diff of drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/resources/org/drools/workbench/screens/guided/dtable/client/resources/i18n/GuidedDecisionTableConstants_zh_TW.properties

-DecisionTableAuditLogInsertWorkItemSetFieldColumn0={0} to set a field on an existing fact with value from Work Item.
+DecisionTableAuditLogInsertWorkItemSetFieldColumn0=\ {0} to set a field on an existing fact with value from Work Item.
that the character \ was introduced again.
This is fixed for 6.4.0 project but not for 7.0.0 project.
Could you please remove the space (as Alex suggested) also in 7.0.0 ?

Angela: 
Hi Michael B,
I have tried to fix it now. Let me know if you still have issues.

mbiarnes:
@manstis 
+1 to merge
The beginning empty space, responsible for the wrong character "\"  was taken out.
The \ is not anymore there.

PR #172, opened 2016-05-04
cleanup in DecisionTableXLSServiceImpl & ScoreCardXLSServiceImpl

zkrejcov:
- close input/output streams
- avoid code duplication
- unify exception handling
- add validation when saving xls dt
- and also properly close the Workbook created as a means of validation
- use File instead of InputStream for the Workbook creation as it is supposed to be more memory-friendly, see https://poi.apache.org/apidocs/org/apache/poi/ss/usermodel/WorkbookFactory.html#create(java.io.InputStream)

@manstis something for your experienced eye, I think. ;) It _should_ be okay, but you just never know...

This might also help BZ1332961.

### DIFF
--- a/drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/main/java/org/drools/workbench/screens/dtablexls/backend/server/DecisionTableXLSServiceImpl.java
+++ b/drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/main/java/org/drools/workbench/screens/dtablexls/backend/server/DecisionTableXLSServiceImpl.java
@@ -17,7 +17,6 @@
 package org.drools.workbench.screens.dtablexls.backend.server;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -30,6 +29,7 @@ import javax.inject.Named;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.poi.openxml4j.exceptions.InvalidFormatException;
+import org.apache.poi.ss.usermodel.Workbook;
 import org.apache.poi.ss.usermodel.WorkbookFactory;
 import org.drools.decisiontable.InputType;
 import org.drools.decisiontable.SpreadsheetCompiler;
@@ -147,8 +147,28 @@ public class DecisionTableXLSServiceImpl
                         final InputStream content,
                         final String sessionId,
                         final String comment ) {
+        return writeToFile(resource, content, sessionId, comment, true);
+    }
+
+    @Override
+    public Path save( final Path resource,
+                      final InputStream content,
+                      final String sessionId,
+                      final String comment ) {
+        return writeToFile(resource, content, sessionId, comment, false);
+    }
+
+    private Path writeToFile( final Path resource,
+                              final InputStream content,
+                              final String sessionId,
+                              final String comment,
+                              boolean create) {
         final SessionInfo sessionInfo = getSessionInfo( sessionId );
-        log.info( "USER:" + sessionInfo.getIdentity().getIdentifier() + " CREATING asset [" + resource.getFileName() + "]" );
+        String userAction = "UPDATING";
+        if (create) {
+            userAction = "CREATING";
+        }
+        log.info( "USER:" + sessionInfo.getIdentity().getIdentifier() + " " + userAction + " asset [" + resource.getFileName() + "]" );
 
         FileOutputStream tempFOS = null;
         OutputStream outputStream = null;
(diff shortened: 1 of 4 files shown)


drools-wb-screens/drools-wb-scorecard-xls-editor/drools-wb-scorecard-xls-editor-backend/src/main/java/org/drools/workbench/screens/scorecardxls/backend/server/ScoreCardXLSServiceImpl.java
```diff
@@ -249,15 +247,24 @@ public Path copy( final Path path,
     @Override
     public List<ValidationMessage> validate( final Path path,
                                              final Path resource ) {
+        InputStream inputStream = null;
```
triceo: Since I see Java8-specific code elsewhere in this PR, maybe we want to convert this into try-with-resources?

zkrejcov: I was thinking about that, but (from a brief look) it seems we wouldn't be able to do
throw ExceptionUtilities.handleException( ioe );
so I rather kept it as it is.

drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/main/java/org/drools/workbench/screens/dtablexls/backend/server/DecisionTableXLSServiceImpl.java
```diff
@@ -147,8 +147,27 @@ public Path create( final Path resource,
...
+
+    @Override
+    public Path save( final Path resource,
+                      final InputStream content,
+                      final String sessionId,
+                      final String comment ) {
+        return create(resource, content, sessionId, comment, false);
+    }
+
+    Path create( final Path resource,
```
manstis: Personally, I'd prefer this to be called ```writeToFile(..)``` rather than ```create(..)``` since it doesn't create for the "update scenario".. also it'd be good to have the code formatter ran over this; as the indentation is screwy.

drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/main/java/org/drools/workbench/screens/dtablexls/backend/server/DecisionTableXLSServiceImpl.java
```diff
@@ -147,8 +147,27 @@ public Path create( final Path resource,
...
+                        final InputStream content,
+                        final String sessionId,
+                        final String comment,
+                        boolean create ) {
+        final SessionInfo sessionInfo = getSessionInfo( sessionId );String userAction = "UPDATING";
```
manstis: Please split this line for readability sake.

drools-wb-screens/drools-wb-scorecard-xls-editor/drools-wb-scorecard-xls-editor-backend/src/main/java/org/drools/workbench/screens/scorecardxls/backend/server/ScoreCardXLSServiceImpl.java
```diff
@@ -125,67 +126,64 @@ public InputStream load( final Path path,
...
 
+    @Override
     public Path save( final Path resource,
                       final InputStream content,
                       final String sessionId,
                       final String comment ) {
+        return create(resource, content, sessionId, comment, false);
+    }
+
+    private Path create( final Path resource,
```
manstis: Same as for the Decision Table service (e.g. ```writeToFile(...)```).. any reason the method is ```private``` here but package protected in ```DecisionTableXLSServiceImpl```?

manstis:
@zkrejcov A few minor things; otherwise great stuff. Thanks.

Could you please consider my comments before this is merged?

zkrejcov:
Thanks @manstis! I'll fix what you pointed out. Sorry for that - I guess I was missing caffeine or something. ;-)  (I honestly don't get how those two lines got onto one line..)
Any preference/advice on the bits that @triceo pointed out?

manstis:
@zkrejcov I have no real preference about using "try with resources" to be honest; but it should be possible to call ```ExceptionUtilities.handleException(...)``` in a ```catch``` block *with* "try with resources". See https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html (if it helps!)

zkrejcov:
After trying it out a bit, even exceptions from stream closing are included in the stacktrace, as 'Suppressed' (and 'thrown normally' if the try block doesn't throw anything). So, good, we will not loose the info - I will switch to that as well.

zkrejcov:
Changes should now be in. I decided against the try-with-resources as it is not reasonably usable in all of the try blocks and mixing the approaches would create more of a chaos.

zkrejcov:
jenkins retest this

psiroky:
The build is failing because of the uf + uf-ext merge. Fixing that now.

manstis:
+1 to merge.

PR #174, opened 2016-05-09
jcr2vfs: properly shutdwon Weld after test finishes

psiroky:
Backport of #173. Touches just the test.


manstis:
+1 to merge (when green)

PR #175, opened 2016-05-09
[master] BZ 1271539 - fix the browser window title reset

zkrejcov:
This was probably left-over from some testing phase... The code might need more clean-up, but I don't see into it well enough.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-api/src/main/java/org/drools/workbench/screens/guided/dtable/analysis/AnalysisMessage.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-api/src/main/java/org/drools/workbench/screens/guided/dtable/analysis/AnalysisMessage.java
@@ -1,0 +1,103 @@
+package org.drools.workbench.screens.guided.dtable.analysis;
+
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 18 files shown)


manstis:
The code you are removing does not need removing; it's the only feedback Users get on the progress of Guided Decision Table analysis - which,  for large tables, can take 10's of seconds.

zkrejcov:
@manstis and it results in clearing the browser window title. Only for this one editor. Every time you open/focus a guided dt. 
I agree that it is good to give feedback, but... This is not a proper way to report such a thing. Besides the quite unique reporting style that is not used anywhere else in the gui, quite often there is only the beginning of the text displayed by the browser. If I expect some feedback from a webapp, I would rather expect it _inside_ the webapp, not "from the browser". Especially if everything else happens inside the webapp.
So, most times I'd guess, you now have the users do a double take - "what did I do...did I mess something up?" - because they will just see that the title gets cleared - for no obvious reason.

manstis:
I agree with everything you've said; but your PR removes the (poor) feedback we give and replaces it with nothing. If you can replace with a status message in the Message Console (one entry that updates to show progress and then is removed) it'd be perfect :-)

zkrejcov:
Looking into options..

Rikkola:
Doing some changes on this area. Currently I print the info into the message console for testing, but the best I can think of would be in Analysis Reporting panel. A small bar on top or something like that.
Ticket for the issue. 
https://issues.jboss.org/browse/GUVNOR-2563

manstis:
Assigned to @Rikkola .. please advise if this PR is still relevant; needs merging or closing etc.

Rikkola:
Better to close. The ticket I made will replace this once fixed.

PR #178, opened 2016-05-19
add tests for filename validation in gdt wizard - summary page

zkrejcov:
goes together with https://github.com/uberfire/uberfire/pull/376 but is not dependent on it
(part X of moving-tests-to-community series)


manstis:
+1 to merge.

PR #187, opened 2016-05-30
JBPM-5030: Access control based on the new UF's authz subsystem

dgutierr:
Changes required to support the new UF's security APIs

This PR depends on uberfire/uberfire#321


porcelli:
Am I missing something, or this PR removes security from drools-wb? (couldn't find any policy file changed nor restrictions applied)

dgutierr:
Security is enabled for any UF app.

A default app security policy can be set by adding the security-policy.properties files to the webapp's classpath. Also the SecurityManagementPerspective can be added included into the app's menus in case you want to edit user/groups/roles and permissions.

csadilek:
+1

PR #191, opened 2016-06-07
GUVNOR-2521: Consolidate server-side bootstrap logic in various workbench implementations

manstis:
```drools-wb-webapp``` never used to set System Property ```org.uberfire.watcher.autostart`` to ```false``` however following consolidation of the backend startup logic this System Property is set in ```DefaultApplicationScopedProducer``` and hence ```drools-wb-webapp``` was failing to startup various backend services properly.

### DIFF
--- a/drools-wb-webapp/src/main/java/org/drools/workbench/backend/server/AppSetup.java
+++ b/drools-wb-webapp/src/main/java/org/drools/workbench/backend/server/AppSetup.java
@@ -18,6 +18,7 @@ package org.drools.workbench.backend.server;
 
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.event.Event;
 import javax.inject.Inject;
 import javax.inject.Named;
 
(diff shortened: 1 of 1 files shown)


manstis:
@paulovmr Would you mind reviewing please?

manstis:
@paulovmr @csadilek There's an email from QE stating it appears in ```kie-(drools)-wb``` 7.0.0-SNAPSHOT "events" appear to not working correctly; e.g. Project Explorer is not refreshing to show newly created files in Authoring. I'm looking into at the moment. 

Without the fix in this PR ```drools-wb-webapp``` fails to startup any ```WatchServices``` which makes it appear events have stopped working as a lot of the UI uses events from VFS from the ```WatchServices```. On initial investigation ```kie-(drools)-wb``` should be raising this event and hence not be affected...

paulovmr:
@manstis Sure, thanks for fixing that. Please, let me know if I can help with this investigation.

+1 to merge (when Jenkins is green)

manstis:
Jenkins ran out of memory.. should be OK though.. let's try again...

manstis:
Jenkins please retest this.

csadilek:
+1, thanks!

PR #192, opened 2016-06-09
PLANNER-579: Reduce WARN messages during startup

mcimbora:
Fixes warnings when starting OptaPlanner WB in Super Dev Mode

         Computing all possible rebind results for 'org.drools.workbench.screens.guided.dtable.client.resources.GuidedDecisionTableResources'
            Rebinding org.drools.workbench.screens.guided.dtable.client.resources.GuidedDecisionTableResources
                        [WARN] Charset declaration detected. A charset at-rule is not relevant inside a style element


manstis:
+1 to merge (I've had this change on a local branch for ages; but never had reason to commit it)

PR #194, opened 2016-06-10
GUVNOR-2507: Guided Decision Table Editor: Add ability to open multiple decision tables in the same view

manstis:
See https://issues.jboss.org/browse/GUVNOR-2507

This PR integrates the changes made in https://github.com/droolsjbpm/kie-wb-common/pull/382 to the Guided Decision Table editor. It also adds the ability to create a new Guided Decision Table and open it in an already open GDT editor (rather than in a new instance of the editor). Tests for new and some tests for some of the existing code.


manstis:
Part of an ensemble:-

https://github.com/uberfire/uberfire/pull/387
https://github.com/droolsjbpm/kie-wb-common/pull/382
https://github.com/droolsjbpm/drools-wb/pull/194

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
```diff
@@ -317,24 +329,42 @@ public void refreshContent( final ObservablePath path,
...
+                                              public String getTitle() {
+                                                  return currentPath.getFileName() + " - " + resourceType.getDescription();
+                                              }
+                                          },
+                                          new Runnable() {
+                                              @Override
+                                              public void run() {
+                                                  //Nothing to do
+                                              }
+                                          } ),
```
csadilek: last two parameters to LockTarget constructor could be lambda expressions to make this statement shorter.

PR #199, opened 2016-06-14
GUVNOR-2552: Guided Decision Table Editor: Fix column Synchronizers

manstis:
See https://issues.jboss.org/browse/GUVNOR-2552

```Synchronizer``` implementations need a small change to help differentiate between different column types; and Unit Tests were changed to represent the *real* runtime configuration (previously they were too isolated; i.e. only testing a single configured ```Synchronizer``` rather than having multiple implementations available at runtime).

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/ActionInsertFactColumnSynchronizer.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/ActionInsertFactColumnSynchronizer.java
@@ -35,6 +35,11 @@ public class ActionInsertFactColumnSynchronizer extends ActionColumnSynchronizer
     }
 
     @Override
+    public boolean handlesAppend( final MetaData metaData ) throws ModelSynchronizer.MoveColumnVetoException {
+        return handlesUpdate( metaData );
+    }
+
+    @Override
     public boolean handlesUpdate( final MetaData metaData ) throws ModelSynchronizer.MoveColumnVetoException {
         if ( !( metaData instanceof ColumnMetaData ) ) {
             return false;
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/ActionInsertFactColumnSynchronizer.java [hunk 1]
```diff
@@ -35,6 +35,14 @@ public int priority() {
...
 
     @Override
+    public boolean handlesAppend( final MetaData metaData ) throws ModelSynchronizer.MoveColumnVetoException {
+        if ( !( metaData instanceof ColumnMetaData ) ) {
+            return false;
+        }
+        return ( (ColumnMetaData) metaData ).getColumn() instanceof ActionInsertFactCol52;
+    }
+
+    @Override
```
paulovmr: @manstis All the new handlesAppend methods will be always the same as the handlesUpdate methods? Maybe a method extraction makes sense here.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/ModelSynchronizerImpl.java
```diff
@@ -205,7 +205,8 @@ public void deleteCell( final GridData.Range rowRange,
     @SuppressWarnings("unchecked")
     public void appendColumn( final BaseColumn column ) throws MoveColumnVetoException {
         final MetaData metaData = new BaseColumnSynchronizer.ColumnMetaDataImpl( column );
-        for ( Synchronizer synchronizer : synchronizers ) {
+        for ( int i = 0; i < synchronizers.size(); i++ ) {
+            final Synchronizer synchronizer = synchronizers.get( i );
```
paulovmr: @manstis Any obscure reason to make this change? Seems like a slower iterator on a List.

paulovmr:
@manstis I made some comments, apart from that all seems good. :)

paulovmr:
+1 to merge (when Jenkins is green)

manstis:
@paulovmr Fixed the reported issues - glad I had it reviewed :-)

The change to iterator has been reverted (it was made when I was trying to figure out why GWT seemed to be iterating the ```Synchronizer``` collection in a different order to my tests, but was later to be found that my tests were too isolated). I've made the ```handleAppend(..)``` calll ```handleUpdate(...)``` too.

paulovmr:
@manstis Thanks!

+1 to merge (when Jenkins is green)

PR #200, opened 2016-06-14
Quick fix for Test Scenario results screen...

manstis:
Added zero-argument constructor as the screen was failing to be instantiated...


manstis:
@Rikkola Could you please +1?

PR #202, opened 2016-06-16
GUVNOR-2561: Guided Decision Table shows options selection twice

manstis:
See https://issues.jboss.org/browse/GUVNOR-2561

I also changed the order of the sections to be Metadata, Attributes, Conditions, Actions (which is the same order top to bottom as columns appear in the table left to right).


Rikkola:
Lightning fast fix!

+1

PR #204, opened 2016-06-17
GUVNOR-2527 : Rule index for Verification & Validation

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2527

A better index for workbench data. It used to be just in lists.
Now it is in maps and you can make queries against the maps. Save the query and rehit it when needed. It is also possible to save a query and listen if the result changes.

Could not find any regression from the changes. Old tests pass, found few issues with smoke testing and added tests for them. 

Speed benefits:
* No more iterative loops for searching
* A cell value change no longer rebuilds all the checks against that row. It updates the value in the index and the checks that care about the change. Rechecking is still a bit greedy, but better than before.
* Removing a column just removes the column and does not rebuild the data index + inspectors


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/InspectorMap.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    protected LeafInspectorList<Value> getNewSubMap() {
+        return new LeafInspectorList<>();
+    }
+
+    @Override
+    public ArrayList<Value> hasConflicts() {
+        for ( final GroupBy groupBy : keys() ) {
+            final ArrayList<Value> result = get( groupBy ).hasConflicts();
+            if ( !result.isEmpty() ) {
```
manstis: Won't this "exit fast" on the first conflict found? Should it not aggregate all conflicts, doing ```result.addAll( get( groupBy ).hasConflicts()``` Am I making any sense!?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/InspectorMap.java
```diff
@@ -0,0 +1,112 @@
...
+                return result;
+            }
+        }
+
+        return new ArrayList<>();
+    }
+
+    @Override
+    public RedundancyResult<GroupBy, Value> hasRedundancy() {
+        for ( final GroupBy groupBy : keys() ) {
```
manstis: I guess likewise for redundancy; doesn't this "exit fast" on the first redundancy found? Does that give full V&V coverage; or will later redundancies only be reported when earlier ones (the first to cause this to exit) are removed?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/ConditionBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+                .select().first();
+    }
+
+    private Short getShortValue() {
+        if ( realCellValue.getNumericValue() != null ) {
+            return ( Short ) realCellValue.getNumericValue();
+        } else if ( realCellValue.getStringValue() == null || realCellValue.getStringValue().isEmpty() ) {
+            return null;
+        } else {
+            return new Short( realCellValue.getStringValue() );
```
manstis: These can throw ```NumberFormatException```, may be better to catch and use a default (0?) value.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/ConditionBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    private Long getLongValue() {
+        if ( realCellValue.getNumericValue() != null ) {
+            return ( Long ) realCellValue.getNumericValue();
+        } else if ( realCellValue.getStringValue() == null || realCellValue.getStringValue().isEmpty() ) {
+            return null;
+        } else {
+            return new Long( realCellValue.getStringValue() );
```
manstis: Same here...

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/ConditionBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    private Double getDoubleValue() {
+        if ( realCellValue.getNumericValue() != null ) {
+            return ( Double ) realCellValue.getNumericValue();
+        } else if ( realCellValue.getStringValue() == null || realCellValue.getStringValue().isEmpty() ) {
+            return null;
+        } else {
+            return new Double( realCellValue.getStringValue() );
```
manstis: Same here...

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/ConditionBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    private BigInteger getBigIntegerValue() {
+        if ( realCellValue.getNumericValue() != null ) {
+            return ( BigInteger ) realCellValue.getNumericValue();
+        } else if ( realCellValue.getStringValue() == null || realCellValue.getStringValue().isEmpty() ) {
+            return null;
+        } else {
+            return new BigInteger( realCellValue.getStringValue() );
```
manstis: Same here...

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/ConditionBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    private BigDecimal getBigDecimalValue() {
+        if ( realCellValue.getNumericValue() != null ) {
+            return ( BigDecimal ) realCellValue.getNumericValue();
+        } else if ( realCellValue.getStringValue() == null || realCellValue.getStringValue().isEmpty() ) {
+            return null;
+        } else {
+            return new BigDecimal( realCellValue.getStringValue() );
```
manstis: Same here...

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/ConditionBuilder.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    private Integer getIntegerValue() {
+        if ( realCellValue.getNumericValue() != null ) {
+            return ( Integer ) realCellValue.getNumericValue();
+        } else if ( realCellValue.getStringValue() == null || realCellValue.getStringValue().isEmpty() ) {
+            return null;
+        } else {
+            return new Integer( realCellValue.getStringValue() );
```
manstis: Same here...

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/Field.java
```diff
@@ -0,0 +1,143 @@
...
+            return name.compareTo( field.name );
+        } else {
+            return factType.compareTo( field.factType );
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        int result = factType.hashCode();
+        result = 31 * result + name.hashCode();
```
manstis: I'm still scared something stupid not using ```result = ~~result" in your ```hashCode``` implementations.. but if you're happy.. then.... you'll be the one scratching your head *if* the JIRAs start rolling in about V&V going crazy after a while..

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/utilities/ColumnUtilities.java
```diff
@@ -415,6 +415,10 @@ public String getBoundFactType( String boundName ) {
         return "";
     }
 
+    public GuidedDecisionTable52 getModel() {
```
manstis: Doesn't ```DecisionTableAnalyzer``` have the ```GuidedDecisionTable52``` object set in it's constructor? Getting it from ```ColumnUtilities``` "smells wrong"; as this would have it set in it's constructor too - possibly from ```DecisionTableAnalyzer```?

drools-wb-webapp/src/main/resources/org/drools/workbench/DroolsWorkbench.gwt.xml
```diff
@@ -3,6 +3,8 @@
     "http://www.gwtproject.org/doctype/2.7.0/gwt-module.dtd">
 <module>
 
+  <inherits name="com.google.gwt.logging.Logging"/>
```
manstis: Oee.. I cannot remember exactly, but log4J is available client side. @mbarkley @csadilek Would you please be so kind as to remind me (and Toni) of the prerequisites for use?

manstis:
@Rikkola Great stuff.. just some (minor) comments really.

It'd be great to have some JavaDocs (or Google Doc, or something) explaining how this (should) work.. so if you was to ever overdose on Power Vodka those left with your legacy would have a chance of following the algorithm!

drools-wb-webapp/src/main/resources/org/drools/workbench/DroolsWorkbench.gwt.xml
csadilek: @manstis yes slf4j is available client-side (simply inject a logger) and preferred: http://docs.jboss.org/errai/4.0.0.Beta1/errai/reference/html_single/#sid-74908675_Logging-ClientSideSetup /cc @Rikkola @mbarkley

psiroky:
I am not 100 % sure, but it seems merging https://github.com/droolsjbpm/kie-wb-common/pull/388 is causing the drools-wb build fail as we also need this PR merged. Or am I overlooking something?

manstis:
@psiroky I believe you're correct.. @zkrejcov has pointed out the same on another failing PR... I did check with @Rikkola before merging the ```kie-wb-commons``` PR and was assured it was independent of others... perhaps this was not the case :-(

psiroky:
This would btw never happen it we had the WB code in one repository - just saying :)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/InspectorMap.java
Rikkola: You are and it might change to that way. Depends what the needs end up being.
Currently I'm just taking the first reason for failure and reporting that. If the user fixes this then the loop continues and it reports the next one.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/InspectorMap.java
Rikkola: Yep, doing this to keep the code more simple and fail faster for speed reasons. If there ever is time I'll be storing the result and just listening to changes on the items in it. This reduces the need to rerun these loops and it probably makes sense to store all the problem values in the result. (Originally thought I'll just keep the first two that cause trouble, but your questions are giving me ideas.)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/ConditionBuilder.java
Rikkola: Good point, better to keep it safe. I'll fix this in the next PRs if you are ok with that. Happens to be that I already had to change this bit, but didn't fix this issue. 
This is a copy paste of the old, just moved to another location, so there should be no regression.

Rikkola: Fixed locally for all the Numbers.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/Field.java
Rikkola: I'm not happy. Thanks again for reminding.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/ConditionBuilder.java
Rikkola: I'll just fix them for this one ;)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/utilities/ColumnUtilities.java
Rikkola: True, I suspect I was lazy here. There is the danger that someday ColumnUtilities loses the model and that causes an unnecessary hit on the analyser codes.

drools-wb-webapp/src/main/resources/org/drools/workbench/DroolsWorkbench.gwt.xml
Rikkola: Ok. Used this for testing. Killing the line.

Rikkola:
@psiroky @manstis My mistake. Double checked and while the code does not depend on the kie-wb-commons change. I forgot the tests that use the event.

Rikkola:
@manstis I'll look into Javadocs, probably the best way since any other doc just gets lost.

Rikkola:
Good review. Fixed the issues except the fast failure. I think we are better having it at the moment.

Rikkola:
More lies. Non test code also uses the event.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/RuleBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+import org.drools.workbench.models.guided.dtable.shared.model.Pattern52;
+import org.drools.workbench.screens.guided.dtable.client.widget.table.utilities.ColumnUtilities;
+
+import static org.drools.workbench.screens.guided.dtable.client.widget.analysis.index.Utils.*;
+
+public class RuleBuilder {
+
+    private final Index                 index;
+    private       GuidedDecisionTable52 model;
```
manstis: @Rikkola I guess this is your change to exposing ```ColumnUtilities.getModel()```? Shame it's not final like the others.. but nitpicking..

manstis:
+1 to merge (just a minor nit-pick about a non-```final``` instance variable)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/RuleBuilder.java
Rikkola: It should be.
Fixed here and in other locations.

Rikkola:
finals added.

PR #207, opened 2016-06-22
Verification and Validation

Rikkola:
Next steps. 
Fixes two bugs really. The summit example @manstis did picked these both up. 

This PR adds support for:
- BRL condition/column
- Action/column cell having a null value
- Action/column cell having no value
- Action/column cell having more than one values


manstis:
@Rikkola Nice small PR for me to start my day with ;-)

Rikkola:
@manstis Hehe, have to keep pushing these, but I think this is the last one with this big impact. 
Again, only one who needs this ATM is me and I have it locally, so no hurry.

manstis:
@Rikkola 3 failing tests (not sure why they should fail.. unless related to something in your PR?) Could you please confirm everything passes locally for you (with the latest fetch from master)?

manstis:
Jenkins please retest this.

Rikkola:
Could be that my merges broke something when I cherry picked from one topic branch to another. Checking.

Rikkola:
Ok. I see them. I broke something not related to V&V.

Rikkola:
Same 3 also fail in master.
(Still doesn't mean I didn't break them)

Rikkola:
Fix on the way.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/RuleInspector.java
```diff
@@ -56,6 +71,44 @@ public RuleInspector( final Rule rule,
 
             patternInspectorList.add( patternInspector );
         }
+
+        updateBRLActionInspectors( rule.getActions()
+                                       .where( Action.superType().is( ActionSuperType.BRL_ACTION ) )
+                                       .select().all() );
+        rule.getActions()
+            .where( Action.superType().is( ActionSuperType.BRL_ACTION ) )
+            .listen().all( new AllListener<Action>() {
```
manstis: FYI, this type of construct can be simplified with lamdas to read ```.listen().all( (all)->updateBRLActionInspectors(all) );``` or even (IIRC) ```.listen().all( (all)->this::updateBRLActionInspectors );``` (which IMO is slightly *less* readable!)

Rikkola:
Works on Idea, fails in CLI. 
Could be this.
https://groups.google.com/forum/embed/#!topic/mockito/LCWH3ykLJtA

Rikkola:
Or not. Something similar.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/FieldConditionsBuilder.java
```diff
@@ -0,0 +1,97 @@
...
+                                                        conditionCol52,
+                                                        row.get( columnIndex ) );
+            field.getConditions().add( condition );
+            rule.getConditions().add( condition );
+        }
+    }
+
+    private Field resolveField( final ConditionCol52 conditionCol52 ) {
+        return Utils.resolveField( pattern,
+                                                conditionCol52.getFieldType(),
```
manstis: Minor nit-pick, re: formatting.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/RuleBuilder.java
```diff
@@ -59,14 +62,25 @@ private void resolvePatterns() {
...
-                                   pattern ).buildConditions();
+            new FieldConditionsBuilder( index,
+                                        model,
+                                        rule,
+                                        row,
+                                        utils,
+                                        pattern ).buildConditions( pattern52.getChildColumns() );
+        }
+
+        for ( final CompositeColumn<? extends BaseColumn> column : model.getConditions() ) {
```
manstis: You could possibly wrap the iterator over ```getPatterns()``` and ```getConditions()``` into a single loop and use ```instanceof Pattern52``` and ```instanceof BRLConditionColumn``` inside the same loop.

manstis:
@Rikkola a few trivial-ignorable comments (mainly FYI)... when green I'll +1 and merge.

manstis:
@Rikkola let me check the failing tests from CLI here too (w/o your PR.. I can fix as appropriate)

manstis:
@Rikkola probably this "Actually the issue is that the anonymous class is not public, it is compiled by the Sun compiler as a package visible class" as the test sub-classes a class with a package-protected method and then spies on the sub-class.

Rikkola:
Ah, I checked the class, but not the method. Well spotted.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/RuleInspector.java
Rikkola: Tried that but GWT compile broke. Have you been able to do that on client side?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/index/RuleBuilder.java
Rikkola: Ok. Good to know. I'll change this.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/RuleInspector.java
manstis: @Rikkola Yes, I do use it quite a lot e.g. https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java#L962

manstis: @Rikkola I find lamdas work OK (on the whole, although IIRC @porcelli had an issue with one use-case).. none of the Java 8 "stream" API is implemented in GWT2.8.0.beta1 so any use of that client-side is broken (but it's available in newer GWT2.8's)

Rikkola: Ok, thank you. Need to check what I did wrong.

Rikkola:
Fixed formatting and the iterator. I still need to rebase once #208 is in.

manstis:
Jenkins please retest this.

manstis:
@Rikkola #208 is merged.

manstis:
+1 to merge.

PR #208, opened 2016-06-22
GUVNOR-2544: Guided Decision Table Editor: Layout manager for tables. Fix failing tests (only fail on CLI).

manstis:
Fixes build failures on master (existing Unit Tests fail on CLI but are OK in IDEA).


manstis:
@Rikkola Could you please review...

Rikkola:
+1 If Jenkins is ok.

PR #212, opened 2016-06-23
GUVNOR-2559: Guided Decision Table Editor: Improve appearance (for Red Hat Summit) - Preload 'Glyphicons Halflings' font-family

manstis:


ederign:
If that is the only way, +1.

manstis:
@ederign It's the only way I found at least. When I render a Glyphicons character to the Canvas the Context2D call returns immediately and then the browser downloads/installs the fonts. Without "pre-rendering" it to an off-screen ```DIV``` the Canvas shows the default/fallback font character first and then when redrawn the correct character.. open to other ideas though!

ederign:
@manstis maybe we can (in future) pre-load this in startup and hold uberfire startup via 
https://github.com/ederign/uberfire/blob/b3c84158216b01f4f415122fbc9abf2b84040019/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/Workbench.java#L172-L172

manstis:
See https://issues.jboss.org/browse/GUVNOR-2577

PR #213, opened 2016-06-23
GUVNOR-2573: [Guided Decision Table] Audit log data refreshed only when dtable editor is reopened

manstis:
See https://issues.jboss.org/browse/GUVNOR-2573


csadilek:
+1

PR #214, opened 2016-06-23
GUVNOR-2551: Add descriptive tooltip to the Radar menu item

manstis:
See https://issues.jboss.org/browse/GUVNOR-2551


csadilek:
+1

PR #215, opened 2016-06-23
Verifier checks

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2196
https://issues.jboss.org/browse/GUVNOR-2200

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/RangeCheck.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/RangeCheck.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.workbench.screens.guided.dtable.client.widget.analysis.checks;
+
+import java.util.Collection;
+
+import org.drools.workbench.models.datamodel.oracle.DataType;
+import org.drools.workbench.screens.guided.dtable.client.resources.i18n.AnalysisConstants;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.ConditionsInspector;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.InspectorList;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.RuleInspector;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.RuleInspectorCache;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.condition.BooleanConditionInspector;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.condition.ComparableConditionInspector;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.condition.ConditionInspector;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.condition.NumericIntegerConditionInspector;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.checks.base.OneToManyCheck;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.checks.util.Redundancy;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.index.Field;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.index.FieldCondition;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.index.Rule;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.reporting.Issue;
+import org.drools.workbench.screens.guided.dtable.client.widget.analysis.reporting.Severity;
+
+public class RangeCheck
+        extends OneToManyCheck {
+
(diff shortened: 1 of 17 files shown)


Rikkola:
Taking these back. Looks like there is some sort of bug related to resolving subsumption and that affects the range checks.

Rikkola:
I'll reopen this. Working on the fix. 

The bug I found is not caused by these changes. It has been in for a while. Also the bug is not causing false reports, so the coverage this gives is better than nothing.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/RangeCheck.java
```diff
@@ -0,0 +1,193 @@
...
+                   public boolean accept( final RuleInspector other ) {
+                       return !ruleInspector.getRule().getUuidKey().equals( other.getRule().getUuidKey() );
+                   }
+               } );
+    }
+
+    @Override
+    public void check() {
+
+        // For some reason these clones always turn out to be evil.
```
manstis: Why?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/RangeCheck.java
```diff
@@ -0,0 +1,193 @@
...
+
+            return clonedConditionInspectors;
+        }
+
+        private ConditionInspector resolveInspector( final ConditionInspector originalInspector ) {
+
+            if ( originalInspector.getCondition() instanceof FieldCondition ) {
+                final FieldCondition fieldCondition = ( FieldCondition ) originalInspector.getCondition();
+
+                if ( fieldCondition.getField().getFieldType().equals( "Integer" ) ) {
```
manstis: Why is ```Integer``` treated differently to ```DataType.isNumeric(..)```?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/base/Checks.java
```diff
@@ -244,6 +247,14 @@ private void assertChecks( final RuleInspector ruleInspector,
         removedChecks.addAll( allChecks.get( removedRuleInspector ) );
         allChecks.remove( removedRuleInspector );
 
+        for ( final RuleInspector ruleInspector : allChecks.keySet() ) {
```
manstis: Is this a slow move back to the highly iterative approach? (I just remember the ```rechecks``` related code)

manstis:
Sweet.

+1 to merge.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/RangeCheck.java
Rikkola: I don't know. I have watched too many movies.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/RangeCheck.java
Rikkola: This is because x > 0 and x >= 1 are equal. The normal Inspector does not pick this up.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/base/Checks.java
Rikkola: It is. I'm planning to do reuse the maps I'm using for indexing for the checks and inspectors. That will remove this.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/checks/RangeCheck.java
manstis: So what about ```x > 0L``` and ```x >= 1L``` or ```x > 0.0``` and ```x >= 1.0```?

zkrejcov: As for `x > 0.0` and `x >= 1.0` - there is e.g. 0.5 ... so these two 'conditions' are very different, whereas with integers, `x > 0` and `x >= 1` is basically the same, as there is nothing in between 0 and 1. Good question about the long type...

Rikkola: Yep. Long is missing. Made a ticket. While this is missing for Long it does not cause false reports. Our V&V coverage just is not as good as it could be.
https://issues.jboss.org/browse/GUVNOR-2576

Another with special treatment is boolean. Since x == true and x != false are equal.

For values with decimals or any "ranged" comparable we use check the operator and then we make use of the Comparator interface.

Rikkola: And BigInteger

Rikkola:
@manstis @csadilek I believe this is ready to be merged. Remaining changes are done in new pull requests.

Rikkola:
Thank you.

PR #216, opened 2016-06-24
GUVNOR-2575: Unable to view 'Source' option for a XLSX Decision Table

manstis:
See https://issues.jboss.org/browse/GUVNOR-2575


manstis:
@Rikkola When you get a chance could you please review?

Rikkola:
+1

PR #218, opened 2016-06-27
GUVNOR-2554: Improve behavior of context popup

manstis:
See https://issues.jboss.org/browse/GUVNOR-2554


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/menu/CellContextMenuViewImpl.java
```diff
@@ -97,24 +97,40 @@ public void enableDeleteCellMenuItem( final boolean enabled ) {
     @SuppressWarnings("unused")
     @EventHandler("cellContextMenuCut")
     public void onClickCellContextMenuCut( final ClickEvent e ) {
+        if ( isDisabled( cellContextMenuCut ) ) {
```
manstis: @paulovmr @ederign @Rikkola @porcelli FYI (I got burned) "disabled" Bootstrap 3 items aren't "disabled" to DOM events... so you need to check whether they're actually disabled...

Rikkola: Thanks for the tip.

Rikkola:
+1 Nice short tests.

PR #221, opened 2016-06-27
GUVNOR-2581: Guided Decision Table Editor: Reuse old location when loading different versions of Decision Table

manstis:
See https://issues.jboss.org/browse/GUVNOR-2581


csadilek:
+1

PR #222, opened 2016-06-28
GUVNOR-2578 : Verification & Validation: Subsumption does not see fields from two different pattens as equals

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2578

V&V used the field object of the pattern to group conditions. When it should use the field of an object type. I made the two field types into two different classes, hoping they do not get mixed up in the future.


manstis:
+1 to merge.. thanks @Rikkola

PR #223, opened 2016-06-28
Add a test for RHBRMS-1841.

zkrejcov:
Operators 'in' & 'not in' should lead to String type column/cell, so that a comma can be used regardless of the field type.

(Part n+1 of 'moving sele tests to community' series.)


manstis:
Thanks @zkrejcov 

+1 to merge.

PR #224, opened 2016-06-28
GUVNOR-2555: Allow Copy and Paste across data types where applicable

manstis:
See https://issues.jboss.org/browse/GUVNOR-2555


manstis:
Hi @Rikkola Would you mind giving this the once over please?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/utils/CellUtilitiesConvertToLongTest.java
```diff
@@ -0,0 +1,77 @@
...
+                { new Long( "3" ), new Byte( "3" ) },
+                { null, new Double( "4.0" ) },
+                { null, new Float( "5.0" ) },
+                { new Long( "6" ), new Integer( "6" ) },
+                { new Long( "7" ), new Long( "7" ) },
+                { new Long( "8" ), new Short( "8" ) },
+                { new Long( "9" ), "9" },
+                { null, true },
+                { null, new Date() },
+                { null, "bananna" }
```
Rikkola: Banana

Rikkola:
+1
The converters looks like something I should reuse in the analysis codes.

PR #225, opened 2016-06-28
GUVNOR-2556: Right-clicking a cell should select it

manstis:
See https://issues.jboss.org/browse/GUVNOR-2556

Moved "Context Menu" support into it's own class; re-used the utility class in https://github.com/uberfire/uberfire/pull/422 to identify "selected cells" and wrote some tests.


manstis:
See https://github.com/uberfire/uberfire/pull/422 too.

csadilek:
+1

PR #227, opened 2016-06-29
GUVNOR-2593: Guided Decision Table Editor: Column resizing should be persisted

manstis:
See https://issues.jboss.org/browse/GUVNOR-2593

```wires-grids``` columns have their ```width``` property set as the User resizes them; this however was not copied to the ```GuidedDecisionTable52``` related ```BaseColumn``` instances that form the persisted state. Hence when the UI was saved the column widths were not saved.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/BaseMultipleDOMElementUiColumn.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/BaseMultipleDOMElementUiColumn.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/BaseMultipleDOMElementUiColumn.java [hunk 1]
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
```
manstis: This file was renamed; not "new".. so it bares the old date..

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/BaseUiColumn.java
```diff
@@ -1,74 +1,69 @@
...
-import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.columns.multiple.impl.BaseGridColumnMultipleDOMElementRenderer;
+import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.columns.GridColumnRenderer;
 
 /**
  * Base column for Decision Tables.
  * @param <T> The Type of value presented by this column
- * @param <F> The Factory to create DOMElements for this column
  */
-public abstract class BaseUiColumn<T, W extends Widget, E extends BaseDOMElement, F extends MultipleDOMElementFactory<W, E>> extends BaseGridColumn<T> implements HasMultipleDOMElementResources {
+public abstract class BaseUiColumn<T> extends BaseGridColumn<T> {
```
manstis: This is the "new" file that both ```BaseMultipleDOMElementUiColumn``` and ```BaseSingletonDOMElementUiColumn``` extend... silly old git likes to make things more complicated than they appear ;-)

Rikkola:
+1

PR #230, opened 2016-07-01
GUVNOR-2597: Guided Decision Table Editor: Exiting 'pinned mode' doesn't restore correct TransformMediator

manstis:
See https://issues.jboss.org/browse/GUVNOR-2597


manstis:
See https://github.com/uberfire/uberfire/pull/429 too.

Rikkola:
+1

PR #231, opened 2016-07-01
GUVNOR-2524: Guided Decision Table Editor: Increase Unit Test coverage of GuidedDecisionTableModellerPresenter

manstis:
See https://issues.jboss.org/browse/GUVNOR-2524

Pretty self explanatory really...


Rikkola:
+1

PR #234, opened 2016-07-05
GUVNOR-2599: Guided Decision Table Editor: Update use of SyncBeanManager to use ManagedInstance<T>

manstis:
See https://issues.jboss.org/browse/GUVNOR-2599


Rikkola:
+1

PR #235, opened 2016-07-06
GUVNOR-2528 : Verification & Validation: Store check state

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2528

Added SubsuptionBlocker and Conflict. The code runs both checks quite a few times against the same list. This speeds up the validation.

I also added Subsumption result in the same way, but based on JProfiler this just slows the validation down.

The PR also contains other changes that clean our bottle necks spotted with JProfiler. I could spend a week tuning the code, but cutting it here for now.


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/DecisionTableAnalyzerConflictResolverTest.java
```diff
@@ -0,0 +1,162 @@
...
+        final GuidedDecisionTable52 table52 = analyzerProvider.makeAnalyser()
+                                                              .withPersonAgeColumn( ">" )
+                                                              .withPersonAgeColumn( "<" )
+                                                              .withData( DataBuilderProvider
+                                                                                 .row( null, null )
+                                                                                 .end() )
+                                                              .buildTable();
+
+
+        // After a save has been done, the server side sometimes sets the String field value to "" for numbers, even when the data type is a number
```
manstis: Yes, caused by [this](https://github.com/droolsjbpm/drools/commit/bd2b052e425688e214574048763bce1b9e3cf6dc#diff-c1898016135de0a31337480ccd605a3cR45). so when a ```DTCellValue52``` is deserialized and the default constructor called the value is set to a ```String```

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/SpeedTest.java
```diff
@@ -0,0 +1,163 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.kie.workbench.common.widgets.client.datamodel.AsyncPackageDataModelOracle;
+import org.kie.workbench.common.widgets.decoratedgrid.client.widget.data.Coordinate;
+import org.mockito.Mock;
+
+import static org.drools.workbench.screens.guided.dtable.client.widget.analysis.TestUtil.*;
+
+@RunWith( GwtMockitoTestRunner.class )
+@Ignore( "Just for profiling" )
```
manstis: Does it do any harm not having this ignored? (even if it was originally written to assist profiling)

manstis:
+1 to merge... feel free to unignore test if you want to at a later date.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/DecisionTableAnalyzerConflictResolverTest.java
Rikkola: Ok, good to know.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/SpeedTest.java
Rikkola: It adds quite a lot to the test running time. Depends what is the amount of rows that was left in. I also copied the tests from another file. So the scenarios are already covered.

Rikkola: Actually. Because this is just copy paste, I should just delete the test. I can always add it back. Ignored tests are pretty much the same as dead/commented out code. Just confusing.

PR #238, opened 2016-07-07
GUVNOR-2563 : Guided Decision Table V&V: Reports check status in Analysis Reporting screen

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2563

https://dl.dropboxusercontent.com/u/2111740/guvnor/analysis.png

For long running validation shows the current status of validation. 
Yellow when running. Green when finished.
Not a progress bar. Tried those and they updating them did not work as expected.

Also cleaned up some redundant code from V&V tests.


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/DecisionTableAnalyzer.java
```diff
@@ -146,19 +141,28 @@ protected AnalysisReport makeAnalysisReport() {
...
+    private void sendReport( final AnalysisReport report ) {
+        getAnalysisReportScreen().showReport( report );
+    }
+
+    private void sendStatus( final Status status ) {
+        getAnalysisReportScreen().showStatus( status );
+    }
+
+    protected AnalysisReportScreen getAnalysisReportScreen() {
+        return IOC.getBeanManager().lookupBean( AnalysisReportScreen.class ).getInstance();
```
manstis: I know this existed in the old code; but if we're instantiating a managed bean programmatically we need also to handle the life-cycle (i.e. destroy it when done). It might be better to ```@Inject``` ```AnalysisReportScreen``` into ```GuidedDecisionTablePresenter``` and pass as a parameter to ```DecisionTableAnalyzer``` constructor. 

Just a thought.. otherwise we should probably only create a single instance in ```DecisionTableAnalyzer``` constructor and destroy it when the editor closes. 

```AnalysisReportScreenActivity``` is ```@Dependent``` meaning every time an instance is requested a new one is created... so ```getAnalysisReportScreen()``` is probably creating a new instance each time it's called.

File: drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/AnalysisReportScreenViewImpl.java [hunk 2]
```diff
@@ -16,58 +16,58 @@
...
                    RequiresResize {
 
     private AnalysisReportScreen presenter;
 
-    interface Binder
-            extends
-            UiBinder<Widget, AnalysisReportScreenViewImpl> {
 
-    }
+    @DataField( "progressBar" )
```
manstis: I know you tried a progress bar before settling on the "traffic light".. would it make sense to update the code to reflect the solution you ended up implementing?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/AnalysisReportScreenViewImpl.java
```diff
@@ -107,11 +107,27 @@ public void clearIssue() {
     }
 
     @Override
+    public void showStatusComplete() {
```
manstis: It would probably be worth ensuring these colours are in keeping with [PatternFly](https://www.patternfly.org/pattern-library/widgets/#alerts).. and even use PatternFly/BS3 "alerts"?

manstis:
A couple of comments... let me know your thoughts.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/DecisionTableAnalyzer.java
Rikkola: Ok. Made some changes to this area. There should be only a single instance of AnalysisReportScreen and the code should reuse it. I'll make sure there is one.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/AnalysisReportScreenViewImpl.java
Rikkola: It was ( I took the escape ), but since you had doubts somebody else might too. It is not the traditional progress bar we are familiar with. I'll change it to avoid confusion.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/AnalysisReportScreenViewImpl.java
Rikkola: Tried to look for better images from Bootstrap, but looks like PatternFly alerts are just what I wanted. Thank you for the tip.

Rikkola:
Good comments. I'll fix the lookupBean in the next PR and update this one with the progress bar and PatternFly alerts today.

Rikkola:
New look: https://dl.dropboxusercontent.com/u/2111740/guvnor/analysis.png
Fixed all but the `IOC.getBeanManager().lookupBean` issue. Fixed that locally for the next PR, just need to finish it.

manstis:
@Rikkola Please let me know when you want this PR reviewed again (nice new look BTW)

Rikkola:
@manstis ok. This one can be reviewed, please. The issues found here are fixed in the next PR.

manstis:
+1 to merge.

PR #239, opened 2016-07-08
GUVNOR-2604: Guided Decision Table Editor: Column definitions DisclosurePanel should scroll if content exceeds available height

manstis:
See https://issues.jboss.org/browse/GUVNOR-2604

It should be noted (and I have put similar comment on the JIRA) this is a temporary workaround as all peripheral screens and widgets relating to column definition(s) are to be rewritten as Phase 3 of the editor re-write.


csadilek:
+1

PR #244, opened 2016-07-21
Include guvnor client-side exclusions in beans.xml of drools-wb-webapp module

mcimbora:
This is required when running the workbench in multi module setup which includes guvnor. Includes additional exclusion for kie-wb-common module.


manstis:
+1 to merge.

Thanks @mcimbora

PR #246, opened 2016-07-26
GUVNOR-2615: Editing dialog for 'Pattern' doesn't show selected value

manstis:
See https://issues.jboss.org/browse/GUVNOR-2615


manstis:
@Rikkola @karreiro Please review.

Rikkola:
+1

PR #247, opened 2016-07-26
GUVNOR-2618: Mandatory new Fact Pattern binding

manstis:
See https://issues.jboss.org/browse/GUVNOR-2618

@jomarko Please see the revised message. I've included the word "binding" as it makes it consistent with other messages relating to this field/value (e.g. if you try to enter a duplicate "name" etc).


manstis:
@Rikkola @karreiro Please review.

karreiro:
+1 :)

PR #251, opened 2016-07-27
GUVNOR-2609: Field hint is not updated after selection of a fact

manstis:
See https://issues.jboss.org/browse/GUVNOR-2609


manstis:
@Rikkola @karreiro Could you please review?

karreiro:
+1

manstis:
Thanks @karreiro

PR #259, opened 2016-08-02
GUVNOR-2630 : Verification & Validation: Condition and action changes are ignored

Rikkola:


Rikkola:
@manstis @karreiro This is ready to be reviewed.

manstis:
+1 to merge.

csadilek:
The ```RuleInspectorDumper``` is introduced here but not used (neither in app code nor tests)? Should it be?

Rikkola:
It is used for debugging and bug hunting, comparing two strings is easier than using the super dev mode debug view for objects. It does feel stupid to keep it in production code. If anyone has a better place for it I can move it. I'm sure that I will also need it in the future.

csadilek:
@Rikkola ok, got it. It's not a big deal as the type will be pruned in a production compile anyway, thanks to the GWT compiler. I just wasn't sure if that was its only purpose.

PR #262, opened 2016-08-04
JBPM-5233: Data source management

wmedvede:
@manstis @csadilek 
Hello Mike, Christian,
This is the data sources management PR related to drools-wb. Basically adds a couple of compileSourcesArtifact. Would you mind to take a look?

Thanks!!
Walter.

Depends on:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/249
https://github.com/droolsjbpm/kie-wb-common/pull/416


manstis:
@wmedvede Do we need these changes? Runtime Data Source support is only added to Livespark. These webapps have no dependency on Data Sources (unless they're being brought in transiently - in which case they should probably be excluded).

wmedvede:
@manstis @csadilek I've applied the sugestions by creating the two gwt, modules, the producer, etc, and it's wroking well. There's no need to touch all the worbenches, so this PR can be closed.

PR #267, opened 2016-08-16
GUVNOR-2638 : Guided Decision Table V&V - Changes for a new line are not picked up correctly

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2638

* Improved test setups so many tests have changes
* Any files all red or all green are just moved. Except `FileInspectorUpdateTest`  

The `MultiMap` changes are what fix the bug on the indexing side.
`RuleInspectorUpdater` is what fixes the bug on the inspector side. Refreshing the `RuleInspector`.


Rikkola:
@karreiro you might be the best to review. Ask away if you have any questions.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/inspectors/RuleInspector.java
```diff
@@ -64,8 +69,8 @@
     private final InspectorList<PatternInspector>            patternInspectorList    = new InspectorList<>();
     private final InspectorList<ConditionInspector>          brlConditionsInspectors = new InspectorList<>(true);
     private final InspectorList<ActionInspector>             brlActionInspectors     = new InspectorList<>(true);
-    private final InspectorList<ActionsInspectorMultiMap>    actionsInspectors       = new InspectorList<>( true );
-    private final InspectorList<ConditionsInspectorMultiMap> conditionsInspectors    = new InspectorList<>(true);
+    private       InspectorList<ActionsInspectorMultiMap>    actionsInspectors       = null;
+    private       InspectorList<ConditionsInspectorMultiMap> conditionsInspectors    = null;
```
csadilek: The immutable final field is definitely better. Why set this to null? It's initialized in the constructor anyway (unconditionally).

csadilek: I will eat this words ;). Misread the code below.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/util/maps/ArrayMultiMap.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache.util.maps;
+
+import java.util.ArrayList;
+import java.util.SortedMap;
+
+class ArrayMultiMap<Key extends Comparable, Value>
```
csadilek: ArrayListMultiMap would be a better name here as it's backed by an ArrayList not an Array. BTW, why are we not using Guava here which provides all these MultiMap implementations and also works in GWT? It's already on the classpath too.

csadilek:
+1.

I would prefer switching to Guava's MultiMap though, and creating a custom subtype if we need one (e.g. ArrayListMultiMap already exists there). These types are already compiled into our application so no need to add more...

Rikkola:
I took a look the Guava one and it didn't quite fit the needs. That was some time ago, back then Guava limited me more than I wanted. I listen to changes in the map and the lists inside the map. 
But now that the design on the area where I use it is getting more stable, I can look if it can be changed. What I do here is probably getting closer to what Guava had. For example the ArrayListMultiMap you mentioned.

I do prefer to use anything that already exists.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/util/maps/ArrayMultiMap.java
Rikkola: Guava did not exactly fit my needs when I started this, but looking at it now. Looks like my implementation is getting closer and closer to what Guava has.

Rikkola: Hehe. The more I dig into the Guava source code the more I similarities I find. Perfect example how simple thing becomes complicated and then you were better going with something someone else did anyway.
Well, at least now it is clear what I need to override and extend and why.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/testutil/AbstractDecisionTableBuilder.java
```diff
@@ -63,6 +63,10 @@ public GuidedDecisionTable52 buildTable() {
         return table;
     }
 
+    {
+
+    }
```
karreiro: I think that we can remove this

karreiro:
+1 Nice test coverage :-)

csadilek:
@Rikkola I am happy to merge and we'll migrate to Guava after, so up to you, just let me know. I won't be working the next couple of days so @cristianonicolai can merge as well.

Rikkola:
@csadilek @cristianonicolai Sure, this can be merged. Going to work on this area for weeks. Plenty of time to migrate to Guava.

cristianonicolai:
@csadilek @Rikkola no worries, I can get it merged once tests are green.

PR #269, opened 2016-08-31
RHBPMS-856: Condition column headers are not getting reflected when a XLS Decision Table is converted to Guided Decision Table

manstis:
See https://issues.jboss.org/browse/RHBPMS-856

This PR ensures the ```CONDITION``` and ```ACTION``` column headers in the XLS file are correctly used for columns generated by the conversion.


manstis:
Part of a (small) ensemble:

- https://github.com/droolsjbpm/drools/pull/889
- https://github.com/droolsjbpm/drools-wb/pull/269

manstis:
Jenkins please retest this.

csadilek:
+1 (build problem unrelated)

do you need to internationalize "undefined" and "Converted from"?

manstis:
Jenkins please retest this.

manstis:
@csadilek It wouldn't hurt to i18n those constants.. can you recommend anything (for server side, 6.5.x and master)? I'd need to use the locale on the client/browser vs the locale on the server...

csadilek:
@manstis on the server, you can get the locale via ``RpcContext.getServletRequest.getLocale()``` and combined with standard Java resource bundles we should be able to make it work. Alternatively the server could just return the property name and you'd look up the actual value on the client?

In either case, not sure if it's worth the trouble....

manstis:
"In either case, not sure if it's worth the trouble....".. I agree (even if we i18n for *one* User, should another User - in a different locale - open the same file the text would be wrong anyway..)..

manstis:
Jenkins please retest this.

manstis:
Jenkins please retest this.

manstis:
There are test failures relating to this PR.. investigating...

PR #282, opened 2016-09-06
RHBRMS-2055: XLS Decision Table conversion fails to convert some elements

manstis:
See https://issues.jboss.org/browse/RHBRMS-2055


csadilek:
+1

PR #284, opened 2016-09-07
[6.5.x] RHBRMS-2055: XLS Decision Table conversion fails to convert some elements

manstis:
See https://issues.jboss.org/browse/RHBRMS-2055

(cherry picked from commit 2396a93db3edc2b7db4562a7f6d44224b58f3126)

This is the corollary of https://github.com/droolsjbpm/drools-wb/pull/282 for 6.5.x.


csadilek:
+1

PR #287, opened 2016-09-09
GUVNOR-2673 Default ExceptionHandler that pops up any uncaught RuntimeException

mcimbora:


manstis:
-1 (following a quick review.. I'd expressly asked for uberfire's ```DefaultErrorCallback``` *NOT* to be replaced... Need to look at this (and the others in more detail - Monday).

csadilek:
@manstis DefaultErrorCallback is still here, unchanged. This should not result in a change in behaviour.

csadilek:
@manstis maybe this line is what you didn't see: https://github.com/droolsjbpm/kie-wb-common/pull/462/files#diff-a16fcaf1425005de11fd05cffed014c7R66

manstis:
@csadilek @mcimbora Yes, that is what I did not see (with my quick review).

OK.. if you need this before Monday (when I'll look properly) perhaps @csadilek will +1 and merge them all.

manstis:
@mcimbora @csadilek Can this be merged? IIRC I saw on IRC some comment that ```@UncaughtExceptionHandler``` was not working as expected should an ```Exception``` be thrown in a ```@AfterInitialisation``` annotated method?

manstis:
+1 to merge.

Thanks for the explanation (and pointers) @csadilek

PR #288, opened 2016-09-10
[UF-403] Fork and Mirror

adrielparedes:
The goal is to have the possibility to fork any git repository inside the niogit directory.
To complete the goal we need:

* Separate git repositories into different directories. Those directories are named like the Organizational Units.
* Create a Fork command that copy all the content of the repository to fork and add the reference to the internal path of the origin repository.
* Write information about fork so we can track where those forks are, how many forks has those repositories, etc.

### DIFF
--- a/drools-wb-webapp/src/main/java/org/drools/workbench/client/perspectives/AdministrationPerspective.java
+++ b/drools-wb-webapp/src/main/java/org/drools/workbench/client/perspectives/AdministrationPerspective.java
@@ -24,6 +24,7 @@ import org.drools.workbench.client.resources.i18n.AppConstants;
 import org.guvnor.asset.management.client.editors.repository.wizard.CreateRepositoryWizard;
 import org.guvnor.common.services.shared.security.AppRoles;
 import org.guvnor.structure.client.editors.repository.clone.CloneRepositoryPresenter;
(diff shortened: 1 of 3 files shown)


manstis:
ok to test

manstis:
The [PR](https://github.com/droolsjbpm/guvnor/pull/361) (in guvnor) has lead to failure of many tests in ```drools-wb``` (and consequentially I also expect in other modules that have tests for indexing).

adrielparedes:
@manstis where can I find BaseIndexingTest? I need to change something there about internal git path (all this because of a uberfire change)

manstis:
@adrielparedes Hello, you can find it [here](https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-refactoring/kie-wb-common-refactoring-backend/src/test/java/org/kie/workbench/common/services/refactoring/backend/server/BaseIndexingTest.java).

adrielparedes:
@manstis I'm not adding the ForkPresenter/View so I'm going to close this PR because is not necessary anymore. Are you ok with that?

manstis:
@adrielparedes I am OK with closing this PR.

PR #289, opened 2016-09-12
RHBRMS-2558 : Verification & Validation: Property setting for disabling client side V&V

Rikkola:
https://issues.jboss.org/browse/RHBRMS-2558

Implementation is close to what GUVNOR-2596 does for master, but due to differences with these two branches and the other changes done in GUVNOR-2596 commits, I decided to just copy the minimum for 6.5.x. This means this is not a cherry-pick, but a copy paste from master with few changes so that the functionality fits into 6.5.x.


manstis:
+1 to merge (when green).

PR #290, opened 2016-09-12
GUVNOR-2678: Guided Decision Table Editor: Linking columns fails for BRLActionColumns with out template keys

manstis:
See https://issues.jboss.org/browse/GUVNOR-2678


manstis:
@Rikkola would you mind please reviewing?

Rikkola:
Nice and simple.

+1

PR #295, opened 2016-09-19
Add ID to GridLienzoPanel for Selenium tests

manstis:
Add an ID to ```GridLienzoPanel``` to support identification of ```KeyDownHandler``` target for Selenium tests.


manstis:
@jhrcek FYI..

manstis:
@csadilek @Rikkola @karreiro Could you please review? @jhrcek needs it to be able to target ```KeyDownEvents``` to a specific element (previously such events were handled by the "Root" panel).

PR #296, opened 2016-09-19
GUVNOR-2688: Guided Decision Tables: Default column values not applied for newly added rows

manstis:
See https://issues.jboss.org/browse/GUVNOR-2688

The issue was that the old code was calling into ```DTCellValue52( Object o )``` constructor and not ```DTCellValue52( DTCellValue defaultValue )``` constructor.


csadilek:
+1

PR #297, opened 2016-09-19
GUVNOR-2689: Guided Decision Tables: Using DatePickers to set date-effective/date-expired causes controls to disappear

manstis:
See https://issues.jboss.org/browse/GUVNOR-2689

{{ValueChangeHandler}} is called when the ```DatePicker``` is hidden which led to nested invocation and related error. When the "date" value is changed the ```DatePicker``` container (Panel) is cleared and re-constructed. The clearance led to the ```DatePicker``` being hidden and the {{ValueChangeHandler}} being invoked.. which led to an attempt to clear the panel, hiding the ```DatePicker``` etc. 

Use of ```ChangeDateHandler``` does not suffer the same nested invocation.


csadilek:
+1

PR #300, opened 2016-09-21
GUVNOR-2696: Guided Decision Table: *.XLSX can not be converted to *.GDST

manstis:
See https://issues.jboss.org/browse/GUVNOR-2696

We simply did not validate a file's extension against either XLS or XLSX.


drools-wb-screens/drools-wb-dtable-xls-editor/drools-wb-dtable-xls-editor-backend/src/main/java/org/drools/workbench/screens/dtablexls/backend/server/conversion/DecisionTableXLSToDecisionTableGuidedConverter.java
```diff
@@ -163,7 +167,7 @@ public ConversionResult convert( final Path path ) {
         ConversionResult result = new ConversionResult();
 
         //Check Asset is of the correct format
-        if ( !xlsDTableType.accept( path ) ) {
+        if ( !( xlsDTableType.accept( path ) || xlsxDTableType.accept( path ) ) ) {
             result.addMessage( "Source Asset is not an XLS Decision Table.",
```
jomarko: Just a minor thing, maybe also in the error message should be mentioned as XLS so XLSX extension.

manstis: Good catch. Fixed.

manstis:
@Rikkola Would you mind passing your eyes over this?

Rikkola:
I like how you use the actual file to test. 
+1

PR #302, opened 2016-09-22
GUVNOR-2695: Guided Decision Tables: Default column values not applied for newly added rows - part 2

manstis:
See https://issues.jboss.org/browse/GUVNOR-2695

The issue was for enumeration definitions containing a splitter between "real value" (used in DRL) and "display value" (used in the UI) the key for the enumeration definition map was incorrectly the whole enumeration definition; and not just the key component.


manstis:
@Rikkola one for your Finnish eye balls please ;-)

PR #305, opened 2016-09-23
GUVNOR-2585: In test scenario, expert contraint of type enum can be set but are always false

karreiro:

Auto-import enums from the class in order to avoid this:
enum-bug


karreiro:
@manstis Could you review this? :-)

manstis:
@Rikkola is king of Test Scenarios.. I'd prefer to defer review to him.

jhrcek:
@karreiro just out of curiosity: what tool did you use to "record" the gif?

karreiro:
@jhrcek I'm using this one: https://github.com/GNOME/byzanz ;-)

PR #306, opened 2016-09-23
[6.5.x] GUVNOR-2585: In test scenario, expert contraint of type enum can be set but are always false

karreiro:

Auto-import enums from the class in order to avoid this:
enum-bug


karreiro:
@manstis Could you review this? :-)

manstis:
@Rikkola is king of Test Scenarios.. I'd prefer to defer review to him.

manstis:
Requires https://github.com/droolsjbpm/drools/pull/935 too.

Rikkola:
+1

PR #309, opened 2016-10-04
GUVNOR-2709 : Move client side Verification codes to a separate module

Rikkola:
What is happening here.
* Verification code is moved to drools-wb/drools-wb-services/drools-wb-verifier
* drools-wb-verifier tries to be clean from GWT, there is still the GWT Scheduler and i18n to be cleaned.

GWT Scheduler might not be needed once this is a web worker.

i18n is best to do out side of the web worker since the worker does not know what language is selected. Also this reduces the amount of permutations needed. As a side effect we get a verification module that can be ran on the server side if needed.

The code is bound to be split even more. Right now everything is in -api.

https://issues.jboss.org/browse/GUVNOR-2709
There are 3 PRs in the change bundle:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/287 
https://github.com/droolsjbpm/drools-wb/pull/309
https://github.com/droolsjbpm/kie-wb-distributions/pull/369


Rikkola:
@manstis @karreiro If you would please review, or anyone else that has time.

Rikkola:
`AnalyzerConfiguration` is the new thing here. Everything else has just been dragged and dropped to a new location.

Rikkola:
jenkins please retest

manstis:
@Rikkola You may need to sync your ```droolsjbpm-build-bootstrap``` fork with ```master```

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/Util.java
```diff
@@ -20,14 +20,16 @@
...
-import org.drools.workbench.screens.guided.dtable.client.widget.analysis.index.keys.Value;
+import org.drools.workbench.services.verifier.api.client.cache.util.maps.MultiMap;
+import org.drools.workbench.services.verifier.api.client.index.keys.Value;
+import org.drools.workbench.services.verifier.api.client.index.keys.Key;
 
 import static org.junit.Assert.*;
 
 public class Util {
 
+    // TODO: Check if dead code
```
manstis: Did you check if it was dead?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/ParameterizedCommand.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.drools.workbench.services.verifier.api.client;
+
+/**
+ * A command representing a future activity. This was deliberately created in
+ * addition to the existing GWT Command to allow better re-use of menu
+ * structures when a WorkbenchPart is embedded within Eclipse.
+ */
+public interface ParameterizedCommand<T> {
```
manstis: I assume you duplicated this class to keep the Verifier GWT module as clean\isolated as possible?

Rikkola: There should be only one, since we run the command only in one location. I'll remove the one that is not needed.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/index/DataType.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.workbench.services.verifier.api.client.index;
+
+public class DataType {
```
manstis: Again, I assume the data-type definitions have been duplicated to avoid a dependency on the module that contains the existing definitions; to presumably help with integration with Web Workers (at a later date/PR)?

Rikkola: This I had to duplicate to keep the dependencies at minimum. I hope to just use FQCN in the future.

manstis:
@Rikkola When this is green; it looks "good to go"..

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/Util.java
Rikkola: It is dead. I'll remove it.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/ParameterizedCommand.java
Rikkola: Ok. Need to keep this in for now, but removing the comment since it is confusing. This is not here because of Eclipse.

Rikkola:
Removed dead code/comments and rebased the droolsjpbm-buildstrap.

Jenkins please retest.

Rikkola:
Cause for failing builds might be that the Maven rules are more strict on this build than when run it locally. Fixed broken deps.

PR #312, opened 2016-10-07
[6.5.x] GUVNOR-2585: In test scenario, expert contraint of type enum can be set but are always false

karreiro:

The GUVNOR-2585 was not resolved by the PR https://github.com/droolsjbpm/drools-wb/pull/306 (the fields form the modal were not being imported correctly). This PR imports all the fields correctly.


karreiro:
@manstis Could you please review this? :-)

manstis:
@Rikkola could you please review.

PR #314, opened 2016-10-11
GUVNOR-2583: Test scenario does not work if it contains java enum in the given part

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2583


karreiro:
@manstis Could you take a look? 

Since the (VerifyFact)[https://github.com/droolsjbpm/drools/blob/master/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/shared/VerifyFact.java] and the (VerifyField)[https://github.com/droolsjbpm/drools/blob/master/drools-workbench-models/drools-workbench-models-test-scenarios/src/main/java/org/drools/workbench/models/testscenarios/shared/VerifyField.java] dont't have any type information, sounds fair enough get the needed from `scenario.getImports().getImports()`.

manstis:
@karreiro This seems to fix ```master``` issue just fine. Thank-you.

PR #315, opened 2016-10-12
GUVNOR-2725: Test Scenarios: NPE when running a Scenario multiple times

manstis:
See https://issues.jboss.org/browse/GUVNOR-2725


manstis:
@Rikkola @karreiro; @nheron reported this (on ```master``` and ```6.5.x```).

The fix doesn't attempt to add "fields' types" for primitives (actually when a ```ModelField[]``` is null for a given FQCN actually). The fix also preserves the Test Scenario's "original" ```Imports``` between successive runs. Previously the list of imports would continue to grow between runs.

karreiro:
+1 Thanks @manstis

PR #320, opened 2016-10-20
Do not delete GWT unitCache during clean

csadilek:
@manstis @Rikkola Removing this line from the clean configuration makes SDM startup ~40 seconds faster (once the unitCache is in place):

```Module setup completed in 45802 ms
vs.
Module setup completed in 5721 ms```

@mbarkley and I tested this and we no longer ran into the memory issues, which required the deletion of this directory, that we experienced with earlier versions of GWT 2.8.

Would you guys mind merging this and testing it for a bit in drools-wb before we apply that fix everywhere else?

/cc @porcelli @ederign @psiroky @cristianonicolai

### DIFF
--- a/drools-wb-webapp/pom.xml
+++ b/drools-wb-webapp/pom.xml
@@ -1127,8 +1127,7 @@
                 <include>src/main/webapp/org.drools.workbench.DroolsWorkbench/</include>
                 <include>src/main/webapp/WEB-INF/deploy/</include>
                 <include>src/main/webapp/WEB-INF/classes/</include>
-                <include>src/main/webapp/WEB-INF/lib/</include>
-                <include>**/gwt-unitCache/**</include>
+                <include>src/main/webapp/WEB-INF/lib/</include>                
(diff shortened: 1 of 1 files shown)


ederign:
+1 for me. I'm working right now with drools-wb and can report any issues (if I found one).

csadilek:
We might have found another issue with these caches involving the reactor and production compiles. Please wait until we clarify before merging.

csadilek:
The error we run into after keeping the persistent caches is luckily fixed in GWT 2.8-rc3: https://github.com/gwtproject/gwt/issues/9424

So, we can activate the caches once we upgrade to rc3, which we can do any time now (already released).

csadilek:
Please merge this any time after: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/304

ederign:
@csadilek mode detailed results: the startup here is ~120 seconds faster after this PR! 😍 

Thanks for that (and @mbarkley that was also involved)!

PR #325, opened 2016-10-28
GUVNOR-2730 : [Test Scenario] Scenario have to be run twice to see fired rules

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2730


Rikkola:
@karreiro One more.

karreiro:
+1

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/test/java/org/drools/workbench/screens/testscenario/client/ScenarioEditorPresenterTest.java
```diff
@@ -167,8 +168,15 @@ public void testRunScenario() throws Exception {
 
         presenter.onRunScenario();
 
-        verify(view).initKSessionSelector(eq(path), any(Scenario.class));
-        verify(view).showAuditView(anySet());
+        InOrder inOrder = inOrder( view );
```
Rikkola: Not sure if it a good practice to test the order, but in this case when we are in a hurry I rather do that, than refactor to something where testing the order is not needed.

PR #327, opened 2016-11-01
GUVNOR-2731 : Guided Decision Tables: Verification for a single table fails to fire up

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2731


Rikkola:
@manstis Could you take a look if I fire this up at the right time. Before this V&V only starts on dtable selection, but with a single table the selection never happens.

Rikkola:
@manstis Still not sure if this is the best. I used the activate document method for single dtable.

PR #331, opened 2016-11-03
GUVNOR-2735: Guided Decision Table Editor: When opening an empty 'Graph' the editor is not populated

manstis:
See https://issues.jboss.org/browse/GUVNOR-2735

This PR ensures the "Guided Decision Table Graph Editor" is correctly setup when there are no Decision Tables in the graph. This PR also removes the cumbersome differentiation between "active" and "selected" and ensures the first Decision Table (or singular Decision Table in the "regular" editor) is pre-selected and V&V activated etc.


manstis:
Part of an emsemble:

- https://github.com/droolsjbpm/kie-wb-common/pull/538
- https://github.com/droolsjbpm/drools-wb/pull/331

manstis:
@Rikkola would you mind taking a look (since you're next to know the most about this editor!)

PR #332, opened 2016-11-07
GUVNOR-2738: Guided Decision Table Editor: Highlight rows in table when V&V report item selected

manstis:
See https://issues.jboss.org/browse/GUVNOR-2738

This PR adds highlighting of rows in a Decision Table based on the Analysis Report selection.


manstis:
@Rikkola as discussed on IRC...

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
```diff
@@ -597,6 +605,26 @@ void onUpdatedLockStatusEvent( final @Observes UpdatedLockStatusEvent event ) {
         }
     }
 
+    void onIssueSelectedEvent( final @Observes IssueSelectedEvent event ) {
+        if ( event == null ) {
```
Rikkola: Can this be null or are you just being sure?

Rikkola: Either way, I'm ok that it is there.

manstis: It probably cannot be null (if invoked by CDI).. I guess I got carried away with defensive programming (this time!)

PR #333, opened 2016-11-08
GUVNOR-2739: Guided Decision Table Graph Editor: Add ability to create new Decision Tables from within editor

manstis:
See https://issues.jboss.org/browse/GUVNOR-2739

This PR adds the ability to create new Decision Tables whilst editing a graph.

It refactors use of the Decision Table Wizard so it's more easily reusable from both the ```NewGuidedDecisionTableHandler``` and ```GuidedDecisionTableGraphEditorPresenter```.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenter.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenter.java
@@ -17,6 +17,7 @@
 package org.drools.workbench.screens.guided.dtable.client.editor;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


manstis:
Part of an ensemble:

- https://github.com/droolsjbpm/kie-wb-common/pull/545
- https://github.com/droolsjbpm/drools-wb/pull/333

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenter.java
```diff
@@ -336,6 +345,20 @@ public void init() {
...
+    }
+
+    void onNewDocument() {
+        final Path contextPath = context.getActivePackage().getPackageMainResourcesPath();
+        helper.createNewGuidedDecisionTable( contextPath,
+                                             "",
+                                             GuidedDecisionTable52.TableFormat.EXTENDED_ENTRY,
+                                             view,
+                                             ( path ) ->
+                                                     onOpenDocumentsInEditor( new ArrayList<Path>() {{
```
csadilek: Collections.singletonList(path) would be much better than the double brace initialization which creates an anonymous class every time.

manstis:
@csadilek Thanks.. updated.

csadilek:
+1

PR #335, opened 2016-11-09
GUVNOR-2629 : [Guided Decision Table] UI is not refreshed after deletion of the column from the table

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2629

The Jira only reports removing a column, but the actual bug is more than that. Updating and loading of dtables with != null or == null operators was broken.


Rikkola:
And ComparableConditionInspector.java has issues with comparing null to null, but all fixed now.

manstis:
@Rikkola I've had a quick blast through the PR.. I'll look in more detail tomorrow.

Rikkola:
No problem. I the PR is urgent ask for a review in IRC. 
This is a good one to have for the release.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/SpecialOperatorCellUpdateManager.java
```diff
@@ -0,0 +1,67 @@
...
+package org.drools.workbench.screens.guided.dtable.client.widget.analysis.cache;
+
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+import org.drools.workbench.services.verifier.api.client.index.Action;
+import org.drools.workbench.services.verifier.api.client.index.Condition;
+import org.drools.workbench.services.verifier.api.client.index.Index;
+import org.drools.workbench.services.verifier.api.client.index.keys.Values;
+import org.kie.workbench.common.widgets.decoratedgrid.client.widget.data.Coordinate;
+
+public class SpecialOperatorCellUpdateManager
```
Rikkola: That is indeed better.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/cache/DTableUpdateManager.java
```diff
@@ -79,102 +73,29 @@ public boolean update( final List<Coordinate> coordinates ) {
...
-                return true;
-            } else if ( values.isThereChanges( oldValues ) ) {
-                condition.setValue( values );
-                return true;
-            } else {
-                return false;
-            }
-        }
+    private boolean isConditionColumnWithSpecialOperator( final BaseColumn baseColumn ) {
+        final ArrayList specialOperators = new ArrayList();
```
Rikkola: Makes sense.

Rikkola:
@manstis Renamed the Special Update Manager and added NullEqualityOperator enum for null equality operators.

PR #341, opened 2016-11-16
GUVNOR-2744: field edit button disabled if fact is not selected

jomarko:


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/ActionSetFieldPopup.java
```diff
@@ -385,13 +386,25 @@ private void doBindingLabel() {
...
+            this.editField.setEnabled( false );
         }
     }
 
+    boolean isChooseConstant(String text) {
+        if ( text == null || text.isEmpty() ) {
+            return false;
+        }
+        
+        return GuidedDecisionTableConstants.INSTANCE.Choose().compareTo(text) == 0 ||
```
manstis: You should only need to check for ```GuidedDecisionTableConstants.INSTANCE.Choose()``` as this is added [here](https://github.com/jomarko/drools-wb/blob/42d50dfe809488b3520b29c73c3e52a4fed15a08/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/ActionSetFieldPopup.java#L489)

manstis: BTW it is not "nice" having to perform this check, but without more extensive refactoring it is probably the easiest option.

jomarko:
Thanks for fast review @manstis . I updated the PR according to your hint. 

I agree that after more sophisticated refactoring could be the fix nicer. However I chosen the simpler solution because I am not so experienced with this repo yet.

PR #342, opened 2016-11-16
RHBPMS-1449: added test for change of operator

jomarko:


manstis:
@jomarko Could you please confirm this is just a (passing) test?

jomarko:
@manstis Yes, the operator change was not covered in the update tests.

PR #345, opened 2016-11-21
GUVNOR-2749: Guided Decision Table Editor: New rows do not have the same height as existing rows

manstis:
See https://issues.jboss.org/browse/GUVNOR-2749


manstis:
@Rikkola @karreiro Could you please review.

karreiro:
+1

PR #347, opened 2016-11-22
GUVNOR-2755: Guided Decision Table Editor: View Menu zoom level does not reset to 100% when re-opening Editor

manstis:
See https://issues.jboss.org/browse/GUVNOR-2755

Moving to ```@Dependent``` scoped beans not only ensures the menus are correctly re-initialised but also means they work as expected when multiple Editor instances are open. Win, win!


manstis:
@Rikkola @karreiro Could you please review when you get a moment.

PR #351, opened 2016-11-24
UF-427: Admin perspective cannot be set as Home Perspective

paulovmr:


manstis:
I'd be interested in understanding the issue and cause better before commenting. Why would a Perspective not be able to be "Home" - usually this is driven by the "default" parameter in the WorkbenchPerspective annotation.

paulovmr:
Hi @manstis , it's possible to change the home perspective here:

home_perspective

Part of this issue was fixed in another PR, that added the method `setDefaultScreen` to the `AdminPage` class. Before that, it was necessary to send a place request parameter containing the default screen when opening this perspective, what is never done when opening the home perspective. This PR fixes another part of the problem, that sets the correct admin page (root) as default.

PR #357, opened 2016-12-08
GUVNOR-2758: Collapse Empty Rules warnings into one line

jomarko:
This PR tries to minimize amount of warnings in Analysis Report panel by collapsing Empty Rule warnings into single line.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/AnalysisReportScreen.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/AnalysisReportScreen.java
@@ -115,11 +115,7 @@ public class AnalysisReportScreen {
     }
 
     private ArrayList<Issue> getIssues( final AnalysisReport report ) {
-        final IssuesSet issues = new IssuesSet();
-        for ( final Issue issue : report.getAnalysisData() ) {
-            issues.add( issue );
-        }
-        return new ArrayList<>( issues );
+        return new ArrayList<>( new IssuesSet( report.getAnalysisData() ) );
     }
 
     @DefaultPosition
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/IssuesSet.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/IssuesSet.java
@@ -16,16 +16,27 @@
 package org.drools.workbench.screens.guided.dtable.client.widget.analysis.panel;
 
 import java.util.Comparator;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 import org.drools.workbench.services.verifier.api.client.reporting.ExplanationProvider;
+import org.drools.workbench.services.verifier.api.client.reporting.ExplanationType;
 import org.drools.workbench.services.verifier.api.client.reporting.Issue;
 
 public class IssuesSet
         extends TreeSet<Issue> {
 
-    public IssuesSet() {
+    private Set<ExplanationType> MERGEABLE_ISSUES = EnumSet.of(
+        ExplanationType.EMPTY_RULE,
+        ExplanationType.MISSING_RESTRICTION,
+        ExplanationType.MISSING_ACTION
+    );
+
+    public IssuesSet( List<Issue> issues ) {
         super( new Comparator<Issue>() {
             @Override
             public int compare( final Issue issue,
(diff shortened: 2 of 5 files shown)


manstis:
@jomarko When you click on an issue in the Analysis Report the offending rows are highlighted in the table. Can you please confirm the same behaviour happens with this collapsing of Analysis Report entries? I.E. if I have 2 empty rows, although only one occurrence is listed in the Analysis Report *both* rows are highlighted in the table?

jomarko:
@manstis yes it works for me in such way.

screenshot from 2016-12-08 16-17-02

manstis:
Thank-you for confirming @jomarko Let's wait for @Rikkola to approve before I merge.

jomarko:
@manstis sure, no problem. I discussed this change with @Rikkola this morning. So I am glad if he will do also the review.

Rikkola:
@jomarko Like we talked in IRC. The reporting can get crowded and we should make the small list  have as few issues listed as possible. Because of this it makes sense to merge empty rule for example in one row.

Now there is one problem with merging the reports in `SingleCheck`. We have 3 types of Checks
* `SingleCheck` - for one row
* `PairCheck` - one row against other
* `OneToManyCheck` - one row against every other row

We use the types, and rows assigned to them, to refresh the Checks. So if row 1 changes. It reruns the check `SingleChecks` with row 1, `PairChecks` with row 1 and all the `OneToMany` checks.
Because of this if we merge the lines in the `SingleCheck`. If you have 3 rows all empty. Remove the first ( or fill in the first ) and you still have rows 1-3 listed as problem lines in the other 2.

Going to write some JavaDocs for the above. Should have done that ages ago. You needed them now.

The better place for the line merge is in the report visualization. We report as detailed as possible with noise, but then visualize it in a more simple way. I would do this in `AnalysisReportScreen`.

jomarko:
@Rikkola I amended the PR. Now all EmptyRules, MissingActions and MissingConditions should be merged in the report panel.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/AnalysisReportScreen.java
```diff
@@ -153,4 +164,33 @@ public void showStatus( final Status status ) {
...
+    private Predicate<Issue> shouldNotBeMergedInReport = new Predicate<Issue>() {
+        @Override
+        public boolean test(Issue issue) {
+            return !( issue.getExplanationType() == ExplanationType.EMPTY_RULE ||
+                      issue.getExplanationType() == ExplanationType.MISSING_RESTRICTION ||
+                      issue.getExplanationType() == ExplanationType.MISSING_ACTION );
+        }
+    };
+
+    private void mergeIssues( Set<Issue> targetIssueSet, List<Issue> issues, ExplanationType type) {
```
jhrcek: Here's how I would implement the changes in this class. But feel free to ignore this :-)

```
    // Issues that we want to report on single line (the sets of lines where they occurred will be merged)
    private Set<ExplanationType> MERGEABLE_ISSUES = EnumSet.of(
            ExplanationType.EMPTY_RULE,
            ExplanationType.MISSING_RESTRICTION,
            ExplanationType.MISSING_ACTION
    );

    private ArrayList<Issue> getIssues(final AnalysisReport report) {
        Set<Issue> mergedIssues = MERGEABLE_ISSUES.stream()
                .map(typeToMerge -> mergeIssues(report.getAnalysisData(), typeToMerge))
                .filter(Optional::isPresent).map(Optional::get).collect(toSet());

        Set<Issue> individualIssues = report.getAnalysisData().stream()
                .filter(issue -> !MERGEABLE_ISSUES.contains(issue.getExplanationType()))
                .collect(toSet());

        final IssuesSet allIssues = new IssuesSet();
        allIssues.addAll(mergedIssues);
        allIssues.addAll(individualIssues);

        return new ArrayList<>(allIssues);
    }

    private Optional<Issue> mergeIssues(List<Issue> issues, ExplanationType typeToMerge) {
        Set<Issue> issuesToMerge = issues.stream()
                .filter(issue -> issue.getExplanationType() == typeToMerge)
                .collect(toSet());

        Set<Integer> affectedRows = issuesToMerge.stream()
                .flatMap(issue -> issue.getRowNumbers().stream())
                .collect(toSet());

        return issuesToMerge.stream()
                .findFirst() // Will be Optional.empty() if no issue of "typeToMerge" was present
                .map(issue -> new Issue(issue.getSeverity(), typeToMerge, affectedRows));
    }
```

jomarko: I will probably apply some of those tips. Thanks @jhrcek .

triceo: @jhrcek @jomarko  When dealing with the `IssuesSet`, you can merge those two collections by first merging their streams and then applying a collector.

Also, `issuesToMerge`` is needlessly turned into a collection - instead, the `flatMap` on `affectedRows` can be applied directly on the first stream, bypassing the need for the first collection. Although I can see a small benefit in improved readability, IMHO it is so small in this case that it doesn't justify wasting memory and CPU cycles.

jomarko:
@Rikkola according to our discussion on irc. I moved the merging logic into `IssuesSet` class.

PR #358, opened 2016-12-08
GUVNOR-2771: GDT Editor: Existing values get cleared when erasing value from value list

manstis:
See https://issues.jboss.org/browse/GUVNOR-2771


manstis:
@Rikkola @karreiro Hi guys, would one of you please review when you get a chance. Thanks.

PR #364, opened 2016-12-12
GUVNOR-2783: [Guided Decision Table] Exception thrown during handling Boolean cells

manstis:
See https://issues.jboss.org/browse/GUVNOR-2783

New columns can have the default value for ```boolean``` cells set to ```null```

@karreiro Could you please bare this in mind with the changes you are making for column definition? 

A ```boolean``` column should not really permit ```null``` default values; only ```true``` or ```false``` (unless, or until, we support tri-state ```boolean``` fields, ```null```, ```true``` and ```false``` - but it's unsupported atm).


karreiro:
Ok, @manstis. I'll keep that in mind.

+1

jomarko:
Hi @manstis. I would like to ask just for clarification. You said null values are not supported atm. Does it mean, that if I define condition column that compares some boolean field (lets say `mandatory`) for equality, then all generated rules will contain either `mandatory == true` or `mandatory == false`?

manstis:
@jomarko Yes that is correct. We've never supported the (really needed) "tri-state" in any of the editors where ```boolean``` is concerned; only ```true``` or ```false```.

PR #365, opened 2016-12-12
[6.5.x] RHBRMS-2714: [GSS](6.4.z) GDT editor: existing values get cleared when erasing value from value list

manstis:
See https://issues.jboss.org/browse/RHBRMS-2714

This is the corollary of https://github.com/droolsjbpm/drools-wb/pull/358 for 6.5.x

It is however a different implementation as the code-base in this area between 7.x and 6.5.x is different.


manstis:
@Rikkola could you please do me the honours of reviewing?

PR #366, opened 2016-12-13
GUVNOR-2775 : Rule Verification: Make it possible to disable/enable checks

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2775

Isolated bit of the hit modes work that is coming. Adds a white list for the checks that are allowed to run. Web worker still only runs with the default. Default has all the checks.

Also ran into few bugs ( that I made ) and for this reason improved the logging.


Rikkola:
@jomarko You might want to take a look at this, but anyone is free to comment.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/IssuesSet.java
```diff
@@ -23,18 +23,18 @@
...
 
 public class IssuesSet
         extends TreeSet<Issue> {
 
-    private Set<ExplanationType> MERGEABLE_ISSUES = EnumSet.of(
-        ExplanationType.EMPTY_RULE,
-        ExplanationType.MISSING_RESTRICTION,
-        ExplanationType.MISSING_ACTION
-    );
+    private Set<CheckType> MERGEABLE_ISSUES = EnumSet.of(
```
triceo: Can we make this `static`? No need to carry this collection on every instance.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/panel/IssuesSet.java
```diff
@@ -82,13 +82,14 @@ private int compareRowNumbers( final Set<Integer> rowNumbers,
                                 .filter( Optional::isPresent ).map( Optional::get ).collect( Collectors.toSet() ) );
 
         addAll( issues.stream()
-                      .filter( issue -> !MERGEABLE_ISSUES.contains( issue.getExplanationType() ) )
```
triceo: If we invert the flow here, as in `issues.stream().filter(...).filter(...).foreach(add(...))`, no collection will ever need to be created - saving CPU cycles on creating the collection and then immediately GC'ing it.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/DetectImpossibleMatchCheck.java
```diff
@@ -67,21 +68,20 @@ public Issue getIssue() {
...
+                                         fieldName,
+                                         toHumanReadableString( conflict.getOrigin()
+                                                                        .getConflictedItem() ),
+                                         toHumanReadableString( conflict.getOrigin()
+                                                                        .getConflictingItem() ),
+                                         ruleInspector.getRowIndex() + 1 );
+    }
 
-        return issue;
+    public static CheckType getType() {
```
jomarko: @Rikkola this method is kind of suspicious. I can not find another use of this method in the PR? Is it intended?

Rikkola: Yep. This is some left over code that should not be here. Removing it.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/DetectRedundantActionCheck.java
```diff
@@ -73,7 +75,7 @@ public Issue getIssue() {
                                                          new HashSet<>( Arrays.asList( ruleInspector.getRowIndex() + 1 ) ) );
         } else {
             return new ValueForActionIsSetTwiceIssue( Severity.WARNING,
-                                                      ExplanationType.VALUE_FOR_ACTION_IS_SET_TWICE,
+                                                      CheckType.VALUE_FOR_ACTION_IS_SET_TWICE,
```
jomarko: Maybe also here should be used the variable `checkType`?

jomarko: Yes, I mean, there is used correct CheckType at the commented line, but probably it would be worth not to throw two kinds of issues from one kind of checker.

jomarko:
@Rikkola I am done with review. I have just one thing in mind similar to issue we found in `DetectRedundantActionCheck`.

In classes `DetectRedundantRowsCheck` and `SingleHitCheck` the `CheckType` is not hardcoded in constructor whereas in other checkers it is. At least for the `DetectRedundantRowsCheck` it seems to me as indication that the checker doesn't satisfy single responsibility principle we discussed recently.

However, If you decide to keep the current implementation, I am fine with merge. CheckWhiteList seems to me as good step for changing the set of used checkers. Good job!

Rikkola:
I decided to split `DetectRedundantActionCheck`, but afraid to do it for `DetectRedundantRowsCheck`.

`DetectRedundantRowsCheck` is the most resource consuming check we have. I would keep cleaning it up on my TODO list, but will do it when I also have time to do performance improvements and run JProfiler on the side. So for now, I'll keep it as it is.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/PriorityListCheck.java
```diff
@@ -0,0 +1,59 @@
...
+import org.drools.workbench.services.verifier.api.client.configuration.CheckWhiteList;
+import org.drools.workbench.services.verifier.api.client.reporting.Issue;
+
+/**
+ * Wraps more than one check into one.
+ * Each check will look for failure in the given order.
+ * Once failure is found the rest of the checks are ignored.
+ * <br>
+ * <br>
+ * √This is used for example by the conflict-subsubsumption-redundancy chain.
```
manstis: Silly point, but what is the square root of "This"? ;-)

manstis:
@Rikkola Was you planning on adding a UI/User Preferences (see @paulovmr Preferences API) to allow Users to dis/enable the different checks? (I bet it's a sub-task on a JIRA somewhere... but just asking to be sure)

Rikkola:
@manstis This just covers the needs for hit policies. Trying to do the least possible amount of work. Based on the hit policy that is selected, I will change this white list.

There is a separate ticket for the feature where the user can change these settings, but need to see when and if there is time for it.

Rikkola:
But need to check the preferences API, might be a one day thing then. UI is what takes time.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/PriorityListCheck.java
Rikkola: hehe

Rikkola:
@manstis So I think QA ( and you ) took a look. If your happy this is ready to be merged.

PR #367, opened 2016-12-13
GUVNOR-2784: [Guided Decision Table] Changes in individual decision tables doesn't reflect into guided decision table graph

manstis:
See https://issues.jboss.org/browse/GUVNOR-2784

The issue was that the links in a graph point to versions of the tables; to support opening "old" versions of the graph (that should then show the older versions of the tables). However when updating a single table (not a graph) the links in the graph file still point to the original versions. This PR updates the graphs' links when a single table is saved.


manstis:
Jenkins please retest this.

manstis:
Hi @Rikkola @karreiro Could you please review when you get a moment free.

PR #372, opened 2017-01-02
Upgrade JPA (2.0 -> 2.1) + Hibernate (4.x -> 5.x)

psiroky:
Please don't merge yet. This should be merged with all the other IP BOM PRs.

### DIFF
--- a/drools-wb-jcr2vfs-migration/drools-wb-jcr2vfs-import/pom.xml
+++ b/drools-wb-jcr2vfs-migration/drools-wb-jcr2vfs-import/pom.xml
@@ -180,6 +180,11 @@
       <artifactId>zip4j</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- Investigate why is gwt-user needed and ideally remove it, see https://issues.jboss.org/browse/DROOLS-1391 -->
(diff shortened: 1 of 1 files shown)


psiroky:
@manstis @Rikkola could you please quick review this?

drools-wb-jcr2vfs-migration/drools-wb-jcr2vfs-import/pom.xml
```diff
@@ -180,6 +180,11 @@
       <artifactId>zip4j</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- Investigate why is gwt-user needed and ideally remove it, see https://issues.jboss.org/browse/DROOLS-1391 -->
```
Rikkola: Could be because we use some legacy bits from 6.x and they depend on GWT.

psiroky: What's weird is that this is only needed when using JPA 2.1 + Hibernate 5.x. With the old JPA 2.0 + Hibernate 4.x it was working alright.

PR #373, opened 2017-01-03
GUVNOR-2773 : [Guided Decision Table] ValidationWebWorkerServlet fails to load web worker

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2773

Depends on:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/343

Fixing this by merging drools-wb-verifier-web-worker and drools-wb-verifier-backend. -backend only had one class. I originally followed the approach where client side GWT generation and backend are in different files, but since this is not a "real" GWT application that produces a .war having both the servlet and code generation in one module should be ok.


jomarko:
Hi @Rikkola . Thanks for the fix. Just question about the fix. The problem was that maven-resources-plugin didn't copied the resources from drools-wb-verifier-backend into drools-wb-verifier-webworker during copy-resources goal?

Rikkola:
@jomarko The problem was probably that the drools-wb-verifier-backend and drools-wb-verifier-web-worker were not build in the correct order or by the same slave. Not entirely sure how the build servers did this.
But because of this the resource that maven-resources-plugin was trying to copy was not where it was supposed to be.

The thing is, can't really test this locally. It has always worked in my local builds. Now that it is merged the next builds will show if this actually was the issue and this fixes the issue.

PR #374, opened 2017-01-03
GUVNOR-2814: Guided Decision Table Graph: User is always prompted to save or discard data

manstis:
See https://issues.jboss.org/browse/GUVNOR-2814

This PR ensures the "original" ```hashCode``` is correctly set when a graph (or individual table) is loaded. It does however highlight the status of https://issues.jboss.org/browse/GUVNOR-2262 that is related to the JIRA for this PR in that you are no longer prompted when closing a "dirty" guided decision table as the underlying model lacks a ```hashCode``` implementation (other than the default provided by the JVM).

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenter.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenter.java
@@ -78,6 +78,8 @@ import org.uberfire.lifecycle.OnClose;
 import org.uberfire.lifecycle.OnFocus;
 import org.uberfire.lifecycle.OnMayClose;
 import org.uberfire.lifecycle.OnStartup;
+import org.uberfire.mvp.Command;
+import org.uberfire.mvp.ParameterizedCommand;
 import org.uberfire.mvp.PlaceRequest;
 import org.uberfire.mvp.impl.PathPlaceRequest;
 import org.uberfire.workbench.events.NotificationEvent;
@@ -98,7 +100,6 @@ public class GuidedDecisionTableGraphEditorPresenter extends BaseGuidedDecisionT
     private final Event<SaveInProgressEvent> saveInProgressEvent;
     private final LockManager lockManager;
 
-    private Integer originalGraphHash;
     private GuidedDecisionTableEditorGraphContent content;
     private LoadGraphLatch loadGraphLatch = null;
     private SaveGraphLatch saveGraphLatch = null;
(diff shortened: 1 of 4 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenter.java
```diff
@@ -108,28 +109,41 @@
...
+        private LoadGraphLatch( final int dtGraphElementCount,
+                                final ParameterizedCommand<GuidedDecisionTableView.Presenter> onDocumentGraphEntryLoadedCommand,
+                                final Command onAllDocumentGraphEntriesLoadedCommand ) {
             this.dtGraphElementCount = dtGraphElementCount;
-            this.dtToSelectPath = dtToSelectPath;
+            this.onDocumentGraphEntryLoadedCommand = onDocumentGraphEntryLoadedCommand;
+            this.onAllDocumentGraphEntriesLoadedCommand = onAllDocumentGraphEntriesLoadedCommand;
         }
 
-        private void fireDecisionTableSelectedEvent( final GuidedDecisionTableView.Presenter dtPresenter ) {
```
jomarko: Hello @manstis . Thanks for your fix in such short time. I have just short question. I see you deleted the method `fireDecisionTableSelectedEvent`, however you added the same code twice in your PR.

https://github.com/manstis/drools-wb/blob/3a5cf506f20a302a819d246c173d87a0cbbf9fee/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenter.java#L446

https://github.com/manstis/drools-wb/blob/3a5cf506f20a302a819d246c173d87a0cbbf9fee/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenter.java#L574

Is it intended? Couldn't be worth to keep the private method? Since there are two occurrences of the same code.

What do you think?

manstis: Hi @jomarko I changed the ```LoadGraphLatch``` class to invoke callbacks when a graph entry (i.e. Decision Table) was loaded and when all entries had been loaded. This allowed me to set the "original graph hash" after the graph was loaded. I have however changed this PR to share a private method that builds the command to select a graph entry (Decision Table).

manstis:
@Rikkola @karreiro Could you please review at your leisure?

karreiro:
Thanks @manstis :-)

PR #379, opened 2017-01-10
GUVNOR-2819 : [Guided Decision Table] Restore to previous version throws exception

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2819


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/controller/AnalyzerControllerImplTest.java
```diff
@@ -76,6 +106,50 @@ public void areHandlersAreSetUp() throws Exception {
     }
 
     @Test
+    public void areHandlersAreTornDownOnTerminate() throws
```
manstis: If I was to be picky I'd say this test name should more correctly be "areHandlersTornDownOnTerminate"

Rikkola:
I'll update it. Best to fix anything that causes doubt before merging.

Rikkola:
Fixed.

PR #383, opened 2017-01-13
DMN style Hit Policies for Decision Tables

Rikkola:
This adds several features listed in the tickets below. 
Existing Verification checks are disabled if they make no sense for the selected hit policy.

This commit gives us DMN style Hit Policies. What is still missing is Verifier managing overlaps, salience and activation-group. This doesn't mean it is possible to make a table that has a Hit Policy and does not work like the Hit Policy should work. It just means that the user is more likely to add lines that can never fire or can be removed. 

This now depends on https://github.com/droolsjbpm/drools/pull/1084

> 
> Guided Decision Tables: Add hit policy selection into dtable creation
> https://issues.jboss.org/browse/GUVNOR-2774
> 
> Guided Decision Table V&V - Hit policy: None
> https://issues.jboss.org/browse/GUVNOR-2650
> 
> Guided Decision Tables: Prevent the use of salience and agenda/activation-group if the selected hit mode requires them
> https://issues.jboss.org/browse/GUVNOR-2785
> 
> Guided Decision Table: Show current hit policy in the table
> https://issues.jboss.org/browse/GUVNOR-2798


Rikkola:
Looks like the tests pass. 
@manstis This could use a review please ( and/or anyone else who is interested).
Not in a hurry working on top of this.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorView.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.drools.workbench.screens.guided.dtable.client.handlers;
+
+import com.google.gwt.user.client.ui.IsWidget;
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+
+public interface HitPolicySelectorView
+        extends IsWidget {
+
+    void setSelection( final String name );
```
manstis: It's inconsistent that ```addHitPolicyOption(...)``` and ```getSelectedPolicy()``` use ```GuidedDecisionTable52.HitPolicy``` whereas ```setSelected(...)``` passes a ```String```. Can this not be ```GuidedDecisionTable52.HitPolicy``` too?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/GuidedDecisionTableOptions.java
```diff
@@ -41,11 +42,15 @@
...
 
+    @UiField(provided = true)
+    HitPolicySelector hitPolicySelector;
+
     private boolean isUsingWizard = false;
 
     private GuidedDecisionTable52.TableFormat tableFormat = GuidedDecisionTable52.TableFormat.EXTENDED_ENTRY;
 
     public GuidedDecisionTableOptions() {
+        hitPolicySelector = new HitPolicySelector( IOC.getBeanManager().lookupBean( HitPolicySelectorView.class ).getInstance() );
```
manstis: ```GuidedDecisionTableOptions``` is injected into ```NewGuidedDecisionTableHandler``` and hence you should not need to instantiate the View in this way. (i.e. you should be able to inject it too). At the very least you'll need to handle destruction of the view instance somewhere/somehow.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.enterprise.context.Dependent;
+
+import com.google.gwt.user.client.ui.Composite;
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+import org.drools.workbench.screens.guided.dtable.client.resources.HitPolicyInternationalizer;
+import org.gwtbootstrap3.extras.select.client.ui.Option;
+import org.gwtbootstrap3.extras.select.client.ui.Select;
+
+@Dependent
+public class HitPolicySelectorViewImpl
```
manstis: New Views should really use ```errai-ui``` and be ```@Templated```.. out of interest is ```HitPolicySelectorViewImpl.html``` actually used!??! It looks like you was going the (Errai) "templated" way but then decided to just construct the View programmatically!?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/configuration/CheckConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+                .addAll( Arrays.asList( CheckType.values() ) );
+
+        return checkConfiguration;
+    }
+
+    public static CheckConfiguration newEmpty() {
+        return new CheckConfiguration();
+    }
+
+    public Set<CheckType> getCheckWhiteList() {
```
manstis: You've renamed ```CheckWhiteList``` to ```CheckConfiguration``` elsewhere; should this not therefore be ```getCheckConfiguration()```?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-backend/src/main/java/org/drools/workbench/services/verifier/webworker/client/CheckWhiteListProvider.java
```diff
@@ -0,0 +1,46 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.drools.workbench.services.verifier.webworker.client;
+
+import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
+import org.drools.workbench.services.verifier.api.client.configuration.CheckConfiguration;
+import org.drools.workbench.services.verifier.api.client.reporting.CheckType;
+
+public class CheckWhiteListProvider {
```
manstis: Should this really be ```CheckConfigurationProvider```?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/DetectImpossibleMatchCheck.java
```diff
@@ -57,28 +62,56 @@ public void check() {
...
+        final Optional<Field> field = getField();
+        if ( field.isPresent() ) {
+            return field.get()
+                    .getFactType();
+        } else {
+            return "";
+        }
+    }
+
+    private Optional<Field> getField() {
```
manstis: Sweet use of ```Optional``` :-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorView.java
Rikkola: It should. I changed away from the string when I added i18n. Missed this spot.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
Rikkola: The html file is a dead file. And I think it went like you said ( did this part a month ago ), first I planned to do the template and since this ended up being just a wrapper for a single widget I dropped it. 
I guess if we are dropping the widgets, I need to use the html file.

Rikkola: And now that I'm starting to edit this, I remember. I tried to get the selection events and adding new items into the dropbox in the template to work and after some time went with the widget that too a second to set up. I'll check if I'm smarter this year.

Rikkola: Good news, I am smarter ;-)

Rikkola:
Going to also change the First to Reserved hit. And now that I think about this. This PR should not be accepted before I have updated docs.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorTest.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    public void getActiveHitPolicy() throws
+                                     Exception {
+//        when( view.getSelectedPolicy() ).thenReturn( GuidedDecisionTable52.HitPolicy.FIRST_HIT );
+//        assertEquals( GuidedDecisionTable52.HitPolicy.FIRST_HIT,
+//                      selector.getHitPolicy() );
+
+    }
```
jomarko: Hi @Rikkola please could you finish this test?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/DetectConflictingRowsCheck.java
```diff
@@ -47,15 +58,16 @@ public void check() {
     }
 
     @Override
-    public boolean isActive( final CheckWhiteList whiteList ) {
-        return whiteList.getAllowedCheckTypes()
-                .contains( CHECK_TYPE );
+    public boolean isActive( final CheckConfiguration whiteList ) {
```
jomarko: `WhiteList` class was moved to `CheckConfiguration` class, is it intended to keep attribute names (`whiteList`) according to old class name?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
```diff
@@ -787,11 +788,19 @@ public void newAttributeOrMetaDataColumn() {
         if ( isReadOnly() ) {
             return;
         }
-        view.newAttributeOrMetaDataColumn();
+        view.newAttributeOrMetaDataColumn( getReservedAttributeNames() );
```
jomarko: Shouldn't be this `view.newAttributeOrMetaDataColumn( getExistingAttributeNames() );`?

I am asking just because of the signature: `newAttributeOrMetaDataColumn( final Set<String> existingAttributeNames );` in: `GuidedDecisionTableView.java.`

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/DetectConflictingRowsCheck.java
Rikkola: Nope that needs to go. It can end up being a lot more than just a white list.

Rikkola: it = `CheckConfiguration`

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/NewGuidedDecisionTableHandler.java
```diff
@@ -68,15 +67,13 @@ public NewGuidedDecisionTableHandler( final PlaceManager placeManager,
                                           final GuidedDTableResourceType resourceType,
                                           final GuidedDecisionTableOptions options,
                                           final BusyIndicatorView busyIndicatorView,
-                                          final NewGuidedDecisionTableWizardHelper helper,
-                                          final SyncBeanManager iocManager ) {
+                                          final NewGuidedDecisionTableWizardHelper helper) {
```
jomarko: Very minor formatting issue.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/NewGuidedDecisionTableHandler.java
```diff
@@ -107,20 +104,18 @@ public void create( final Package pkg,
...
-                                           baseFileName,
-                                           options.getTableFormat() );
+                                           baseFileName );
         }
     }
 
     private void createEmptyDecisionTable( final Path contextPath,
-                                           final String baseFileName,
-                                           final GuidedDecisionTable52.TableFormat tableFormat ) {
+                                           final String baseFileName) {
```
jomarko: Very minor formatting issue.

jomarko:
Hi @Rikkola . I put some comments to this PR, however in general I think it is another your fine PR with good code coverage. Thanks.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
Rikkola: Actually the other way around. `existingAttributeNames` should be `reservedAttributeNames`.  This is because reserved has both, the attribute names we block and the ones that already exist.
Still, well spotted.

Rikkola:
Thank you for the review @jomarko . Made the required changes. 

Currently doing the docs. Hit Policies might stay as they are so not working on those.

Rikkola:
@manstis @jomarko 
Ok did the fixes and verified that these changes fit in with the latest Hit Policy specification. This still lacks the Hit Policy Resolved, but that can be done as a separate patch.

Also rebased with the latest master. Tests are still running.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/GuidedDecisionTableOptions.java
```diff
@@ -57,6 +67,15 @@ public boolean isUsingWizard() {
         return this.tableFormat;
     }
 
+    public GuidedDecisionTable52.HitPolicy getHitPolicy() {
+        return hitPolicy;
+    }
+
+    @UiHandler(value = "hitPolicySelector")
+    void chkUseWizardClick( ValueChangeEvent<GuidedDecisionTable52.HitPolicy> valueChangeEvent ) {
```
manstis: Are you confident of this method name?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
```diff
@@ -0,0 +1,84 @@
...
+import org.jboss.errai.common.client.dom.Node;
+import org.jboss.errai.common.client.dom.Span;
+import org.jboss.errai.common.client.dom.UnorderedList;
+import org.jboss.errai.ui.shared.api.annotations.DataField;
+import org.jboss.errai.ui.shared.api.annotations.Templated;
+
+@Dependent
+@Templated
+public class HitPolicySelectorViewImpl
+        extends Composite
```
manstis: Oeee... what a mish-mash of types.. extending ```Widget``` but containing ```Element``` children..

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    UnorderedList listOfAvailableHitModes;
+
+    @Override
+    public void setSelection( final GuidedDecisionTable52.HitPolicy hitPolicy ) {
+        selectedValue.setTextContent( HitPolicyInternationalizer.internationalize( hitPolicy ) );
+    }
+
+    @Override
+    public void addHitPolicyOption( final GuidedDecisionTable52.HitPolicy hitPolicy ) {
+        final AnchorElement anchor = Document.get()
```
manstis: I'm not saying there is; I'm not saying there isn't.. but this type of "view" construction makes use of GWT's ```Panel``` and ```Composite``` seem really nice.. Might be worth asking @mbarkley or @csadilek if there's a better way (other than writing our own element-based widget library).

Rikkola: Hehe, I remember a time when making a dropdown was 1-3 lines and did not take 5 files and +100 lines of code (but not making the mistake of saying everything was better back then, one of those files is the test for example).

I have to say I am not convinced widget libraries are a bad thing. I guess the goal is to have the HTML handed over to us so we save time with UI. In the end we only get the huge changes as HTML and then end up spending more time than we save on the small daily changes like this. Also not thinking Errai UI is a bad thing, just not using Widgets adds a lot of styling and setups.

There is also more functions that can break here, when comparing to `new Dropdown()`. I can get the wrong styling ( I did ), leave out the event firing ( did that too ), but this was the first time I did this so that partially explains it.

Got this bit from the Uberfire examples, but maybe there is a better way and the examples need an update.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
Rikkola: I'll check how it is done in other location. Got this from the Errai docs.

Rikkola: And looks like that was not the latest. Just Googled for Errai UI, should have used the web site.

Rikkola: And even the latest still uses Composite.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
manstis: @Rikkola Their examples don't extend anything; http://docs.jboss.org/errai/4.0.0.Beta2/errai/reference/html_single/ (see their ```LoginForm``` example, amongst others)

Rikkola: Once you scroll down to EventHandlers they do, but then again you need the `Composite` there for GWT event handlers. Currently extending nothing and working around that with the Uberfire Callback.

manstis:
@Rikkola @jomarko Can this be merged now?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
jomarko: @Rikkola at this place, I am fine with your solution, I am just interested if you had also tried implementing this [1] interface in `HitPolicySelectorViewImpl`, what could make the handler registration more simple. I am just asking, not sure if there wouldn't be needed more coding than in your solution.

[1]
http://www.gwtproject.org/javadoc/latest/com/google/gwt/event/logical/shared/HasValueChangeHandlers.html

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/DetectConflictingRowsCheck.java
```diff
@@ -47,15 +58,16 @@ public void check() {
     }
 
     @Override
-    public boolean isActive( final CheckWhiteList whiteList ) {
-        return whiteList.getAllowedCheckTypes()
-                .contains( CHECK_TYPE );
+    public boolean isActive( final CheckConfiguration checkConfiguration ) {
```
jomarko: The implementation of `isActive` method is the same in 3/5 occurrences. Maybe it would be worth to extract it into some parent class? What do you think @Rikkola ?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/configuration/AnalyzerConfiguration.java
```diff
@@ -27,22 +27,22 @@
...
                                   final RunnerType runnerType ) {
         this.webWorkerUUID = PortablePreconditions.checkNotNull( "webWorkerUUID",
                                                                  webWorkerUUID );
         this.dateTimeFormatter = PortablePreconditions.checkNotNull( "dateTimeFormatter",
                                                                      dateTimeFormatter );
         this.uuidKeyProvider = PortablePreconditions.checkNotNull( "uuidKeyProvider",
                                                                    uuidKeyProvider );
-        this.checkWhiteList = PortablePreconditions.checkNotNull( "checkWhiteList",
-                                                                  checkWhiteList );
+        this.checkConfiguration = PortablePreconditions.checkNotNull( "checkWhiteList",
```
jomarko: Shouldn't be similar occurrences of `checkWhiteList` replaced by `checkConfiguration`?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/configuration/AnalyzerConfiguration.java
```diff
@@ -59,8 +59,8 @@ public String formatDate( final Date dateValue ) {
         return dateTimeFormatter.format( dateValue );
     }
 
-    public CheckWhiteList getCheckWhiteList() {
-        return checkWhiteList;
+    public CheckConfiguration getCheckWhiteList() {
```
jomarko: Here same as above?

jomarko:
Hi @Rikkola, @manstis . I did the review right now. I put some minor things into comments, however I didn't find any other problem that could prevent from merging.

manstis:
@Rikkola @jomarko OK, just let me know when you want it merged.

@Rikkola IDK if you was going to make any of @jomarko change suggestions.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
Rikkola: @jomarko Late reply. 
That is the easy implementation, but setting it up gets messy unless you can use GWT Component and we are trying to get rid of GWT Components and Widgets since the future GWT versions will not support them. The new way is to use Elements. This however makes life a bit more complicated, at least right now when we are in transition phase. One example is this implementation that would have been a quick thing with the old way, but not this should probably be made into a reusable widget somewhere in our code base.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/DetectConflictingRowsCheck.java
Rikkola: Moved this down to ChecBase.

Rikkola:
@jomarko @manstis This and the one in drools could use a second review. Here the new changes are in the second commit. In drools the HP resolving has completely changed.

The second commit depends on this. https://github.com/droolsjbpm/drools/pull/1084

manstis:
Jenkins please retest this.

manstis:
@Rikkola branch names for ```drools``` and ```drools-wb``` PRs are different so Jenkins cannot correlate the two.. I suspect that is leading to the compilation error on this PR :-(

Rikkola:
@manstis Yes, had to either do it this way or open yet another PR. But the one in drools can be merged individually and then this one should pass.

manstis:
Jenkins please retest this.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/converters/column/impl/MetaDataColumnConverterTest.java
```diff
@@ -0,0 +1,121 @@
...
+    public void setUp() throws
+                        Exception {
+        metaDataColumnConverter = new MetaDataColumnConverter();
+    }
+
+    @Test
+    public void handles() throws
+                          Exception {
+        assertTrue( metaDataColumnConverter.handles( new MetadataCol52() ) );
+        assertFalse( metaDataColumnConverter.handles( new MetaDataColumnConverterCantTouchThisBreakItDownStopHammerTime() ) );
```
manstis: You are clearly following corporate "Enterprise naming" guidelines :-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/handlers/HitPolicySelectorViewImpl.java
jomarko: @Rikkola late reply but reasonable reply. Thanks for explanation.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/ModelSynchronizerImpl.java
```diff
@@ -277,7 +282,7 @@ public void deleteColumn( final BaseColumn column ) throws MoveColumnVetoExcepti
                 final boolean isSalienceUseReverseOrderUpdated = BaseColumnFieldDiffImpl.hasChanged( AttributeCol52.FIELD_REVERSE_ORDER,
                                                                                                      diffs );
                 if ( isSalienceUseRowNumberUpdated || isSalienceUseReverseOrderUpdated ) {
-                    updateSystemControlledColumnValues();
```
jomarko: In fact `updateSystemControlledColumnValues()` calls `systemControlledColumnValuesSynchronizer.updateSystemControlledColumnValues();`. It can suggest that `updateSystemControlledColumnValues()` is no more needed? However it can also suggest nothing. Just my notice.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/PairCheck.java
```diff
@@ -16,6 +16,7 @@
 package org.drools.workbench.services.verifier.core.checks.base;
 
 import org.drools.workbench.services.verifier.api.client.configuration.AnalyzerConfiguration;
+import org.drools.workbench.services.verifier.api.client.configuration.CheckConfiguration;
```
jomarko: Unused import I think.

PR #384, opened 2017-01-13
GUVNOR-2848 : [Guided Decision Table] After context change appending of row throws exception

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2848

Each editor focus added event listeners.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/controller/Events.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/controller/Events.java
@@ -33,7 +33,6 @@ public class Events {
 
     private Set<HandlerRegistration> eventBusHandlerRegistrations = new HashSet<>();
 
-
     public Events( final EventBus eventBus,
                    final AnalyzerControllerImpl analyzerController ) {
         this.eventBus = PortablePreconditions.checkNotNull( "eventBus",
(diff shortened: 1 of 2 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/controller/Events.java
```diff
@@ -43,21 +43,23 @@ public Events( final EventBus eventBus,
...
-                                                               analyzerController ) );
-        eventBusHandlerRegistrations.add( eventBus.addHandler( UpdateColumnDataEvent.TYPE,
-                                                               analyzerController ) );
-        eventBusHandlerRegistrations.add( eventBus.addHandler( AppendRowEvent.TYPE,
-                                                               analyzerController ) );
-        eventBusHandlerRegistrations.add( eventBus.addHandler( InsertRowEvent.TYPE,
-                                                               analyzerController ) );
-        eventBusHandlerRegistrations.add( eventBus.addHandler( AfterColumnInserted.TYPE,
-                                                               analyzerController ) );
+        if ( canSetUp ) {
```
manstis: Yeah this works doesn't it... it may have been nicer to record the ```HandlerRegistration``` from the ```add(...)``` method and only reattach the handler if the registration is ```null```.. still your change works..

Rikkola: Right, I can set if the set is empty and remove this boolean. Good point, the day has been too long. Will fix this.

Rikkola:
Updated and removed the dirty boolean.

PR #385, opened 2017-01-17
Don't package server marshaller classes in jar.

mbarkley:
Removing the generated Errai server marshallers from drools-wb-verifier-backend jar fixes NPEs that several people have experienced after logging into drools-wb in SDM from missing marshaller definitions.

@manstis can you please review this? I'm not exactly sure what drools-wb-verifier-backend does or how to test that it still works.


manstis:
@pefernan Max has fixed it for us. Yay!

Rikkola:
@mbarkley drools-wb-verifier-backend uses GWT to generate a JavaScript file that we use as a web worker in the main client application.

pefernan:
@manstis +1 for Max ;)

PR #389, opened 2017-01-20
GUVNOR-2741: Convert templaye key into column header metadata

jomarko:


jomarko:
screenshot from 2017-01-20 15-43-23
screenshot from 2017-01-20 15-42-55

manstis:
@jomarko Related (and IIRC we removed it a while ago); now we have a second row in the header for ```BRLConditionVariableColumn``` wouldn't it be an idea to merge the first row header? This should be achievable by using the same (```equals(..)```) ```BaseHeaderMetaData``` for the first row. Also, can you advise if "(BRL) action" columns need to have their converter improved too?

jomarko:
@manstis Merging the first row of the header is definitely good idea. I will definitely do it. I could also have a look on BRL actions. Thanks for pointing this out.

PR #390, opened 2017-01-21
GUVNOR-2859: Enhance decision table column headers

jomarko:
This PR should:
* Merge first row of _BRL Condiiton_ column header if there are multiple template keys
* Render second row of _BRL Action_ column header if the template key is used
* Merge first row of _BRL Action_ column header if there are multiple template keys
* Render two rows for _Insert Fact_ and _Set Field Value_ columns header

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/converters/column/impl/ActionSetFieldColumnConverter.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/converters/column/impl/ActionSetFieldColumnConverter.java
@@ -75,8 +75,18 @@ public class ActionSetFieldColumnConverter extends BaseColumnConverterImpl {
     @Override
     public List<GridColumn.HeaderMetaData> makeHeaderMetaData( final BaseColumn column ) {
         return new ArrayList<GridColumn.HeaderMetaData>() {{
-            add( new BaseHeaderMetaData( column.getHeader(),
-                                         ActionCol52.class.getName() ) );
+            if ( column instanceof ActionSetFieldCol52 ) {
+                ActionSetFieldCol52 actionSetFieldColumn = ( ActionSetFieldCol52 ) column;
+                if ( actionSetFieldColumn.getBoundName() != null && !actionSetFieldColumn.getBoundName().isEmpty() ) {
+                    add( new BaseHeaderMetaData( actionSetFieldColumn.getBoundName(),
+                                                 ActionCol52.class.getName() ) );
+                    add( new BaseHeaderMetaData( column.getHeader(),
+                                                 actionSetFieldColumn.getBoundName() ) );
+                }
+            } else {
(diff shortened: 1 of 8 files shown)


jomarko:
@manstis Hi, your concern was reasonable. I need to fix this. I will let you know when I am done.

jomarko:
@manstis Hi I updated the PR according our discussion in IRC. However, here is the current situation:
* Swapping two Simple Conditions of different pattern - works
* Swapping two Simple Conditions of same pattern - works
* Swapping BRL Condition and Simple Condition - not works (same without my PR)
* Swapping Template Keys of BRL Condition - not works (same without my PR)
* Swapping two Simple Actions of different pattern - not work
* Swapping two Simple Actions of same pattern - works
* Swapping BRL Action and Simple Action - not works (same without my PR)
* Swapping Template Keys of BRL Action - not works (same without my PR)

Please, do you have any ideas why swapping of two simple actions doesn't work after my PR?

manstis:
@jomarko Hi,

I'd be more alarmed that "Swapping two Simple Conditions of different pattern - works" was found. You should not be able to move a "condition" from one "pattern" to another. So IMO the "Swapping two Simple Actions of different pattern - not works" is OK, but "Swapping two Simple Conditions of different pattern - works" is NOT OK!

jomarko:
@manstis sorry, I was not clear in the comment. I updated the comment now.

manstis:
@jomarko As discussed; this PR fixes the intent of the associated JIRA and can therefore be merged. Could you however please open another JIRA (assigned to me) to look into the apparent DnD issues (as this may highlight a problem elsewhere). Thank-you for all your help.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/converters/column/impl/ActionSetFieldColumnConverter.java
```diff
@@ -75,8 +75,18 @@ public boolean handles( final BaseColumn column ) {
     @Override
     public List<GridColumn.HeaderMetaData> makeHeaderMetaData( final BaseColumn column ) {
         return new ArrayList<GridColumn.HeaderMetaData>() {{
-            add( new BaseHeaderMetaData( column.getHeader(),
-                                         ActionCol52.class.getName() ) );
+            if ( column instanceof ActionSetFieldCol52 ) {
+                ActionSetFieldCol52 actionSetFieldColumn = ( ActionSetFieldCol52 ) column;
+                if ( actionSetFieldColumn.getBoundName() != null && !actionSetFieldColumn.getBoundName().isEmpty() ) {
```
manstis: This should never be false; but if it is you probably want to include the second-level ```BaseHeaderMetaData``` entry within the ```if(..) {..}``` block.

jomarko: Amended, thanks.

jomarko:
@manstis the jira for the remaining problem is here https://issues.jboss.org/browse/GUVNOR-2866

PR #394, opened 2017-01-31
GUVNOR-2873: Guided Decision Table Editor: Change Edit, Insert and View menus to use MenuItems

manstis:
See https://issues.jboss.org/browse/GUVNOR-2873


manstis:
Part of an ensemble:-

- https://github.com/uberfire/uberfire/pull/617
- https://github.com/droolsjbpm/drools-wb/pull/394

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/editor/menu/InsertMenuBuilderTest.java
```diff
@@ -66,213 +77,189 @@
...
+            add(new DTCellValue52(1));
+            add(new DTCellValue52("descr"));
+            add(new DTCellValue52("md"));
+        }});
+
+        uiModel.selectCells(0,
+                            2,
+                            1,
+                            2);
+
```
csadilek: I really didn't want to chime but this is an example where our coding convention gets absurd. These are constants and there's no debugging or readability advantage. The latter might be subjective so nothing to be done here...

PR #395, opened 2017-02-09
Ensure LogoNavBar is top-most

manstis:
This PR ensures the ```LogoNavBar``` is top-most by moving the ```AppNavBar``` down one in order (both had an order of ```Integer.MAX_VALUE``` and hence their runtime order could be indeterminate).


csadilek:
+1

PR #396, opened 2017-02-10
[GUVNOR-2880] remove unused field causing issues on Tomcat

psiroky:
The biggest issue with the User field was that it was @RequestScoped.
The ScenarioRunnerService is being used as part of AsyncJobs (e.g.
when the project/maven/test REST endpoint is triggered). On Tomcat
this does not work, becuase the request thread is different from the
processing thread and thus the request scoped variables are not
available. For some reason this works on WildFly 10 / EAP7
(most likely because we use the async EJBs there, instead of the
thred pool).

@Rikkola could you please review?


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/main/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerService.java
```diff
@@ -146,7 +143,7 @@ public void testAssumptionFailure(Failure failure) {
 
         testResultMessageEvent.fire(
                 new TestResultMessage(
-                        identity.getIdentifier(),
+                        TEST_SCENARIO_SERVICE_ID,
```
psiroky: I am not exactly sure about this. Since the `TestResultMessage` is directly in guvnor, it seems like something more generic, so not sure if removing the identity from there is alright.

Rikkola:
@psiroky Sorry, took a while. Looks like I left the review open and didn't click the Approve button like I meant to.

PR #397, opened 2017-02-15
GUVNOR-2866: [Guided Decision Table] Columns swaping doesn't work in all cases

manstis:
See https://issues.jboss.org/browse/GUVNOR-2866

There actually wasn't any support for DnD'ing "BRL" columns.


manstis:
@Rikkola @karreiro Would you mind reviewing please, when you get a chance?

karreiro:
@manstis I think that there are some formatting issues.

manstis:
@karreiro Unfortunately there are no formatting issues; there is a new "code style formatter" in ```droolsjbpm-build-bootstrap``` specifically you could read [this](https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/358) and the referenced PRs.. I did ask for it to be communicated to BSIG, but it looks like it never happened :-(

karreiro:
@manstis thanks! :-)

PR #399, opened 2017-02-17
GUVNOR-2827: [Guided Decision Table] Not possible to create two conditions based on the same negated pattern

manstis:
See https://issues.jboss.org/browse/GUVNOR-2827

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopup.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopup.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 import com.google.gwt.dom.client.Style;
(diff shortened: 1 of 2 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopup.java
```diff
@@ -435,140 +434,142 @@ ListBox loadPatterns() {
...
+        pop.addAttribute(GuidedDecisionTableConstants.INSTANCE.ChooseExistingPatternToAddColumnTo(),
+                         pats);
 
-        pop.add( new ModalFooterChangePattern( new Command() {
+        pop.add(new ModalFooterChangePattern(new Command() {
             @Override
             public void execute() {
-                String[] val = pats.getValue( pats.getSelectedIndex() ).split( "\\s" );
-                editingPattern = model.getConditionPattern( val[ 1 ] );
+                String[] val = pats.getValue(pats.getSelectedIndex()).split("\\s");
```
manstis: Here's the change!

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopup.java
```diff
@@ -578,139 +579,149 @@ public void execute() {
...
+                                                 }
+                                             }
+        ));
 
-        view.enableFooter( false );
+        view.enableFooter(false);
         pop.show();
     }
 
+    Optional<Pattern52> extractEditingPattern(final String[] metadata) {
```
manstis: ..and here.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopupTest.java
```diff
@@ -89,343 +90,420 @@ public void setUp() throws Exception {
...
+                   column,
+                   true,
+                   false);
 
         popup.applyChanges();
-        assertNull( popup.getEditingCol().getBinding() );
+        assertNull(popup.getEditingCol().getBinding());
     }
 
+    @Test
```
manstis: New tests.

manstis:
@Rikkola @karreiro I've commented where the actual changes are.. otherwise it's all formatting.

manstis:
Part of an ensemble:-

- https://github.com/droolsjbpm/drools/pull/1105
- https://github.com/droolsjbpm/drools-wb/pull/399

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopup.java
Rikkola: Ok, I'm learning new things about optional. This can come in handy.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopupTest.java
Rikkola: Is there a test for that checks the IllegalStateException? Would be good to check that situation.

Rikkola:
Would just like to check if the IllegalStateException is checked with a test.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopupTest.java
```diff
@@ -89,343 +91,450 @@ public void setUp() throws Exception {
...
+    public void extractEditingPatternWithoutNotPatterns() {
+        final String[] metadata = new String[]{"Applicant", "", "true"};
+        final Optional<Pattern52> found = popup.extractEditingPattern(metadata);
+
+        assertFalse(found.isPresent());
+    }
+
+    @Test(expected = IllegalStateException.class)
```
manstis: @Rikkola new test..

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/ConditionPopup.java
```diff
@@ -435,282 +434,293 @@ ListBox loadPatterns() {
...
+    Command getShowPatternOKCommand(final ListBox pats,
+                                    final FormStylePopup pop) {
+        return () -> {
+            String[] val = pats.getValue(pats.getSelectedIndex()).split("\\s");
+            final Optional<Pattern52> p = extractEditingPattern(val);
 
-                //Clear Field and Operator when pattern changes
-                editingCol.setFactField( null );
-                editingCol.setOperator( null );
+            //This should never be thrown; but let's fail fast to make debugging easier
```
manstis: @Rikkola I've also added a comment about ```IllegalStateException``` - it shouldn't ever be thrown as it will suggest something rather more serious has gone wrong; but I thought it better to fail fast than have some weird exception elsewhere.

Rikkola: Feels like the correct way to do this. Most likely there is never a reason for ISE to happen in our app, but best to use it where we have doubt.

PR #401, opened 2017-02-21
GUVNOR-2843: [Guided Decision Table] Template key input hides content partially

manstis:
See https://issues.jboss.org/browse/GUVNOR-2843


manstis:
@Rikkola @karreiro Here's another one if you don't mind... thanks.

PR #407, opened 2017-03-02
[DROOLS-1452] remove no longer needed exclusion

psiroky:
sisu-guice was replaced directly by the standard guice,
which no longer depends on sisu-guava

Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/401


psiroky:
The same change already done as https://github.com/droolsjbpm/drools-wb/commit/65e242fd9bc7c5955f8d993f92318a4cc4141c3e.

PR #408, opened 2017-03-02
GUVNOR-2869 : Guided Decision Table V&V - Fix Verification reporting levels

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2869

Reporting was still a bit too eager, this silences it down. Quite a lot to do on reporting configuration, but given the time I think best just to harden the core.


Rikkola:
@manstis @jomarko Pretty simple, just silences the verification when it is too noisy. Still I could use a review, please.

jomarko:
Hi @Rikkola. PR is completely OK. Maybe just small comment, I am not sure if is the best practice to name test methods like get....... However, I am also for merging.

PR #409, opened 2017-03-03
GUVNOR-2896: Reflect negated pattern in condition column header

jomarko:
This PR adds the prefix 'not ' into the condition column header for
negated patterns.

screenshot from 2017-03-03 08-50-02


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/ConditionColumnSynchronizerTest.java
```diff
@@ -827,7 +912,7 @@ public void testDelete3() throws ModelSynchronizer.MoveColumnVetoException {
     @Test
     public void testMoveColumnTo_MoveLeft() throws ModelSynchronizer.MoveColumnVetoException {
```
manstis: Would it not have been better to create a new test rather than change this one from using a bound name to negated? (Or have I missed something more subtle?)

manstis:
Jenkins please retest this.

jomarko:
Hi @manstis. I incorporated your comment. I also decided to introduce few methods which keep code that was repeatedly used for initialization. Let me know what do you think. Thanks.

manstis:
Jenkins please retest this.

PR #410, opened 2017-03-06
GUVNOR-2924: Introduce tests for the Attribute Column control rows

jomarko:
PR Introduces tests for Attribute columns part of decision table control
header. PR tests if there is created row for each attribute column and if
there are registered proper handlers for actions like removing the
column, hiding the column etc.


jomarko:
Please do NOT merge yet.

Rikkola:
I do like how AttributeColumnConfigRow now is it's own class and tested as a separate unit. Looks fine to me, but letting @manstis take a look.

manstis:
@jomarko Thanks. Ideally I'd like to see all the "peripheral screens" (new column definition, currently moving to a Wizard being written by @karreiro and editing existing columns, to re-use a pre-populated Wizard in the future to be written by @karreiro) including the existing "header" widget (to be redesigned by UX) replaced. I wouldn't want you to worry to unduly on refactoring the existing code in the shorter-term.

PR #411, opened 2017-03-08
Fix enforcer rules during full profile build

cristianonicolai:


cristianonicolai:
@psiroky @manstis similar to https://github.com/droolsjbpm/jbpm-wb/pull/648

PR #412, opened 2017-03-08
GUVNOR-2929: Add tests for hiding columns of the decision table

jomarko:
Extract the check box for hiding used in column PopUps into separate
class. Then test if this class registers appropriate ClickHandler. It
means if the click handler changes the underlying model correctly.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/HideColumnCheckBox.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/HideColumnCheckBox.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/HideColumnCheckBox.java
```diff
@@ -0,0 +1,18 @@
+package org.drools.workbench.screens.guided.dtable.client.widget;
```
manstis: License please.

jomarko:
Hi @manstis . Thanks for the review. I added the license and I also imported the code-style xml into my Idea. However I am not sure how to adapt this commit according the imported code-style, any suggestions?

manstis:
@jomarko Thanks.. as for code formatting; just open each class you've added or changed for this PR and select Code (menu)-> Reformat code (or Ctrl-Alt-L). If "more" is formatted than you changed in a class do not worry; I can review changes ignoring white-space (they have been quite a few PRs recently from other people with code formatting changes).

jomarko:
Pretty cool featrue this (Ctrl-Alt-L). Applied to classes form my PR. Thanks @manstis

manstis:
@jomarko Yes.. and if you highlight a section of code and do it; only the highlighted block is reformatted :-)

PR #413, opened 2017-03-08
PLANNER-704: Workbench: scoreHolder global can only be of org.optaplanner.core.api.score.holder.ScoreHolder type (or subclass)

mcimbora:
* Refactor AddGlobalPopup & GlobalsEditorPresenter

As an addition to the above, save/delete checks have been added. In the long term this will be resolved by https://issues.jboss.org/browse/UF-489.

Part of:
https://github.com/droolsjbpm/guvnor/pull/421
https://github.com/droolsjbpm/drools-wb/pull/413
https://github.com/droolsjbpm/optaplanner-wb/pull/138


mcimbora:
@manstis / @Rikkola Could you please have a look at the PR? Thanks in advance.

drools-wb-screens/drools-wb-globals-editor/drools-wb-globals-editor-backend/src/test/java/org/drools/workbench/screens/globals/backend/server/GlobalsEditorServiceTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+        globalsEditorService.generate( path,
+                                       "test",
+                                       globalsModel,
+                                       "comment" );
+
+        verify( ioService,
+                times( 1 ) ).write( any( org.uberfire.java.nio.file.Path.class ),
+                                    anyString(),
+                                    anyMap(),
```
manstis: I would probably be nice to capture the ```Map``` and check it has the "generated" flag set to ```true```

drools-wb-screens/drools-wb-globals-editor/drools-wb-globals-editor-client/src/test/java/org/drools/workbench/screens/globals/client/editor/GlobalsEditorPresenterTest.java
```diff
@@ -0,0 +1,167 @@
+package org.drools.workbench.screens.globals.client.editor;
```
manstis: License please.

drools-wb-screens/drools-wb-globals-editor/drools-wb-globals-editor-client/src/main/java/org/drools/workbench/screens/globals/client/editor/GlobalsEditorViewImpl.java
```diff
@@ -108,66 +113,64 @@ public String getValue( final Global global ) {
...
-                                final Global global,
-                                final String value ) {
-                if ( isReadOnly ) {
-                    return;
-                }
-                if ( Window.confirm( GlobalsEditorConstants.INSTANCE.promptForRemovalOfGlobal0( global.getAlias() ) ) ) {
-                    dataProvider.getList().remove( index );
-                }
+        deleteGlobalColumn.setFieldUpdater( ( index, global, value ) -> {
+            if ( Window.confirm( translationService.format( GlobalsEditorConstants.GlobalsEditorViewImplPromptForRemovalOfGlobal,
```
manstis: Surely, if the global was generated for OptaPlanner, you'd want to prevent removal from the table too? I check against deletion of the file containing the generated global but not against removal of the definition from the file itself. Have I missed something?

mcimbora: @manstis If the file has been generated, 'remove' button is disable.

manstis: @mcimbora Thanks. I missed that.. now I see it :-)

drools-wb-screens/drools-wb-globals-editor/drools-wb-globals-editor-client/src/main/java/org/drools/workbench/screens/globals/client/editor/AddGlobalPopupView.java
```diff
@@ -0,0 +1,53 @@
...
+public interface AddGlobalPopupView extends UberElement<AddGlobalPopupView.Presenter> {
+
+    interface Presenter {
+
+        void validate();
+
+        void onAddButtonClicked();
+
+        void onCancelButtonClicked();
+    }
```
Rikkola: We might have to refresh the MVP coding conventions to the team at some point. I see views in presenter implementations, presenter interfaces in View interfaces. 

Personally I don't care that much about the above. I think we agreed to drop the Presenter interfaces and I have to say I can't remember where we agreed to put the View. My guess is in a separate class :)

What I do care is the use of "on" on the start of each method that is called from view. So here my suggestion is that you could also add that to the `validate()` method. Just that nobody calls if from anywhere else. Unless that is meant to be called from other presenters/parts of the code than `AddGlobalPopupViewImpl`.

Rikkola:
I think this is something we can reuse at some point. For example many have asked if it would be possible to store Test Scenario test runs.

mcimbora:
@manstis @Rikkola Thanks for a prompt review, PRs updated.

PR #421, opened 2017-03-14
GUVNOR-2934 : [Guided Decision Table] Multiple verification issues affects each other

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2934

@jomarko I added the file and the test `testFileGuvnor2934`. Do you think the unit test pick everything up?


Rikkola:
Extending the tests for this one.

Rikkola:
@jomarko Hi, added the updated file.

Rikkola:
@jomarko Ok, again `MultipleVerificationIssuesFromFile` now has the tests.

Rikkola:
Looks like reporting is missing the redundancy between lines 4 and 5

jomarko:
Hi @Rikkola. According to our discussion on IRC. Could you please close this PR?

Explanation for others: This PR verifies table that contains multiple issues like Redundant Rows, Subsumptant rows, Conflicting rows, Missing Action rows, Missing range rows etc. The problem is that the verification result doesn't contain issues that would Toni and I expect. We realized that it is very hard to start solving this problem because the constructed decision table is kind of big and offers a lot of possibilities where the problem is.

We agreed that we will start on the opposite site. I will start to create similar PR, but just with decision tables, that contains at most 2 kinds of issues. Lets say (Missing range and Redundat rows) or (Redundant rows and Missing action) or (Deficient rows and Missing condition) etc. Once we know that 2 different kinds of issues in the table are revealed successfully we will continue with 3 kinds of issues etc. This could help to point out the root cause.

PR #429, opened 2017-03-24
GUVNOR-2854: Increase coverage for rows insertion, deletion and moving

jomarko:
This PR also fixes two issues:
1. After deletion of a row, the PriorityOver wasn't updated for rows
with higher row number.
2. After insertion of row to position X, the rows that had
PriorityOver X weren't updated to X+1.


jomarko:
@Rikkola please could you have a look when you have a moment? Thanks.

PR #427, opened 2017-03-22
GUVNOR-2934: Verify report of multiple kinds of issues in one table

jomarko:

PR #430, opened 2017-03-27
GUVNOR-2964: Redundancy and Subsumption shown as erros for Unique HP

jomarko:
Issue depends on #427 . Please do not merge yet.

### DIFF
--- a/drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/reporting/Issue.java
+++ b/drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/reporting/Issue.java
@@ -64,4 +64,36 @@ public class Issue {
     public void setDebugMessage( final String debugMessage ) {
         this.debugMessage = debugMessage;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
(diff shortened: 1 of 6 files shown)


drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/reporting/Issue.java
```diff
@@ -64,4 +64,36 @@ public String getDebugMessage() {
...
+        return debugMessage != null ? debugMessage.equals(issue.debugMessage) : issue.debugMessage == null;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = severity != null ? severity.hashCode() : 0;
+        result = 31 * result + (rowNumbers != null ? rowNumbers.hashCode() : 0);
+        result = 31 * result + (checkType != null ? checkType.hashCode() : 0);
+        result = 31 * result + (debugMessage != null ? debugMessage.hashCode() : 0);
+        return result;
```
Rikkola: GWT Client side and hashCode tend to cause trouble, the workaround is to use ~~

Like this:
https://github.com/kiegroup/drools-wb/blob/1fabbdea01b75ac325e97df72d58d87a65d3955d/drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/index/FieldBase.java#L100-L105

jomarko:
Conflicts resolved.

PR #435, opened 2017-04-05
GUVNOR-2962: Reproducer tests for unexpected Missing Range report

jomarko:
This PR adds reproducer tests for unexpected throwing of the Missing Range message. Tests are ignored now and waiting for the issue fix.


jomarko:
@karreiro you are right. I put there note that such action is needed. Thanks.

PR #439, opened 2017-04-06
[GUVNOR-2990] broken reproducer for gap analysis with 3 ranges

mariofusco:


mariofusco:
Replaced by https://github.com/kiegroup/drools-wb/pull/441

PR #440, opened 2017-04-06
minor improvements

mariofusco:


mariofusco:
@Rikkola these are just minor readability improvements that shouldn't change the internal behaviour in any way. Feel free to discard them is you disagree with the code style though.

jomarko:
@mariofusco thanks for changes. I think they are definitely worth to merge. Please, just see few my comments at first.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/PriorityListCheck.java
```diff
@@ -42,18 +42,8 @@ public boolean isActive( final CheckConfiguration checkConfiguration ) {
...
-            if ( check.hasIssues() ) {
-                checkWithIssues = check;
-            }
-
-            if ( checkWithIssues != null ) {
-                break;
-            }
-        }
+    public boolean check() {
+        checkWithIssues = filteredSet.stream().filter( Check::check ).findFirst().orElse( null );
```
mariofusco: @jomarko anyMatch() returns a boolean, while findFirst() the first object found in the stream. They have different purposes.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/CheckStorage.java
```diff
@@ -51,25 +50,16 @@ public CheckStorage( final CheckFactory checkFactory ) {
...
 
         for ( final RuleInspector other : all ) {
             if ( !ruleInspector.equals( other ) ) {
-                final Optional<PairCheckBundle> pairCheckList = checkFactory.makePairRowCheck( ruleInspector,
-                                                                                               other );
-                if ( pairCheckList.isPresent() ) {
-                    checks.add( pairCheckList.get() );
-                    pairCheckStorage.add( pairCheckList.get() );
-                }
+                checkFactory.makePairRowCheck( ruleInspector, other ).ifPresent( pairCheckStorage::add );
```
mariofusco: @jomarko checks Set was only used as return value, but then the returned value was not used by the invoker of that method, so I just make the method to return void.

jomarko: @mariofusco thanks

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/PriorityListCheck.java
jomarko: @mariofusco thanks

PR #442, opened 2017-04-07
GUVNOR-2782: Solve marshalling issues of ImpossibleMatchIssue

jomarko:
After reading Errai Marshalling wiki it seems to me that I solved the problem, however I was not able to test appropriately the changes because of GWT stuff. Trying to test better. Please do not merge yet.

### DIFF
--- a/drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/reporting/ImpossibleMatchIssue.java
+++ b/drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/reporting/ImpossibleMatchIssue.java
@@ -15,9 +15,7 @@
  */
 package org.drools.workbench.services.verifier.api.client.reporting;
 
-import java.util.Arrays;
-import java.util.HashSet;
-
+import java.util.Set;
 import org.jboss.errai.common.client.api.annotations.MapsTo;
 import org.jboss.errai.common.client.api.annotations.Portable;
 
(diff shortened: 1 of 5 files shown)


Rikkola:
@jomarko Hi, is this bug still an issue? I could take a quick look, looks like we could get it fixed with some effort.

jomarko:
@Rikkola I added `ReceiverErraiTest` however this fails because `MarshallingWrapper` returns empty string, could you please help me with this? I hope it is just some setup issue of the test, did I miss something in the added gwt.xml?

manstis:
@jomarko @Rikkola Please advise if/when this can be merged.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/reporting/ImpossibleMatchIssue.java
```diff
@@ -38,10 +36,10 @@ public ImpossibleMatchIssue( @MapsTo("severity") final Severity severity,
                                  @MapsTo("fieldName") final String fieldName,
                                  @MapsTo("conflictedItem") final String conflictedItem,
                                  @MapsTo("conflictingItem") final String conflictingItem,
-                                 @MapsTo("rowNumbers") final Integer... rowNumbers ) {
+                                 @MapsTo("rowNumbers") final Set<Integer> rowNumbers ) {
```
Rikkola: Ok, so is this the actual fix?

Rikkola:
@jomarko Tried my best to test this, but the only way is to use GWT Test Case and we should avoid that. I made a ticket for the proper test. This PR does fix the issue so we just need to clean it up and get it in. For this to happen we need to ignore the fact that the test coverage will be incomplete in drools-wb. 

https://issues.jboss.org/browse/GUVNOR-3071

jomarko:
@Rikkola this PR updated according our irc discussion.

jomarko:
@manstis it seems to me as needed also in 7.0.x, what is your opinion?

manstis:
@jomarko Yes, I agree... let's close as many holes as possible.

PR #444, opened 2017-04-07
[*DO NOT MERGE*] RHBPMS-4645: After re-opening a saved Guided Rule in BxMS 6.4 the content is changed

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4645

- Before:
before

- After:
after

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/factPattern/PopupCreator.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/factPattern/PopupCreator.java
@@ -27,6 +27,7 @@ import org.drools.workbench.models.datamodel.oracle.DataType;
 import org.drools.workbench.models.datamodel.oracle.FieldAccessorsAndMutators;
 import org.drools.workbench.models.datamodel.oracle.ModelField;
 import org.drools.workbench.models.datamodel.rule.CompositeFieldConstraint;
+import org.drools.workbench.models.datamodel.rule.ExpressionField;
 import org.drools.workbench.models.datamodel.rule.ExpressionFormLine;
 import org.drools.workbench.models.datamodel.rule.ExpressionUnboundFact;
 import org.drools.workbench.models.datamodel.rule.FactPattern;
@@ -64,7 +65,7 @@ public class PopupCreator {
     /**
      * @param pattern the pattern to set
      */
-    public void setPattern( FactPattern pattern ) {
+    public void setPattern(FactPattern pattern) {
         this.pattern = pattern;
     }
 
@@ -78,7 +79,7 @@ public class PopupCreator {
     /**
      * @param oracle the oracle to set
      */
-    public void setDataModelOracle( AsyncPackageDataModelOracle oracle ) {
+    public void setDataModelOracle(AsyncPackageDataModelOracle oracle) {
         this.oracle = oracle;
     }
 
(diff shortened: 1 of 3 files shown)


drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/factPattern/PopupCreator.java
```diff
@@ -179,234 +180,274 @@ private boolean hasApplicableFields( final ModelField[] fields ) {
...
+            final String fieldName = box.getItemText(box.getSelectedIndex());
+
+            if ("...".equals(fieldName)) {
+                return;
             }
-        } );
-        popup.addAttribute( GuidedRuleEditorResources.CONSTANTS.AddARestrictionOnAField(),
-                            box );
+
+            factPattern.addConstraint(makeFieldConstraint(fieldName,
```
karreiro: The real change is here.

karreiro:
@porcelli Could you review and merge this? Thanks.

manstis:
@karreiro Hi.. are you confident this fixes the issue!?!? ;-)

There two ways to access "nested" propertied in the Guided Rule Editor:-

1. Use of "sub-fields"
2. Use of "Expressions"

The JIRA shows the User selecting "sub-fields" that, when saved and re-opened, are converted to "Expressions" by the ```DRL``` to ```RuleModel``` unmarshalling code. This was not acceptable to the User. Your change shows when using **sub-fields** that **expressions** are now instead created. OK, the "editor" now "looks" like the editor when re-opening the rule; however you've not IMO fixed what the the User was complaining about.

Personally, I think the change needs to be made [here](https://github.com/kiegroup/drools/blob/master/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java#L1851) (and related "unmarshal" methods) to re-construct "sub-fields" where possible; otherwise "expressions".

karreiro:
Hi @manstis,

Thanks for the comment.

I'm not sure if is possible to handle sub-fields and expressions differently, see:

(they have the same source)

Currently, we are using this logic to define what is an expression:
https://github.com/kiegroup/drools/blob/master/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelDRLPersistenceImpl.java#L3526

Perhaps, we can count the `.`s and consider an expression everything that has more than one `.`, but IDK, it does not look like the best solution.

What do you think?

manstis:
Hi @karreiro 

I agree both end up being the same DRL.. so this is what we _could_ do (I'll leave the choice to you!!!)

1. Comment the JIRA explaining why it re-opens as an expression and not sub-fields; stating our preference to remove use of sub-fields all together. Remove option to create sub-fields from the UI when authoring a Guided Rule (these are saved as DRL and re-opened as _expressions_) however Guided Templates are saved as XML and use the same editor so we'd need to preserve sub-fields for Guided Templates. 

If they agree it's best to be consistent and remove sub-fields; great. Your PRs with a little improvement (removing sub-fields for Guided Rules) would be fine. If they disagree saying the issue is they want to re-open the rule, with sub-fields, exactly as it was then we could try commenting JIRA saying "will not fix" (as ideally we'd like to cover when the Guided Rule Editor is re-written). Then let @etirelli fight our corner!! 

If they disagree we consider option 2 following...

2. Sub-fields are fields in ```ModelFields``` in the ```DataModelOracle``` whereas expressions use ```MethodInformation```. So, when parsing the LHS **if** a component is in ```ModelFields``` we know it can be handled as a sub-field; otherwise it's an expression.

PR #451, opened 2017-04-12
[GUVNOR-2962][GUVNOR-2990] fix missing range check

mariofusco:


drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/SingleRangeCheck.java
```diff
@@ -0,0 +1,353 @@
+/*
+ * Copyright 2005 JBoss Inc
```
Rikkola: Not sure why, but the year is 2005.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/cache/RuleInspectorCache.java
```diff
@@ -192,10 +223,4 @@ public RuleInspector getRuleInspector( final int row ) {
     public AnalyzerConfiguration getConfiguration() {
         return configuration;
     }
-
-    public interface Filter {
-
-        boolean accept( final RuleInspector ruleInspector );
-
-    }
```
Rikkola: Removing this is a good move.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/CheckFactory.java
```diff
@@ -67,8 +66,6 @@ public CheckFactory( final AnalyzerConfiguration configuration ) {
                                                                        configuration  ),
                                       new DetectDeficientRowsCheck( ruleInspector,
                                                      configuration ),
-                                      new RangeCheck( ruleInspector,
```
jomarko: Hi @mariofusco just question, why the new `SingleRangeCheck` is not registered on this place but in the `RuleInspectorCache`?

mariofusco:
@jomarko Because all the checks here work with a different logic: they run once per rule/row and check that single row. Conversely my new check runs just once for the whole dtable taking in input all the rows and checking if there are missing ranges. In this is way it is not only way faster but even more consistent and easier to implement. In general indeed it doesn't make sense to ask if a single row is causing a missing range: the missing range is the result of the fact that all the rows in the dtable (taken as a whole) aren't covering the entire domain and then it needs to be checked at dtable level.

jomarko:
@mariofusco Thanks for the answer. It makes sense definitely.

PR #453, opened 2017-04-13
GUVNOR-2998: GDT Editor - New column (wizard) - Error on 'Set the value of a field' plugin

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2998

- Before:
before

- After:
after

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ActionSetFactPluginTest.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ActionSetFactPluginTest.java
@@ -154,23 +154,6 @@ public class ActionSetFactPluginTest {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


karreiro:
@manstis Could you please take a look?

manstis:
Thanks for the quick fix @karreiro :-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ActionSetFactPluginTest.java
```diff
@@ -225,7 +208,7 @@ public void testIsValidWhenItIsValid() {
         when(presenter.getModel()).thenReturn(model);
         when(actionCol52.getHeader()).thenReturn("header");
         when(actionWrapper.getFactField()).thenReturn("factField");
-        when(actionWrapper.getFactType()).thenReturn("factType");
+        when(actionWrapper.getFactType()).thenReturn("");
```
jomarko: Hi @karreiro I would like to ask about this change. It the insert fact action really valid if the inserted fact type equals to empty string? I would expect that the insert fact action must handle with some specific fact type like `Person`. Or am I missing something?

PR #454, opened 2017-04-17
GUVNOR-3005: Remove search textbox from the menu header

paulovmr:


paulovmr:
@Rikkola Hi Toni, if you could take a look please :)

drools-wb-webapp/src/main/java/org/drools/workbench/client/DroolsWorkbenchEntryPoint.java
```diff
@@ -55,16 +54,18 @@
...
-                                      final DefaultWorkbenchFeaturesMenusHelper menusHelper,
-                                      final WorkbenchMenuBarPresenter menuBar,
-                                      AdminPage adminPage,
-                                      DefaultAdminPageHelper adminPageHelper) {
-        super( appConfigService, pmas, activityBeansCache );
+    public DroolsWorkbenchEntryPoint(final Caller<AppConfigService> appConfigService,
+                                     final Caller<PlaceManagerActivityService> pmas,
+                                     final ActivityBeansCache activityBeansCache,
+                                     final PlaceManager placeManager,
+                                     final SyncBeanManager iocManager,
```
jomarko: Hi @paulovmr, just question, is the `iocManager` still needed after your changes? Is `iocManager` used in some child class of `DroolsWorkbenchEntryPoint`?

paulovmr: Hi @jomarko . No, I will remove it in my next PR, thanks.

PR #470, opened 2017-04-26
Remove dashbuilder references

manstis:
@Rikkola This makes ```drools-wb-webapp``` run OK for me...


manstis:
@Rikkola hmmmm.. the compilation error suggests something is needed from DashBuilder.. perhaps I'll wait until tomorrow and check ```drools-wb-webapp``` again.. in case there's some SNAPSHOTs somewhere that need updating. Jenkins usually knows best; and uses the latest sources..

PR #472, opened 2017-04-26
GUVNOR-3037: Lucene Index: DSL files that are created into default package should be indexed

manstis:
See https://issues.jboss.org/browse/GUVNOR-3037

The issue was that when the file was in the default package; the package name is an empty ```String``` and the ```DRL``` generated from the ```DSL``` (to extract class and field references etc) was of the form ```package rule "test" when... ``` and failed ```DRL``` compilation. The resulting error was logged as a warning (but not listed with the default log level).


drools-wb-screens/drools-wb-dsl-text-editor/drools-wb-dsl-text-editor-backend/src/main/java/org/drools/workbench/screens/dsltext/backend/server/indexing/DslFileIndexer.java
```diff
@@ -57,102 +56,105 @@
...
         final StringBuilder sb = new StringBuilder();
-        final String packageName = getPackageName( path );
-
-        sb.append( "package " ).append( packageName ).append( "\n" );
-        sb.append( "rule \"" + MOCK_RULE_NAME + "\"\n" );
-        sb.append( "when\n" );
-        for ( String e : lhs ) {
-            sb.append( e ).append( "\n" );
+        final String packageName = getPackageName(path);
+        if (!(packageName == null || packageName.isEmpty())) {
```
manstis: Here's the fix!

PR #481, opened 2017-05-04
GUVNOR-3047: [GDT Column Wizard] ValueList and DefaultValue are disabled

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3047

Before:
before

After:
after


karreiro:
@manstis Could please take a look?

PR #482, opened 2017-05-04
[7.0.x] GUVNOR-3047: [GDT Column Wizard] ValueList and DefaultValue are disabled

karreiro:
Cherry picked from https://github.com/kiegroup/drools-wb/pull/481


karreiro:
@manstis Could please take a look?

PR #483, opened 2017-05-08
GUVNOR-2900: [Guided Decision Table] User is not prompted to save or discard changes

manstis:
See https://issues.jboss.org/browse/GUVNOR-2900

Fixes test that relied on hash code.


manstis:
@jomarko @karreiro Could you please review.

PR #485, opened 2017-05-09
GUVNOR-3062: [Guided Decision Table] Just one pattern can be setup in the wizard

manstis:
See https://issues.jboss.org/browse/GUVNOR-3062

@jomarko I'm open to ideas about how this could be tested before merge!?!

GWT's ```MultiSelectionModel``` uses a ```HashMap``` for it's "keys" (see [here](https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/view/client/MultiSelectionModel.java#L78)) which, following GUVNOR-2900, produces different values for a (for example) ```Pattern52``` when it's binding is set vs unset. This leads to the "selected" and "unselected" items in the tables to be unidentifiable.. Using ```System::identityHashCode``` as the "key provider" means the original behaviour is maintained by using the Object's identity.


jomarko:
@manstis I had success with testing locally, but hard to describe, expect PR today.

jomarko:
@manstis this PR works fine. There is just question how to get in my tests. Merge this PR, then rebase and merge my PR? What is your preference?

manstis:
Replaced by https://github.com/kiegroup/drools-wb/issues/486

Thanks @jomarko

PR #487, opened 2017-05-10
GUVNOR-3068: [Guided Decision Table] Limited Entry Table ignore field data types

manstis:
See https://issues.jboss.org/browse/GUVNOR-3068

The problem was not just for "Limited Entry" tables; but any new "Update Fact's Field" action.


manstis:
Jenkins please retest this.

PR #489, opened 2017-05-11
GUVNOR-2782 : [Guided Decision Table] Row with Impossible Match throws Exception

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2782

@jomarko Hey, this was my take on the fix for #442, but while this picks up several of the bugs that I ran into previously it does not pick up of fix the bug you found. Still a good addition to our test coverage and this can be extended from this simple hashcode comparison.

To really test the actual cause we would have to compile to GWT. I tried with GWT Test Case. It is slow and hard to setup. I think a better bet is to make a selenium test since:
a. We already test with Selenium
b. No need to compile any code to JavaScript

Made a ticket for it. https://issues.jboss.org/browse/GUVNOR-3071

### DIFF
--- a/drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/pom.xml
+++ b/drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/pom.xml
@@ -27,6 +27,12 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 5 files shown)


manstis:
Github didn't show any conflicts, so merged....

PR #498, opened 2017-05-18
GUVNOR-3078: [Guided Decision Table Graph] Graph is closed automatically after removing last table

manstis:
See https://issues.jboss.org/browse/GUVNOR-3078

Removing the last document no longer closes the editor.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/BaseGuidedDecisionTableEditorPresenter.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/BaseGuidedDecisionTableEditorPresenter.java
@@ -17,6 +17,7 @@
 package org.drools.workbench.screens.guided.dtable.client.editor;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import javax.enterprise.event.Event;
 
@@ -54,6 +55,7 @@ import org.uberfire.client.workbench.events.ChangeTitleWidgetEvent;
 import org.uberfire.client.workbench.type.ClientResourceType;
 import org.uberfire.ext.editor.commons.client.file.popups.SavePopUpPresenter;
 import org.uberfire.ext.editor.commons.client.history.VersionRecordManager;
+import org.uberfire.ext.editor.commons.client.menu.MenuItems;
 import org.uberfire.ext.editor.commons.client.validation.DefaultFileNameValidator;
 import org.uberfire.ext.widgets.common.client.callbacks.HasBusyIndicatorDefaultErrorCallback;
 import org.uberfire.mvp.PlaceRequest;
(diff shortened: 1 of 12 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/BaseGuidedDecisionTableEditorPresenter.java
```diff
@@ -249,166 +250,171 @@ public void onClose() {
         modeller.onClose();
     }
 
-    protected void onDecisionTableSelected( final DecisionTableSelectedEvent event ) {
-        final GuidedDecisionTableView.Presenter dtPresenter = event.getPresenter();
-        if ( dtPresenter == null ) {
+    protected void onDecisionTableSelected(final DecisionTableSelectedEvent event) {
+        final Optional<GuidedDecisionTableView.Presenter> dtPresenter = event.getPresenter();
```
manstis: Oh look.. here's a change...

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/BaseGuidedDecisionTableEditorPresenter.java
```diff
@@ -249,166 +250,171 @@ public void onClose() {
...
 
         openOtherDecisionTable();
     }
 
     void openOtherDecisionTable() {
         final Set<GuidedDecisionTableView.Presenter> availableDecisionTables = modeller.getAvailableDecisionTables();
-        if ( availableDecisionTables == null || availableDecisionTables.isEmpty() ) {
-            placeManager.forceClosePlace( editorPlaceRequest );
+        if (availableDecisionTables == null || availableDecisionTables.isEmpty()) {
+            decisionTableSelectedEvent.fire(DecisionTableSelectedEvent.NONE);
```
manstis: And here's the crux of the change..

manstis:
Hi @jomarko thanks for taking a look. 

I've moved the "radar" enabling/disabling to the ```RadarMenuBuilder```. 

I cannot however move the "validate" menu item related code to ```activateDocument(..)``` as that method is [not called](https://github.com/kiegroup/drools-wb/pull/498/files#diff-cf029670362eba75c492cbe69b21efe7R258) if there is no concrete document selected. 

Furthermore I cannot (easily) observe ```DecisionTableSelectedEvent``` events in the "validate" menu item itself (as I do for "radar" and "view" menus) as it's instantiation is burried in Uberfire's ```BasicFileMenuBuilder```. 

That said however all the enabling/disabling is in response to a ```DecisionTableSelectedEvent``` event - and the code for the "validate" menu item is in such a handler.. so I think it's OK, but open to other ideas.

jomarko:
@manstis thanks, however, now I don't see the enabling/disabling of Validate menu nowhere. Did you removed it by accident?

manstis:
@jomarko Nope.. it's still [there](https://github.com/kiegroup/drools-wb/pull/498/files#diff-cf029670362eba75c492cbe69b21efe7R255)!

PR #499, opened 2017-05-19
GUVNOR-3078: Save of empty graph never ends

jomarko:
The save operation on graph that doesn't inlcude decision tables was causing never
ending saving. In practise, the saving progress popup was never hidden.

@manstis please take a look when you have a while. I forgot to check this yesterday. Sorry for inconvenience.


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableGraphEditorPresenterTest.java
```diff
@@ -973,6 +974,51 @@ public void checkSaveDocumentGraphEntries() {
...
+               times(1)).showSaving();
+        verify(saveInProgressEvent,
+               never()).fire(any(SaveInProgressEvent.class));
+        verify(dtGraphService,
+               times(1)).save(eq(dtGraphPath),
+                              any(GuidedDecisionTableEditorGraphModel.class),
+                              any(Metadata.class),
+                              eq("message"));
+
+        ArgumentCaptor<NotificationEvent> notification = ArgumentCaptor.forClass(NotificationEvent.class);
```
manstis: Cosmetic minor comment.. if you're using ```final``` (in the method) keep consistent.

jomarko: Makes sense definitely. Thanks. Amended now.

PR #501, opened 2017-05-19
GUVNOR-3092: [Guided Decision Table Graph] Editor does not close when deleting file

manstis:
See https://issues.jboss.org/browse/GUVNOR-3092

@jomarko This was caused by https://issues.jboss.org/browse/GUVNOR-3078

The old code used to close the editor when the last document was deregistered (which for a "graph" was undesirable; but for the single decision table editor deletion of the file deregistered the document which no longer lead to closure).

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableEditorPresenter.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableEditorPresenter.java
@@ -24,6 +24,7 @@ import javax.enterprise.event.Event;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 
+import com.google.gwt.core.client.Scheduler;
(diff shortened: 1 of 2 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/GuidedDecisionTableEditorPresenter.java
```diff
@@ -179,35 +181,45 @@ public void onClose() {
...
 
     @Override
-    public void getAvailableDocumentPaths( final Callback<List<Path>> callback ) {
-        callback.callback( Collections.<Path>emptyList() );
+    public void getAvailableDocumentPaths(final Callback<List<Path>> callback) {
+        callback.callback(Collections.<Path>emptyList());
+    }
+
+    @Override
+    public void removeDocument(GuidedDecisionTableView.Presenter dtPresenter) {
```
manstis: Here's the change. This method is called from ```KieMultipleDocumentEditor``` when the "delete" menu item is clicked.

jomarko:
@manstis sorry I don't manage review today. I will have a look on Monday. Have a nice weekend.

PR #502, opened 2017-05-19
GUVNOR-3094: Validation moved from Wizard Plugins to particular Wizar…

jomarko:
…d Pages

@manstis @karreiro please take a look when you have a while.

screenshot from 2017-05-19 15-31-36
screenshot from 2017-05-19 15-31-22
screenshot from 2017-05-19 15-31-10
screenshot from 2017-05-19 15-31-01
screenshot from 2017-05-19 15-30-50


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ActionWorkItemPluginTest.java
```diff
@@ -124,103 +124,103 @@ public void testSetHeader() throws Exception {
...
+//        }};
+//
+//        when(actionCol52.getHeader()).thenReturn(header);
+//        when(model.getActionCols()).thenReturn(actionCol52s);
+//        when(presenter.getModel()).thenReturn(model);
+//
+//        final boolean unique = plugin.unique(header);
+//
+//        assertFalse(unique);
+//    }
```
karreiro: I think that you can remove these lines :-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/AdditionalInfoPage.java
```diff
@@ -135,10 +151,29 @@ void setupUpdateEngineWithChanges() {
...
+            }
+        }
+
+        for (CompositeColumn<?> cc : presenter.getModel().getConditions()) {
+            for (int iChild = 0; iChild < cc.getChildColumns().size(); iChild++) {
+                if (cc.getChildColumns().get(iChild).getHeader().equals(header)) {
+                    return false;
+                }
+            }
+        }
```
karreiro: Instead of iterating twice every time, what do you think about getting an array of headers from the plugin?

jomarko: Hi @karreiro could you please give me more specific example what do you mean? I didn't find any method for obtaining all column headers form the plugin?

karreiro: Hi, @jomarko. The method does not exist, and I'm proposing its creation. You could add the method signature in `HasAdditionalInfoPage` interface (and consequently, you would have different implementations in `ActionRetractFactPlugin`, `ActionSetFactPlugin`, `ActionWorkItemPlugin`, `ActionWorkItemSetFieldPlugin`, `BRLActionColumnPlugin`, `BRLConditionColumnPlugin`, `ConditionColumnPlugin`).

jomarko: @karreiro I see it now thanks, probably one more question? Do we want keep the behavior that column name has to be unique just separately among condition and action columns? I mean should we keep the option that action and condition column with the same name can exist?

karreiro: @jomarko I'm not sure about the consequences of changing this behavior. Why shouldn't we keep that? :-)

jomarko: @karreiro I am not saying we should or shouldn't. In my eyes is more logical to keep unique column header regardless of column type (condition/action). However I am completely fine with keeping column header unique separately in conditions and actions. I will update the PR and let you know when I am done.

jomarko:
@karreiro please have a look when you have a moment. I implemented the method for obtaining already used column names in each plugin that implements `HasAdditionalInfo` interface. However I am not satisfied with the repetition of the code in particular implementations of `getAlreadyUsedColumnNames()` do you have some suggestion how to improve this?

manstis:
What's the status of this PR? Are there outstanding changes to be completed? Does it need re-review?

jomarko:
@manstis it is done from my site. Just waiting for @karreiro comments, suggestions etc.

karreiro:
@jomarko you could avoid the duplication by extracting the duplicated to code to some utility class in the package `org.drools.workbench.screens.guided.dtable.client.wizard.column.plugins.commons`. However, I think that the code is good enough to be merged ;-)

PR #503, opened 2017-05-19
GUVNOR-3093: [Guided Decision Table Graph] Pinned mode does not work

manstis:
See https://issues.jboss.org/browse/GUVNOR-3093

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableViewImpl.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableViewImpl.java
@@ -19,6 +19,8 @@ import java.util.Set;
 import javax.enterprise.event.Event;
 
 import com.ait.lienzo.client.core.event.INodeXYEvent;
+import com.ait.lienzo.client.core.event.NodeDragMoveHandler;
+import com.ait.lienzo.client.core.event.NodeMouseDoubleClickHandler;
 import com.ait.lienzo.client.core.shape.BoundingBoxPathClipper;
 import com.ait.lienzo.client.core.shape.Group;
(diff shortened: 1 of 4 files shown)


jomarko:
@manstis sorry I don't manage review today. I will have a look on Monday. Have a nice weekend.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableViewImpl.java
```diff
@@ -110,9 +110,27 @@ public GuidedDecisionTableViewImpl(final GridData uiModel,
...
 
+        registerNodeDragMoveHandler();
+        registerNodeMouseDoubleClickHandler();
+    }
+
+    void registerNodeDragMoveHandler() {
         addNodeDragMoveHandler((event) -> presenter.getModellerPresenter().updateRadar());
     }
 
+    void registerNodeMouseDoubleClickHandler() {
```
jomarko: I miss some test for the logic of the double click handler. Ideally I would move the handler logic into the presenter class. It will probably require to make the `isNodeMouseEventOverCaption` as public method of the view.

I admit that the logic is not complex but I think worth to test. If you are not going to add the test please file a jira for me @manstis

manstis:
@jomarko Refactored and tests added. Thanks for your review.

jomarko:
@manstis thanks. Good to merge I think. However it seems something is broken somewhere. All PRs in kiegroup seem to fail.

manstis:
@jomarko Yes, there's an issue in ```droolsjpm-integration``` that @mariofusco (I think) is investigating

manstis:
Jenkins please rebuild this.

ederign:
Jenkins please rebuild this.

manstis:
Jenkins please rebuild this.

manstis:
Jenkins please rebuild this.

manstis:
Jenkins please rebuild this.

manstis:
Jenkins please retest this.

PR #509, opened 2017-05-24
GUVNOR-2980: Test scenarios do not show correct results

manstis:
See https://issues.jboss.org/browse/GUVNOR-2980

This PR removes the ```TextBox``` for rule names and replaces it with a nicer combo/selector widget.

### DIFF
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioWidgetComponentCreator.java
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioWidgetComponentCreator.java
@@ -19,14 +19,9 @@ import java.util.Collection;
 import java.util.List;
 import javax.inject.Inject;
 
-import com.google.gwt.event.dom.client.ChangeEvent;
-import com.google.gwt.event.dom.client.ChangeHandler;
-import com.google.gwt.event.dom.client.ClickEvent;
-import com.google.gwt.event.dom.client.ClickHandler;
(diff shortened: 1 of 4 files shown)


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioWidgetComponentCreator.java
```diff
@@ -35,213 +30,180 @@
...
         this.scenarioWidget = scenarioWidget;
         this.oracle = oracle;
         this.scenario = scenario;
-        this.path = path;
 
-        this.ruleNamesService.call( new RemoteCallback<Collection<String>>() {
-            @Override
-            public void callback( final Collection<String> ruleNames ) {
-                hasRules = !( ruleNames == null || ruleNames.isEmpty() );
+        this.ruleNameSelector.clear();
```
manstis: This is the crux of the change.. using a ```Select``` instead of ```TextBox``` and ```ListBox```

manstis:
This is how it now looks:

rulename-selector

Rikkola:
Looks really nice. Kind of sticks out from the rest of the editor ;-)

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioWidgetComponentCreator.java
Rikkola: Nice. Looks like this gives the search as you type for free.

PR #510, opened 2017-05-24
GUVNOR-3110: [Guided Decision Table] Rule Selector drop-down is incorrectly positioned

manstis:
See https://issues.jboss.org/browse/GUVNOR-3110


jsoltes:

PR #511, opened 2017-05-24
[UF-506] WorkspaceScoped Services

adrielparedes:
Based on this PR: https://github.com/AppFormer/uberfire/pull/718

It change some ApplicationScoped beans to be WorkspaceScoped, check referenced PR

### DIFF
--- a/drools-wb-webapp/pom.xml
+++ b/drools-wb-webapp/pom.xml
@@ -191,6 +191,10 @@
     </dependency>
     <dependency>
       <groupId>org.uberfire</groupId>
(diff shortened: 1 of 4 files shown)


manstis:
@adrielparedes if ```drools-wb-webapp``` needs changes to work with the _newer_ backend (your other PRs) it is very likely Designer, jBPM Workbench, (possibly) Form Modeller, Stunner and definitely ```kie-drools-wb``` and ```kie-wb``` are going to need similar changes. Have you tried your changes with any of these other webapps?

porcelli:
@manstis the new backend won't be used for real during 7 series

PR #519, opened 2017-06-05
GUVNOR-2821: [Guided Decision Table] Wizard allows two columns with the same description

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2821

guvnor-2821

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/ValidatorTest.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/ValidatorTest.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/ValidatorTest.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Test
+    public void testIsConditionValidWhenHeaderIsValid() throws Exception {
+        final ConditionCol52 conditionCol1 = fakeConditionCol52("header1");
+        final ConditionCol52 conditionCol2 = fakeConditionCol52("header2");
+        final Validator validator = new Validator(fakeCompositeColumn(conditionCol1,
+                                                                      conditionCol2));
+
+        assertFalse(validator.isConditionValid(conditionCol2));
```
jomarko: In the context of GUVNOR-2821 I would expect `assertTrue` here, can you please update the `fakeConditionCol52` method to achieve this? Let me know if you don't see my point.

karreiro: Thank you @jomarko :-) Fixed!

PR #520, opened 2017-06-05
GUVNOR-3052: [GDT Column Wizard] Field Binding is disabled for LimitedEntry table

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3052

aaaa

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ConditionColumnPlugin.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ConditionColumnPlugin.java
@@ -193,7 +193,7 @@ public class ConditionColumnPlugin extends BaseDecisionTableColumnPlugin impleme
             editingCol().setOperator(null);
         }
 
-        if (constraintValue() != BaseSingleFieldConstraint.TYPE_LITERAL) {
+        if (!isBindable()) {
             editingCol().setBinding(null);
         }
     }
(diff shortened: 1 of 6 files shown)


karreiro:
@manstis @jomarko Could you please take a look?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ConditionColumnPlugin.java
```diff
@@ -199,7 +199,7 @@ void prepareValues() {
             editingCol().setOperator(null);
         }
 
-        if (constraintValue() != BaseSingleFieldConstraint.TYPE_LITERAL) {
+        if (!(tableFormat() == GuidedDecisionTable52.TableFormat.LIMITED_ENTRY || constraintValue() == BaseSingleFieldConstraint.TYPE_LITERAL)) {
```
jomarko: Hi @karreiro , just question, this condition is quite complex and it seems to me that it should be kept in sync with `valueOptionsPage.canSetupBinding()`. I see that `canSetupBinding()` and this `if` are sligtly different, however, what do you think , could this `if` looks like `if (`valueOptionsPage.canSetupBinding())`?

karreiro: Thank you for the idea, @jomarko. I did something a little bit different, but I kept the idea of DRYing the code ;-)

manstis:
@karreiro Just the conflict to resolve (we guessed we'd have one somewhere)

karreiro:
Jenkins please retest this.

karreiro:
Jenkins please retest this.

karreiro:
****

karreiro:
Jenkins please retest this.

karreiro:
Jenkins please retest this.

PR #521, opened 2017-06-05
GUVNOR-2864: [Guided Decision Table] Values are reset during column edit

karreiro:
See https://issues.jboss.org/browse/GUVNOR-2864

sasasasa


karreiro:
@manstis @jomarko Could you please take a look? :-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/RuleModelCloneVisitor.java
```diff
@@ -52,346 +52,341 @@
...
     }
 
-    private DSLSentence visitDSLSentence( final DSLSentence sentence ) {
-        DSLSentence clone = new DSLSentence();
-        clone.setDefinition( sentence.getDefinition() );
-        clone.setDrl( sentence.getDrl() );
-        // Do not clone values, these are computed from definition and will be cloned internally
-        return clone;
+    private DSLSentence visitDSLSentence(final DSLSentence sentence) {
+        return sentence.copy();
```
karreiro: The real change is here.

karreiro:
@manstis Could you please approve and merge this one? :-)

PR #522, opened 2017-06-06
GUVNOR-3134: Test scenarios fail after upgrade to 7.1 because rule expectation in 6.x doesn't use FQN

manstis:
See https://issues.jboss.org/browse/GUVNOR-3134

This PR effectively reverts https://github.com/kiegroup/droolswb/commit/a12cd9b352322df1e9ea75b93ce455f95192a8a5 made for https://issues.jboss.org/browse/GUVNOR-2980 that attempted to switch to use of 'Fully Qualified Rule Names' for Test Scenarios. However this causes problems with backwards compatibility and an alternative solution has been added for GUVNOR-3134 keeping use of 'Simple Rule Names'.

### DIFF
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioWidgetComponentCreator.java
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioWidgetComponentCreator.java
@@ -82,10 +82,10 @@ public class ScenarioWidgetComponentCreator {
                         scenario.getPackageName());
     }
 
-    Option makeRuleNameOption(final String text) {
+    Option makeRuleNameOption(final String simpleRuleName) {
         final Option o = GWT.create(Option.class);
-        o.setText(text);
-        o.setValue(text);
+        o.setText(simpleRuleName);
+        o.setValue(simpleRuleName);
         return o;
     }
 


manstis:
@jomarko I've tested this with https://issues.jboss.org/browse/GUVNOR-2980 and https://issues.jboss.org/browse/GUVNOR-2840 and both are OK.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioWidgetComponentCreator.java
```diff
@@ -74,21 +74,26 @@ protected void reset(final ScenarioParentWidget scenarioWidget,
...
-        o.setValue(text);
+        o.setText(simpleRuleName);
+        o.setValue(simpleRuleName);
         return o;
     }
 
+    String getSimpleRuleName(final String fullyQualifiedRuleName) {
+        final int dotIndex = fullyQualifiedRuleName.lastIndexOf(".");
+        return dotIndex > -1 ? fullyQualifiedRuleName.substring(dotIndex + 1) : fullyQualifiedRuleName;
+    }
```
yurloc: @manstis I think this will produce an incorrect result if the simple rule name contains a "`.`". Could you instead replace the first occurrence of the current package name?

manstis: @yurloc What a can of worms this is :-(

"First occurrence" works OK where the package names are completely different (or sufficiently different) but not so for child packages; e.g. ```org.pkg.rule1``` and ```org.pkg.child.rule1``` when we'd end up with a rule called ```child.rule2``` which is clearly wrong. The problem is that the index stores fully qualified rule names (as String) but we don't know what portion is the _package_ vs _rule name_. So a fix (with simple Rule Names) for rules containing "```.```" breaks for child packages and a fix for child packages breaks for rules... Perhaps FQRN and a breaking change for 7.x is the best option after all?

yurloc: @manstis Great point. So we've uncovered the root issue with the index, where FQRNs are ambiguous! For example, `com.somepgk.my.r1` can mean either

    package com.somepkg; rule "my.r1"
or

    package com.somepkg.my; rule "r1"

right?

Would it be feasible to remove the root cause by fixing the index so that it stores FQRNs in an unambiguous manner? For example, `com.mypkg."some rule with.dots;and;stuff?"` (rule simple name enclosed with double quotes).

jomarko:
I see the problem of `.` corner case, however I am not sure if the proposed enclosing by `"` wouldn't cause problems if the rule name contains `"` itself. @manstis please let us know when we should check the PR again.

manstis:
@jomarko Yes.. fix for corners is in progress.. stay tuned!

manstis:
Part of an ensemble:

- https://github.com/kiegroup/drools/pull/1299
- https://github.com/kiegroup/kie-wb-common/pull/923
- https://github.com/kiegroup/drools-wb/pull/522

manstis:
Jenkins please retest this.

manstis:
@yurloc @jomarko OK.. all fixed. Rules with "```.```" in their name can be used; and I only show/use "Simple Rule Names" so 6.x legacy Test Scenarios should continue to function unaffected. Please re-review the ensemble.

manstis:
@karreiro Looks like the failing test is caused by the changes to ```DSLSentence``` cloning. Whilst the "value" is copied; the "id" is not.. run ```org.drools.workbench.screens.guided.dtable.client.widget.RuleModelCloneVisitorTest``` and put a breakpoint on L#170 and you'll see. Could you please fix?

manstis:
Jenkins please retest this.

yurloc:
@manstis Is the only effect of this PR changing the parameter name or am I missing something?

manstis:
@yurloc You know, I procrastinates over this PR a while. I toyed with closing it unmerged.. but then I thought "hell, the parameter name does make it _easier_ to know what _type_ of rule name is being passed..." so thought to let it be merged (although for no functional reason).

PR #523, opened 2017-06-07
GUVNOR-3138: Correct BRL Action fragment description

jomarko:
screenshot from 2017-06-07 10-19-42
screenshot from 2017-06-07 10-19-34


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/RuleModellerPageView.java
```diff
@@ -35,21 +38,34 @@
...
+                                final @Named("h5") Heading descriptionHeading) {
         this.ruleModellerContainer = ruleModellerContainer;
+        this.descriptionHeading = descriptionHeading;
     }
 
     @Override
     public void init(final RuleModellerPage page) {
         this.page = page;
+        if (page.plugin() instanceof BRLActionColumnPlugin) {
+
```
manstis: If it is then do what exactly? If we don't want to do anything do we need this ```if(..)``` block at all?

jomarko:
fail not related to this PR

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/RuleModellerPageTest.java
```diff
@@ -117,7 +117,12 @@ public void testPrepareView() throws Exception {
         page.prepareView();
 
         verify(view).init(page);
+        verify(view).setupRuleModellerWidget(any(RuleModeller.class));
         verify(page).markAsViewed();
+
+        verify(translationService).format(GuidedDecisionTableErraiConstants.RuleModellerPage_InsertAnActionBRLFragment);
```
manstis: Is there an equivalent test somewhere (or a new one?) that needs updating to show "Condition BRL Fragment" titles?

manstis:
@jomarko Failure may be unrelated but someone (normally I'd suggest the person submitting the PR) should chase down someone to fix the issue.. 

@lanceleverich @mariofusco Looks like https://github.com/kiegroup/drools/commit/cad144e5a9e5c527e076878ecf4d96718909d829 broke something. Do you have another PR in the pipeline!?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/RuleModellerPageView.java
jomarko: Sorry, I intended to do decision about description in the view, but there is missing `translationService` so I decided to implement the logic in the page. I will remove this empty `if` block. Thanks.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/RuleModellerPageTest.java
jomarko: It is not ==> it will be. Thanks.

jomarko:
@manstis yes, you just figured out the rootcause faster than me. sorry for overhead.

mariofusco:
@manstis @jomarko actually I'm making a bigger refactor moving those classes from drools to jbpm. I'll fix drools-wb accordingly.

manstis:
Jenkins please retest this.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/RuleModellerPage.java
```diff
@@ -75,6 +76,11 @@ public void prepareView() {
 
     private void setupRuleModeller() {
         view.setupRuleModellerWidget(ruleModeller());
+        if (plugin() instanceof BRLActionColumnPlugin) {
```
karreiro: Personally, I would add a method in `HasRuleModellerPage` to get the description. I think that's a good idea to avoid the `instanceOf` and follow the open/closed principle. However, I know that maybe my approach is a little bit more prolix. What do you think?

manstis:
Jenkins please retest this.

jomarko:
@karreiro thanks. I applied your hint. I think it is better approach to have two simple methods than one `instanceof` method. Have a look when you have a moment.

PR #524, opened 2017-06-07
[DROOLS-1608] move processes related classes from drools to jbpm

mariofusco:


mariofusco:
depends on kiegroup/drools#1307

manstis:
@mariofusco Is the move still going ahead; or can this be closed?

mariofusco:
@manstis I just summarized the current situation in my comment here https://github.com/kiegroup/jbpm/pull/872#issuecomment-306816403 If this is not a problem please let's keep this PR open and rediscuss it when @mswiderski will be back.

manstis:
@mariofusco @mswiderski Hi.. is this move still going ahead?

mswiderski:
@manstis that's a good question. I was not involved in the origin of this so not sure. @mariofusco what do you think?

manstis:
Jenkins please retest this.

mariofusco:
ok to test

PR #527, opened 2017-06-09
AF-527: aded new appformer maven support artifact for gwt build

porcelli:
depends on AppFormer/uberfire#740


manstis:
Do we need this added to the webapp's gwt.xml file?

porcelli:
@manstis it `inherits` from kie-api

PR #528, opened 2017-06-13
GUVNOR-3170: Add reproducer tests

jomarko:


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ActionWorkItemSetFieldPlugin.java
```diff
@@ -135,7 +135,7 @@ private boolean acceptParameterType(final PortableParameterDefinition ppd) {
             return false;
         }
 
-        final String fieldClassName = oracle.getFieldClassName(patternWrapper().getFactType(),
```
jomarko: @karreiro please have a look especially on this line. I had problems to mock properly in tests so I did this change, were there special reason for `patternWrapper()` instead of `editingWrapper()`?

karreiro: Personally, since it's a reproducer, I think that is a good idea to keep this code as is.

You could try to add, in the file `ActionWorkItemSetFieldPluginTest.java` (line 467), something like:
```java
PatternWrapper patternWrapper = mock(PatternWrapper.class);
when(patternWrapper.getFactType()).thenReturn("Person");
when(plugin.patternWrapper()).thenReturn(patternWrapper);
```
Probably, you would get the expected result :-)

jomarko: Simple things are sometimes the hardest. Thanks @karreiro . Hint applied.

PR #533, opened 2017-06-19
AF-546: Reduce startup time of AppFormer apps and workbenches

mbarkley:
* Convert @AfterInititialization calls to @PostConstruct
* Remove PlaceManagerActivityService

### DIFF
--- a/drools-wb-webapp/src/main/java/org/drools/workbench/client/home/HomeProducer.java
+++ b/drools-wb-webapp/src/main/java/org/drools/workbench/client/home/HomeProducer.java
@@ -15,8 +15,8 @@
 
 package org.drools.workbench.client.home;
 
-import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.Dependent;
 import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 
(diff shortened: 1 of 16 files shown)


mbarkley:
Relates to:
* https://github.com/errai/errai/pull/267
* https://github.com/AppFormer/uberfire/pull/751
* https://github.com/dashbuilder/dashbuilder/pull/340
* https://github.com/kiegroup/guvnor/pull/464
* https://github.com/kiegroup/kie-wb-common/pull/937
* https://github.com/kiegroup/jbpm-form-modeler/pull/117
* https://github.com/kiegroup/kie-wb-distributions/pull/542
* https://github.com/kiegroup/optaplanner-wb/pull/181
* https://github.com/kiegroup/appformer/pull/88
* https://github.com/kiegroup/jbpm-wb/pull/764

manstis:
@mbarkley Does this [need](https://github.com/manstis/drools-wb/blob/master/drools-wb-webapp/src/main/java/org/drools/workbench/client/home/HomeProducer.java) changing too?

mbarkley:
@manstis TL;DR The change isn't strictly necessary here, but I will update this PR for consistency.

The `HomeProducer` in this repo initialises itself with a `@PostConstruct` so there is nothing that *needs* to be changed (the ones in kie-wb-distributions, in contrast, were initialised by an entrypoint).

Having the producing class as `@Dependent` and the `@Produces` method as `@ApplicationScoped` has minor  benefits as every `@Dependent` bean has a small memory overhead (because they have to be cleaned up if the bean they were injected in is destroyed).

mbarkley:
@manstis Updated the `HomeProducer`.

drools-wb-webapp/src/main/java/org/drools/workbench/client/home/HomeProducer.java
```diff
@@ -36,21 +36,14 @@
...
+@Dependent
 public class HomeProducer {
 
-    private HomeModel model;
-
-    @Inject
-    private PlaceManager placeManager;
-
-    @Inject
-    protected LibraryMonitor libraryMonitor;
```
ederign: @mbarkley are you sure that this is not needed? Sometimes I saw on our codebase this Injection points used to "force" the bean startup.

Maybe is not on this case because this code lives on drools-wb-webapp, but we should be careful doing this.

mbarkley: @ederign The `LibraryMonitorImpl` is `@ApplicationScoped` so it won't actually by created just by injecting it (it would have to actually be used). And the `PlaceManager` is definitely used (indirectly) by several entry points, so I don't think there is any side-effect from deleting these unused references.

PR #538, opened 2017-06-28
GUVNOR-3231: Select placeholder if no operator defined

jomarko:
@manstis @jsoltes please have a look when you have a moment.


jomarko:
@manstis in my opinion the problems is connected with the `if` you highlighted, however in slightly different way. The given `if` is not executed during first dropdown creation, thus the dropdown seems as first added item (equal to) is selected but the `ValueChangeEvent` was not fired. This PR ensures that "please choose" is selected in the dropdown after first initialization.

manstis:
Thanks for the explanation @jomarko 

Looking at the code in my IDE (as opposed to just the PR) I see ```addPlaceHolder(..)``` is called after the initial operators have been populated and therefore your change seems quite reasonable.

manstis:
@jomarko Failing test ("org.drools.workbench.screens.guided.dtable.client.wizard.column.pages.OperatorPageTest.testOperatorPlaceholderSelectedWhenNothingSelectedPreviously") seems related.

jomarko:
@manstis yes, it is related, I looking right now.

jomarko:
@manstis green now.

manstis:
Thanks @jomarko

PR #539, opened 2017-06-30
GUVNOR-3244: GuidedRuleModelIndexVisitor throws NPE when indexing NeighborTable in the dinnerparty example

manstis:
See https://issues.jboss.org/browse/GUVNOR-3244

This PR changes "Guided Decision Table" indexing to use the generic DRL-indexer.


yurloc:
@manstis Looks good. I don't fully understand the code and the changes but I no longer see the reported NPE in server log after importing dinnerparty example. Are there any risks related to the code refactoring that I should focus on? Why is it possible to drop `GuidedDecisionTableModelIndexVisitor` and `GuidedRuleModelIndexVisitor`? Was there some kind of code duplication?

manstis:
Hi @yurloc The "old" ```Indexer``` implementation walked the Guided Decision Table Editor [model](https://github.com/kiegroup/drools/tree/master/drools-workbench-models/drools-workbench-models-guided-dtable/src/main/java/org/drools/workbench/models/guided/dtable/shared/model) and had the ```NPE``` issue for "eval" constraints. The "new" ```Indexer``` converts the GDT model into DRL which is then parsed by a generic DRL ```Descr``` walker to extract index information. The generic DRL is used for all DRL asset-types already so moving GDT to re-use it (a) removed some semantic duplication, (b) moved GDT to a more _robust_ part of the codebase.

manstis:
@yurloc Waiting for your approval. Thanks.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/test/java/org/drools/workbench/screens/guided/dtable/backend/server/indexing/TestGuidedDecisionTableFileIndexer.java
```diff
@@ -43,4 +51,34 @@ public void setResourceTypeDefinition( final GuidedDTableResourceTypeDefinition
         this.type = type;
     }
 
+    @Override
+    protected ProjectDataModelOracle getProjectDataModelOracle(final Path path ) {
+        final ProjectDataModelOracle dmo = new ProjectDataModelOracleImpl();
+        dmo.addProjectModelFields( new HashMap<String, ModelField[]>() {{
+            put( "org.drools.workbench.screens.guided.dtable.backend.server.indexing.classes.Applicant",
```
yurloc: @manstis The string can be replaced with `Applicant.class.getCanonicalName()`. From my point of view it has several advantages:
- will be refactored by the IDE when the class moves or is renamed
- it's shorter
- it's DRY

Please take this as a style recommendation/suggestion. I understand that consistency with other tests in this project is another concern so I'll leave it up to you to decide if this is worth the change.

manstis:
@yurloc Updated.. when green (again) I will merge. Thanks for your review.

PR #541, opened 2017-07-06
GUVNOR-3070: Guided Decision Tables: Column Wizard - Edit mode

karreiro:
See:
- [Guided Decision Tables: Column Wizard - Edit mode](https://issues.jboss.org/browse/GUVNOR-3070)
- [GDT Column Wizard: Hidden column indentation and font](https://issues.jboss.org/browse/GUVNOR-3051)

---

Add a Condition BRL fragment:
brlcondition

Add a simple Condition:
simplecondition

Add an Action BRL fragment:
actionaddbrlfrag

Delete an existing fact:
delete

Execute a Work Item:
executeworkitem

Set the value of a field:
setvalue

Set the value of a field with a Work Item parameter:
setvaluewithworkitem


karreiro:
@manstis @jomarko Could you please take a look? :-)

manstis:
@karreiro Great. I'll take a proper look tomorrow; but for now it looks like you elected to take the more simple route and just edit based on column type rather than infer which Wizard plugin to use based on pattern matching?

karreiro:
Thank for the review, Mike :-)

Regarding the `EventBus`. I think that I'm keeping the old behavior, see:
- https://github.com/karreiro/drools-wb/blob/3e45d08e25208a7e6fbc8f79694950cd26897f94/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/BRLConditionColumnPlugin.java#L361-L364
- https://github.com/karreiro/drools-wb/blob/3e45d08e25208a7e6fbc8f79694950cd26897f94/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/BRLConditionColumnPlugin.java#L401-L405
- https://github.com/karreiro/drools-wb/blob/3e45d08e25208a7e6fbc8f79694950cd26897f94/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/BRLActionColumnPlugin.java#L107-L110
- https://github.com/karreiro/drools-wb/blob/3e45d08e25208a7e6fbc8f79694950cd26897f94/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/BRLActionColumnPlugin.java#L326-L331

Here, you can see the scenario that you mentioned:
mikecommment

Am I missing something?

Thank you.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/PatternPage.java
```diff
@@ -107,6 +109,10 @@ void setupPattern() {
 
         setUpPatternList();
 
+        if (!isPatternCreationEnabled()) {
```
karreiro: I don't think so. I'm disabling the pattern creation only when the wizard is editing a "ActionSetFieldCol52" or a "ActionWorkItemSetFieldCol52" :-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/PluginHandler.java
```diff
@@ -0,0 +1,135 @@
...
+        final ConditionColumnPlugin plugin = conditionColumnPlugin.get();
+
+        plugin.setOriginalPattern52(pattern);
+        plugin.setOriginalColumnConfig52(column);
+
+        openWizard(plugin);
+    }
+
+    void openWizard(final DecisionTableColumnPlugin plugin) {
+        if (presenter.isReadOnly()) {
```
karreiro: Precisely. This is just a safe-guard.

All "newColumnX" had this check, so I think that would be reasonable to keep this verification. However, we can remove this, if you consider that it's redundant.

jomarko:

jomarko:

jomarko:
Well, it seems that jenkins has some troubles with building this PR. I will try locally.

jomarko:
missing_create_pattern_button
screenshot from 2017-07-11 10-33-19

karreiro:

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/AbstractLimitedEntryBRLColumnViewImpl.java
```diff
@@ -1,225 +0,0 @@
...
-        }
-        if ( isNew ) {
-            if ( !isHeaderUnique( editingCol.getHeader() ) ) {
-                Window.alert( GuidedDecisionTableConstants.INSTANCE.ThatColumnNameIsAlreadyInUsePleasePickAnother() );
-                return;
-            }
-            if ( isDefined() ) {
-                doInsertColumn();
-            } else {
-                Window.alert( GuidedDecisionTableConstants.INSTANCE.DecisionTableBRLFragmentNothingDefined() );
```
karreiro: @jomarko Done. Thank you.

karreiro:

karreiro:
Thank you for the review @jomarko. Regarding the issues that you raised:

- https://issues.jboss.org/browse/GUVNOR-3296
It's fixed.

- https://issues.jboss.org/browse/GUVNOR-3297
I don't think that it is a problem, we can discuss more in the ticket: https://issues.jboss.org/browse/GUVNOR-3297?focusedCommentId=13434751&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13434751

- https://issues.jboss.org/browse/GUVNOR-3298
It's fixed.

- https://issues.jboss.org/browse/GUVNOR-3277
It is a bug. However, since it's not a regression, we can solve it in another PR, see:
sdssdsd

Do you think that we're able to merge? :-)

manstis:
Great stuff @karreiro :-)

PR #545, opened 2017-07-14
GUVNOR-3070: Enhance tests

jomarko:
This PR should enhance tests:
- Which assert that completion of wizard (started for column editing)
  invokes the right operation on presenter
- Which assert that the completion of wizard does not depend on the
  column type (new, edited)

Next this PR removes some unused imports and do some changes needed for
mocking in tests.

@karreiro you will probably want to have a look.


karreiro:

jomarko:
Hi @karreiro I updated `PluginHandlerTest` to be again unit test and added verification if `generateColumn()` is invoked into `NewGuidedDecisionTableColumnWizardTest`. Particular `generateColumn()` in plugins are tested already well, thanks to you for this.

I also removed `verify(plugin, never()).isNewColumn();`. My intention was to test, if `isComplete()` check is independent on the column type (new/edited). However now I realized that it was probably not right place&way to test it. 

Thanks for review and please let me know if PR now meets your expectations.

karreiro:
Jenkins could you please retest this.

jomarko:
@karreiro tests should pass now, lets see.

PR #546, opened 2017-07-14
GUVNOR-3196: [Guided Decision Table] BRL Action offers bound fields for inappropriate actions

manstis:
See https://issues.jboss.org/browse/GUVNOR-3196

This also fixes the same (incorrect) behaviour in the Guided Rule Editor and Guided Template Editor.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/RuleModellerActionSelectorPopup.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/RuleModellerActionSelectorPopup.java
@@ -177,6 +177,7 @@ public class RuleModellerActionSelectorPopup extends AbstractRuleModellerSelecto
             addLogicalInsertions();
             addGlobalCollections();
             addFreeFormDRL();
+            addCallMethodOn();
             addCustomActionPlugins();
         }
 
@@ -184,7 +185,7 @@ public class RuleModellerActionSelectorPopup extends AbstractRuleModellerSelecto
     }
 
     // Add DSL sentences
-    private void addDSLSentences() {
+    void addDSLSentences() {
         //DSL might be prohibited (e.g. editing a DRL file. Only DSLR files can contain DSL)
         if (!ruleModeller.isDSLEnabled()) {
             return;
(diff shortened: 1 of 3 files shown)


manstis:
Part of an ensemble:-

- https://github.com/kiegroup/drools/pull/1372
- https://github.com/kiegroup/drools-wb/pull/546

manstis:
@jomarko Heh.. you noticed. Let me write one (I um'ed and er'ed and took the lazy path) ;-)

manstis:
@jomarko Tests added.

jomarko:
@manstis Thanks

drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/RuleModellerActionSelectorPopup.java
```diff
@@ -208,8 +209,8 @@ public void execute() {
     }
 
     //Add update, not modify
-    private void addUpdateNotModify() {
-        List<String> vars = model.getAllLHSVariables();
```
jomarko: @manstis I got back to this PR because of some fails in our selenium tests, I think this change was not needed. Could you please confirm? I will create a PR then.

drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/RuleModellerActionSelectorPopup.java
```diff
@@ -279,8 +280,8 @@ public void execute() {
     }
 
     //Add Modifies
-    private void addModifies() {
-        List<String> vars = model.getAllLHSVariables();
```
jomarko: @manstis I got back to this PR because of some fails in our selenium tests, I think this change was not needed. Could you please confirm? I will create a PR then.

manstis: @jomarko I think you are correct. I thought it did not make sense to be able to "modify" bound fields; but I've just tried with vanilla DRL and it seems quite valid. I also checked the Guided Rule Editor with 6.5.x and "modifying" a bound field works fine.

PR #547, opened 2017-07-14
GUVNOR-3143: [Guided Decision Table] Execute multiple Work Items

manstis:
See https://issues.jboss.org/browse/GUVNOR-3143

This PR prevents the same _Work Item_ from being executed twice.

It also fixes an issue I spotted whilst testing: A ```WID```'s "Display Name" was not used on the "Set field from Work Item parameter" plugin but was used by the "Execute Work Item" plugin.


jomarko:

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/plugins/ActionWorkItemPlugin.java
```diff
@@ -173,10 +173,25 @@ public PortableWorkDefinition getWorkItemDefinition() {
...
         getPresenter()
                 .getWorkItemDefinitions()
+                .stream()
+                .filter(this::isWorkItemAvailable)
                 .forEach(workDefinition -> biConsumer.accept(workDefinition.getDisplayName(),
                                                              workDefinition.getName()));
     }
 
+    private boolean isWorkItemAvailable(final PortableWorkDefinition pwd) {
+        if (isNewColumn()) {
```
jomarko: I don't agree with this `if` and the `testForEachWorkItemWhenOneHasBeenExecutedUpdatedColumn` test. Thanks to this you can still invoke two same WI. 

Follow these steps: 
- Define two WI (WorkItemOne, WorkItemTwo) definitions
- In the same package create GDT
- Add Execute WI column - select WorkItemOne
- Add Execute WI column - select WorkItemTwo
- Edit the first added column - you can select WorkItemTwo
- Now you are in state when two same WI are executed

manstis: @jomarko Yes I agree. Good find :-)

I only work on JIRAs relating to non-DMN things on Fridays at the moment; so will take another look later this week.

jomarko: Thanks for letting know, I have no objections, just let me know when I should have a look again.

manstis:
@jomarko Hi, updated and tested within the UI itself. Seems much better. Thanks!

PR #550, opened 2017-07-18
PLANNER-823: ScoreHolder support does not work in Guided Decision Tables

mcimbora:
Part of:
https://github.com/kiegroup/drools/pull/1373
https://github.com/kiegroup/drools-wb/pull/550
https://github.com/kiegroup/optaplanner-wb/pull/188

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/pom.xml
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/pom.xml
@@ -23,6 +23,10 @@
     </dependency>
     <dependency>
       <groupId>org.drools</groupId>
+      <artifactId>drools-workbench-models-commons</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
       <artifactId>drools-workbench-models-datamodel-api</artifactId>
     </dependency>
     <dependency>
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/GuidedDecisionTableSourceService.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/GuidedDecisionTableSourceService.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
 import javax.inject.Named;
 
@@ -29,6 +30,7 @@ import org.drools.compiler.lang.Expander;
 import org.drools.compiler.lang.dsl.DSLMappingFile;
 import org.drools.compiler.lang.dsl.DSLTokenizedMappingFile;
 import org.drools.compiler.lang.dsl.DefaultExpander;
+import org.drools.workbench.models.commons.backend.rule.RuleModelIActionPersistenceExtension;
 import org.drools.workbench.models.guided.dtable.backend.GuidedDTDRLPersistence;
 import org.drools.workbench.models.guided.dtable.shared.model.GuidedDecisionTable52;
 import org.drools.workbench.screens.guided.dtable.service.GuidedDecisionTableEditorService;
@@ -48,25 +50,37 @@ import org.uberfire.java.nio.file.Path;
 public class GuidedDecisionTableSourceService
         extends BaseSourceService<GuidedDecisionTable52> {
 
-    private static final Logger logger = LoggerFactory.getLogger( GuidedDecisionTableSourceService.class );
+    private static final Logger logger = LoggerFactory.getLogger(GuidedDecisionTableSourceService.class);
 
     private static final DSLFileFilter FILTER_DSLS = new DSLFileFilter();
 
-    @Inject
     private GuidedDTableResourceTypeDefinition resourceType;
 
-    @Inject
     private GuidedDecisionTableEditorService guidedDecisionTableEditorService;
 
-    @Inject
-    @Named("ioStrategy")
     private IOService ioService;
 
-    @Inject
     private FileDiscoveryService fileDiscoveryService;
 
-    @Inject
     private KieProjectService projectService;
+
+    private Collection<RuleModelIActionPersistenceExtension> persistenceExtensions = new ArrayList<>();
+
+    @Inject
+    public GuidedDecisionTableSourceService(final GuidedDTableResourceTypeDefinition resourceType,
+                                            final GuidedDecisionTableEditorService guidedDecisionTableEditorService,
+                                            final @Named("ioStrategy") IOService ioService,
+                                            final FileDiscoveryService fileDiscoveryService,
+                                            final KieProjectService projectService,
+                                            final Instance<RuleModelIActionPersistenceExtension> persistenceExtensionInstance) {
+        this.resourceType = resourceType;
+        this.guidedDecisionTableEditorService = guidedDecisionTableEditorService;
+        this.ioService = ioService;
+        this.fileDiscoveryService = fileDiscoveryService;
+        this.projectService = projectService;
+
+        persistenceExtensionInstance.iterator().forEachRemaining(persistenceExtensions::add);
+    }
 
     @Override
     public String getPattern() {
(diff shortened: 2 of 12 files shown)


mcimbora:
@manstis @jomarko This PR builds on the pluggable extensions created to support scoreHolder variable in the  Guided rule editor. Just needed to ensure we're able to define template variables inside 'scoreHolder' IAction implementations. Would you mind taking a look at the PRs? Thank you.

mcimbora:
(resolving conflicts)

manstis:
I'll look at these on Friday.. can you give a quick explanation of why pre-processing of "templates" is required?

manstis:
@mcimbora Would you mind addressing the conflicts before I review?

mcimbora:
@manstis Sure, just returned back from PTO. I'll have a look today.

mcimbora:
@manstis Conflicts resolved. During the pre-processing phase the variables are substituted for values, so that they can be marshaled by persistence extensions. I changed the method name to "preProcessIActionForExtensions" to better describe the purpose. Otherwise the persistence extensions (https://github.com/kiegroup/drools/blob/master/drools-workbench-models/drools-workbench-models-commons/src/main/java/org/drools/workbench/models/commons/backend/rule/RuleModelIActionPersistenceExtension.java) would need to know how to substitute the variables and I didn't want to pollute the interface. A matter of taste, I am open to discussion.

mcimbora:
@Rikkola I observed exceptions in the Verifier while converting payload from JSON to Object (no marshaller for type insert_planner_type_here). Is there anything specific you needed to do to register drools classes with the marshaller? I see the marshaller definition has been generated in .errai folder for the planner types, any idea?

jomarko:
jenkins please retest this

jomarko:

manstis:

manstis:
@karreiro @jomarko Fancy a look? It affects Decision Tables column wizard.

Rikkola:
@mcimbora Maybe you know this, but in case you don't going to explain how the Verifier is created.

The Verifier is built in drools-wb-services. It is a separate GWT application that we then run as a web service. So any imports added for any webapp are ignored. Now if you make changes to extend the Guided DTable Editor model, then we might have to move the Verifier to kie-wb-distributions or the extensions to drools where we have the dtable model.

mcimbora:
@jomarko Can you please elaborate on the meaning of vars translation? Do you mean substituting template variables by actual values? You can observe the behavior by defining a variable in a planner action, e.g. @{variable}. If you define an actual value in the DT Cell, the value of the cell substitutes the variable (e.g. when you use Source tab). Other approach to verify the behavior of detecting variables is to take look at generated source code of the DT asset (there's always variable/actual value element in the xml).

mcimbora:
@Rikkola That's unfortunate, moving planner extensions to drools introduces drools -> planner dependency. However moving the verifier to kie-wb-distributions means that the verifier can't be used by a standalone drools-wb app if I'm not mistaken. The latter one seems to be unacceptable, that leaves us with option #1? Not sure if that's acceptable for you drools guys? @manstis @jomarko Perhaps you see other options? Based on previous discussion with @Rikkola, the problem is with verifier marshaller not being able to detect planner classes, as it has no direct dependency on any of the planner modules.

manstis:
@Rikkola Why does the V&V code need Optaplanner extensions marshalled (presumably across the Web Worker boundary)? I assume it's because the decision table model is also used by the V&V code? If so, does it need to use it? Could it be sufficiently abstract to also (one day) provide V&V of DRL rules?

Ignoring the above which would (I suspect) require a huge refactor; I wonder how we could have the marshallers generated in Optaplanner (presumably in the ```-api``` or ```-client``` module) and registered with Errai via some CDI magic (outside of Errai) during workbench bootstrap. 

@mbarkley Hi.. any insights? (as far as I understand we have a class Z in module X that module Y tries to unmarshall JSON->object and cannot find a marshalling implementation to handle it. Module Y has no notion of the concrete implementation of Z just an interface). This is not a RPC call; but marshalling across browser process boundaries into a Web Worker. X is in one process, Y is in another. Z needs to live in both.

jomarko:
@mcimbora sorry that I was unclear, but you answered my question, thanks. Unfortunately I found two cases that don't work.
- Lets say there are two modify score actions (e.q. modify soft score and modify hard score) and both of those action use the same @{variable} expression. Even if you fill different values into DT cells just one of those value is used in the generated source
- Lets say there is one action (modify multiple score levels), use @{var1} for soft score, @{var2} for hard score, fill DT cells with some values, switch to source tab, error dialogue will be shown

Please let me know if you would like to fix these two issues as part of this PR or you would like to address them via two new jiras.

mcimbora:
@jomarko Thanks for the review, I'd prefer to fix the issue before it's merged. I'll let you know when the fix is ready.

mbarkley:
@manstis Here are the possible solutions I can think of on the Errai side:

1. Compile Z (where Z must be a portable class and not an interface or abstract class) in both modules.

2. In module Y where Z does not have a marshalling definition, add a [custom marshaller](http://docs.jboss.org/errai/latest/errai/reference/html_single/#sid-5931328_Marshalling-CustomMarshallers) that converts it to some other class with custom logic.

In either case, these are compile-time fixes as there isn't currently anyway to register new marshallers at runtime.

jomarko:
@mcimbora It seems that I can not reproduce the second issue above more. So please take a look just on the first one. Thanks and sorry for inconvenience.

mcimbora:
@jomarko I updated the optaplanner-wb PR and resolved the error related to using multiple score levels. As to the first issue related to sharing the same variable name across multiple columns, I created https://issues.jboss.org/browse/GUVNOR-3349 to resolve it later, as this issue is broader and affects other action types too (e.g. FreeFormLine).

@manstis @mbarkley Thanks for the ideas. Unfortunately the source of the Z class is not available in module Y, therefore a custom marshaller cannot be used. That leaves us with the following options:
1. Include planner classes in drools-workbench-models which verifier depends on
2. Mark the problem as a known issue, the verifier won't work when using optaplanner with the decision tables. Wait for an eventual errai support for runtime marshaller registration.
ATM I'm leaning towards second option, as this avoids drools -> planner dependency. Please let me know if that makes sense to you.

mbarkley:
@mcimbora Maybe I'm missing something but I still don't see why there is a need to register a marshaller at runtime. If your goal is to use an instance of class Z in module Y, surely Z must implement some interface Q that Y knows about, right? If that's the case, you could write a custom marshaller for Q that is backed by some other implementation of Q in Y.

mcimbora:
@mbarkley Thanks, I appreciate the help. Just to give some more context, the optaplanner classes (Z, about 10 different classes) that need to be marshalled on the drools side implement a _marker_ interface (Q), which is available in drools (Y), however the classes are really specific to optaplanner only (they represent score modification actions on the RHS of drools rules). e.g. [ActionHardConstraintMatch](https://github.com/mcimbora/optaplanner-wb/blob/cbbd295ce0184698ae4dbd054465cb1977e91e3a/optaplanner-wb-screens/optaplanner-wb-guided-rule-editor/optaplanner-wb-guided-rule-editor-api/src/main/java/org/optaplanner/workbench/screens/guidedrule/model/ActionHardConstraintMatch.java).

They all have internal state, which representation varies a lot (can be a string variable, a list of strings, etc). The internal state cannot be lost while unmarshaling the object, as the object(s) is/are a part of the Decision Table model, which need to be reconstructed. That makes me believe you'd need a separate drools counterpart for each of those planner classes, which is basically equal to copying the classes to drools module.

How does the planner class get to the Decision Table model, which resides in drools? We use pluggable extensions that insert the custom optaplanner objects into the decision model via CDI. The planner classes are visible in webapp application, however the drools verifier runs as a separate application, therefore the planner classes aren't visible.

mbarkley:
@mcimbora Thanks for the extra context. I have more thoughts on the subject, but unfortunately I am about to sign off and will be on PTO until August 3. Perhaps we can pick up this discussion then? The one thing I will need to understand better is what types are in different GWT modules, versus what types are in separately compiled GWT scripts.

mcimbora:
@mbarkley Sure, sounds good. Let's discuss after you return back then.

@manstis Can we get this in meanwhile? The verifier problem is the only missing part that I'll prioritize once we've come up with a good solution.

manstis:
@mcimbora Yes, these can be merged when they're all green.

Rikkola:
@manstis @mcimbora 

The web worker does the indexing of the model. But we could do this in main app or in server side. I would actually prefer server side. This way the index that V&V uses would be reusable and there would be speed improvements.
And we can V&V anything that can be mapped into the index. So any rule format.

My original plan was to index in the main web application, but due to having spent a billion hours on the task, indexing in web worker was the way to get the work done fastest.

Since moving the indexing anywhere takes so long I propose we do a temporal fix and once I'm done with my current work I return to this and do what is right.

mcimbora:
@Rikkola Sounds good. Let me know if I can help you in any way.

jomarko:
@manstis Thanks for merging. @mcimbora I see this feature is planned for 7.0.0.LA product release, please could you cherry-pick for 7.2.x branch?

mcimbora:
@jomarko Sure, I'll do in the afternoon. Just need to set up the env and check if everything works fine.

jomarko:
@mcimbora thanks

PR #555, opened 2017-07-24
GUVNOR-3318 : [GSS](6.4.x)Incorrect Warning Messages for Conflicting Rows in Guided DT

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3318

Had to redo the fix for 6.5.x since the branches are different. Here single hit for example is not fixed, since this branch does not have a check for single hit.

Also weird how the same test finds bugs from both branches and the cause is different ;-)


Rikkola:
@jomarko Hi, if you could please take a look.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/condition/BRLConditionInspector.java
```diff
@@ -21,37 +21,48 @@
...
+        this.valueAsString = valueAsString;
     }
 
     @Override
-    public boolean conflicts( final Object other ) {
+    public boolean conflicts(final Object other) {
+        if (other instanceof BRLConditionInspector) {
+            final String otherValue = ((BRLConditionInspector) other).valueAsString;
+            if (otherValue != null) {
+                return !otherValue.equals(valueAsString);
```
jomarko: Could I ask you for explanation when should the method `conflicts` return `true` and when `false`.

Rikkola: This is a poor way to test this. The real way would require us to index the actual DRL in the condition/action, but this is something that needs more work.

This way does get us closer. If the string values for any two fields are different, then we say they conflict. In the 6.5.x verifier we do comparison by column, so the DRL for is the same. This should not give any false error reports. 

In master the fix was different. There we look deeper into the BRLColumn, but still even there we should look at the DRL and use that for more complete comparison. Now once we have the DRL indexed properly we could just give in any DRL and we get a complete Verifier for any DRL based asset.

PR #559, opened 2017-07-27
[Guided Rule Editor] Unable to construct From Collect

jomarko:
Part of ensemble:
kiegroup/kie-wb-common#997

@manstis Definitely something for your eyes. Some tests maybe still needed to add and code style is not applied here, but I wanted to ask you if my approach is correct. Now FromCollect widget should offer collections that are in project settings && [(in imports of the given asset  (guided rule / template / table) || in the same package as the asset ]

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/widget/FromCollectCompositeFactPatternWidget.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/widget/FromCollectCompositeFactPatternWidget.java
@@ -51,22 +51,22 @@ public class FromCollectCompositeFactPatternWidget extends FromCompositeFactPatt
 
     private Map<String, String> extraLeftSidePatternFactTypes = null;
 
-    public FromCollectCompositeFactPatternWidget( RuleModeller modeller,
-                                                  EventBus eventBus,
-                                                  FromCollectCompositeFactPattern pattern ) {
-        super( modeller,
-               eventBus,
-               pattern );
+    public FromCollectCompositeFactPatternWidget(RuleModeller modeller,
+                                                 EventBus eventBus,
+                                                 FromCollectCompositeFactPattern pattern) {
+        super(modeller,
+              eventBus,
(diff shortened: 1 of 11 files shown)


drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-backend/src/test/java/org/drools/workbench/screens/guided/template/server/GuidedRuleTemplateEditorServiceImplTest.java
```diff
@@ -0,0 +1,96 @@
+package org.drools.workbench.screens.guided.template.server;
```
manstis: License

jomarko:
@manstis I think I applied your comments and also I am happy with coverage now.

drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/widget/FromCollectCompositeFactPatternWidget.java
```diff
@@ -51,269 +51,239 @@
...
-
-    private boolean modelImportsClass( final String fullyQualifiedClassName ) {
-        for ( Import i : modeller.getModel().getImports().getImports() ) {
-            if ( i.getType().equals( fullyQualifiedClassName ) ) {
-                return true;
-            }
-        }
-        return false;
+        modeller.getDataModelOracle()
+                .getAvailableCollectionTypes()
```
hasys: It looks like this call break the Full Downstream Build

```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/kgaevski/Workspace/github/drools-wb/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/widget/FromCollectCompositeFactPatternWidget.java:[75,17] cannot find symbol
  symbol:   method getAvailableCollectionTypes()
  location: interface org.kie.workbench.common.widgets.client.datamodel.AsyncPackageDataModelOracle
[INFO] 1 error
```

PR #567, opened 2017-08-02
GUVNOR-3333, GUVNOR-3330: [Guided Decision Table] Edit condition column / Pattern change in condition column

karreiro:
See:
- https://issues.jboss.org/browse/GUVNOR-3330
- https://issues.jboss.org/browse/GUVNOR-3333

---

### GUVNOR-3330

Before:
3330-before

After:
3330-after

---

### GUVNOR-3333

Before:
3333-before

After:
3333-after

PR #569, opened 2017-08-04
[7.2.x] GUVNOR-3333, GUVNOR-3330: [Guided Decision Table] Edit condition column / Pattern change in condition column

karreiro:
Cherry picked from https://github.com/kiegroup/drools-wb/pull/567


jomarko:
Downloading downstream build. I will approve in few minutes.

PR #575, opened 2017-08-10
[6.5.x] RHBRMS-2889: [XLS Decision Table] Converted table with static fields is not opened

manstis:
See https://issues.jboss.org/browse/RHBRMS-2889


jomarko:
@manstis you are the fastest programmer I know. Thanks for the fix. Changes looks reasonable. However i would like to check also manually if you don't mind.

jomarko:

jomarko:
well jenkins ignores me, I will build locally

jomarko:
@manstis after conversion the GDT editor is opened successfully. However here comes one additional question. If you have a look on the attached xls decision table in the jira, it uses public static int constants as in condition column so in action column, however the conversion assumes as invalid int just value form the action column. For the condition column is the int value assumed as string. I mean the result of conversion is like shown bellow, where status is int. Please is this correct?

screenshot from 2017-08-11 09-20-31

manstis:
@jomarko Weird. When I convert I get this:

rhbrms-2889

jomarko:
@manstis well, I have to double check then, I will let you know.

manstis:
@jomarko Great. Please let me know when QE ACK for 6.4.5(?) and I'll merge.

manstis:
heh.. I see it's now approved!

jomarko:
@manstis sure, I already asked eligible people to put qe ack into jira.

PR #584, opened 2017-08-22
GUVNOR-3145: Add column\Edit column buttons should be disabled in new Graph

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3145

Demo


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableModellerViewImpl.java
```diff
@@ -244,6 +249,18 @@ public HandlerRegistration addKeyDownHandler(final KeyDownHandler handler) {
         return gridPanel.addKeyDownHandler(handler);
     }
 
+    void enableButtonMenu() {
+        getSubMenu().setVisible(true);
```
manstis: Enabling/disabling method names; however you're changing the visibility. It'd be nice to have the method names and implementation be consistent (personally I prefer to disable the buttons rather than hide as the User can still _see_ there's an action they can perform but it's disabled).

karreiro:
Thank you for the review, @manstis. I did the changes that you suggested. :-)

Demo

PR #585, opened 2017-08-22
GUVNOR-3413: [Guided Decision Table Editor] Mouse Pan restriction does not work correctly when table is smaller than visible space

manstis:
See https://issues.jboss.org/browse/GUVNOR-3413


manstis:
To test:
1) Open "mortgages"
2) Open "Pricing loans"
3) Double-click header to enter "pinned mode"
4) Drag canvas around. 

You should not be able to pan beyond extents of the table (i.e. by default, for the test given, you should not be able to pan the canvas at all. If however you add more rows then vertical panning within the extent of the table should be possible; or you resize columns to make it overflow the width of the visible region).

manstis:
@karreiro I noticed this when looking at your PRs for the scrollbars.

manstis:
@jomarko You are right. I cannot replicate it with ```master``` either.

@karreiro I suspect I ran ```drools-wb-webapp``` with your "Visual Indicator" changes. Please me sure to check the panning when pinned with your changes when considering the comments I made on your PR. The ```RestrictedMousePanMediator``` has different ```TransformMediator``` set depending on whether the table is pinned or not. It might be your changes remove/alter this behaviour.

Closing this PR.

PR #592, opened 2017-08-30
Fix for a broken super dev mode

Rikkola:
Not sure what commit removed this, but this needs to go too.


Rikkola:
Weird. Wasted my morning with this. First rebased my topic branches with the latest and tried drools-wb. Got the error about this.
Then did a fresh build with master from uberfire->drools-wb and it was again there. Nuked my temp files, refreshed Idea, reloaded and created Maven sources in Idea. Still failed.

Removed this one line and it worked.

Still, after all this I am not at all sure that it is needed.

@jomarko I just built everything from latest master up to drools-wb and then tried the Idea hosted mode, but it should happen with gwt:run too. So looks like this was an issue with Idea, but I have absolutely no idea what I still need to refresh. Even redid the indexing.

Rikkola:
Just now not getting it in my topic branch (argh!). I'll close this for now.

Rikkola:
The issue was that Idea could not find the gwt.xml file for refactoring client.

jomarko:
@Rikkola I am sorry that it has slowed down your work today, unfortunately I haven't noticed the PR immediately when you have opened it.

Rikkola:
@jomarko No need for you to be sorry ;-) And I did this PR after I got it working. So no time wasted by review speeds.

PR #595, opened 2017-09-05
GUVNOR-3344: [Guided Decision Table] Editing BRL columns does not work

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3344

Before:
before

After:
after

---

Part of an ensemble:
- https://github.com/AppFormer/uberfire/pull/821
- https://github.com/kiegroup/drools-wb/pull/595

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/BRLActionColumnSynchronizer.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/BRLActionColumnSynchronizer.java
@@ -38,121 +38,126 @@ public class BRLActionColumnSynchronizer extends BaseColumnSynchronizer<BaseColu
     }
 
     @Override
-    public boolean handlesAppend( final MetaData metaData ) {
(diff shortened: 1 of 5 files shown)


karreiro:
Jenkins execute full downstream build.

karreiro:
Hi @manstis,

Thanks for the review. You're right, sorting the columns by their new indexes messes the order defined by the user up.

The process of updating a BRL column is based on inserting a new column (with the updated data) and removing the old one. However, this process was leaving the columns with invalid indexes.

See how the indexes looked like before:
https://gist.github.com/karreiro/e0f86cca63fa6ca00f94b63d280423fe

And how they look now:
https://gist.github.com/karreiro/5c8ad34b9396d9e4f2ba0c98badc692c

PS.: before the fix, if we would save and re-open the Guided Decision Table, the indexes would be already "re-indexed".

karreiro:
Hi @jomarko, nice catch, thanks! It's fixed for "Action BRL columns" too now.

### BRL Action columns

Before:
before-act

After:
after-act

### BRL Condition columns:

Before:
before-cond

After:
after-cond

jomarko:
https://webmshare.com/yYLmW

karreiro:
Hi @manstis and @jomarko,

I'm closing this PR and opening another one.

Instead of "adjusting" the indexes that were being messy by the update process. I found what was messing them.
(I needed to do this because the reindex approach had some undesirable collateral effects).

During the update process of a column, we're creating new columns *and* deleting the old ones. The problem is that the wrong columns were being removed.

Now everything should be fixed by this PR (there's a reproducer there too):
https://github.com/AppFormer/uberfire/pull/845

:-)

PR #599, opened 2017-09-07
GUVNOR-3472: [Guided Decision Table] GDT Graph creation causes Error Popup

manstis:
See https://issues.jboss.org/browse/GUVNOR-3472

@jomarko This PR also fixes an existing "linking" test.

Test ```fieldConstraintWithActionBRLFragmentFieldWithoutTemplateKey()``` never formed a link and hence the ascertain never executed and never failed. The reason was that ```BRLActionColumn``` with no Template Keys (and hence a boolean column) should have links checked against the definition itself.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-api/src/main/java/org/drools/workbench/screens/guided/dtable/shared/DefaultGuidedDecisionTableLinkManager.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-api/src/main/java/org/drools/workbench/screens/guided/dtable/shared/DefaultGuidedDecisionTableLinkManager.java
@@ -16,6 +16,20 @@
 
 package org.drools.workbench.screens.guided.dtable.shared;
 
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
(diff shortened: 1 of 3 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-api/src/main/java/org/drools/workbench/screens/guided/dtable/shared/DefaultGuidedDecisionTableLinkManager.java
```diff
@@ -33,91 +47,165 @@
...
-            } else if ( ac instanceof BRLActionColumn ) {
+            } else if (ac instanceof BRLActionColumn) {
                 final BRLActionColumn fragment = (BRLActionColumn) ac;
-                for ( BRLActionVariableColumn var : fragment.getChildColumns() ) {
-                    if ( factType.equals( var.getFactType() ) && fieldName.equals( var.getFactField() ) ) {
-                        return ac;
+
+                if (hasTemplateKeys(fragment)) {
+                    //If the fragment has Template Keys we can attempt to link child columns
+                    for (BRLActionVariableColumn var : fragment.getChildColumns()) {
```
karreiro: I think that this branch is not covered by tests.

manstis:
@jomarko Both suggestions applied.

PR #600, opened 2017-09-07
GUVNOR-3473: Add Pinned mode indicator

jomarko:
screenshot from 2017-09-07 23-04-39
screenshot from 2017-09-07 23-04-50
@manstis @karreiro please have a look.


manstis:
How about using pushpin icon when pinned and nothing otherwise. We use the lock icon for when the file is locked (by another User).

manstis:
I'll look more tomorrow but I am thinking to adjust the shown icon in the enter/exit pinned mode methods.. but it might be obvious to me tomorrow why you out it where you have!

manstis:
I should add.. it looks ok on the whole :-)

jomarko:
@manstis I had exactly this icon in mind, unfortunately I haven't found it in `org.gwtbootstrap3.client.ui.constants.IconType`, But sure I am open to any icon, also that in standard mode no icon is shown. Lets agree on the final solution in the morning.

jomarko:
pinmode

@manstis Icon shown just in pinned mode, I used icon that was the most similar to pin I know from IDEA. Let me know if you are aware about better/nicer/more self explanatory icon. Thanks.

manstis:
Nice improvement @jomarko 

Shame we can't get the icon we'd like.. I tried locally to no end :-(

jomarko:
@karreiro sending reward to you right now. Thanks I will amend the commit.

PR #605, opened 2017-09-12
GUVNOR-3471: [Guided Decision Table] Delete Multiple Cells and Rows f…

markcoble:
…rom context menu

NOTE:  only change to GuidedDecisionTableModellerContextMenuSupport is to line 79. Setting the getControlKey event to true works for the use case of selecting multiple cells with CTRL and either keeping it held down or not before right clicking to display the context menu.   RangeSelectionStrategy.handleSelection will then not remove selected cells as it was before if the user let go of CTRL before right clicking.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableModellerContextMenuSupport.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableModellerContextMenuSupport.java
@@ -15,6 +15,8 @@
  */
 package org.drools.workbench.screens.guided.dtable.client.widget.table;
 
+import java.util.stream.Stream;
+
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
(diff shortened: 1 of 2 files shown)


manstis:
screenshot from 2017-09-13 09-46-51

jomarko:
@manstis good catch. I agree with you, Right mouse click shouldn't extend the set of selected cells/rows. 
@markcoble otherwise the fix looks good.

markcoble:
@manstis @jomarko Still need to add JUnit tests but have updated the logic according to these discussions.  Please verify functionality and I'll add tests.

manstis:
@markcoble Looks much better.. I'll test in the application tomorrow.

markcoble:
@manstis @jomarko thanks for feedback.  added two test cases to cover pos/neg scenarios.

drools-wb-webapp/src/main/webapp/META-INF/maven/org.drools/drools-wb-webapp/pom.properties
```diff
@@ -0,0 +1,7 @@
+#Generated by Maven Integration for Eclipse
```
manstis: Please delete this file.

drools-wb-webapp/src/main/webapp/META-INF/maven/org.drools/drools-wb-webapp/pom.xml
```diff
@@ -0,0 +1,1330 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
manstis: Please delete this file too!

markcoble:
@manstis @jomarko sorry for those extra adds.  normally add all is not a problem!  Sorry for the slight formatting differences.  Once the Eclipse/IDEA formatting stuff it agreed upon all these will have to be brought in line.  For instances spaces around method arguments/parameters I think is no longer desired so my updates have removed them.  @jomarko requested tests were added though for what I was trying to test don't think they were necessary.

manstis:
Jenkins please retest this.

@markcoble Could you please let me know when this is green and I'll merge. Thanks.

PR #608, opened 2017-09-14
GUVNOR-3491: [Guided Decision Table] Context menu doesn't disappear if editor toolbar is clicked and even if the editor is closed

manstis:
See https://issues.jboss.org/browse/GUVNOR-3491


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableModellerViewImplTest.java
```diff
@@ -284,12 +291,28 @@ public void testAddKeyDownHandlerAttachesToEditor() {
...
-    public void testEnableButtonMenu() {
+    public void testAddMouseDownHandlerAttachesToRootPanel() {
+        //Ensure nobody thinks its a good idea to attach to the GridPanel at some time in the future!
+        //See https://issues.jboss.org/browse/GUVNOR-3491
+        final MouseDownHandler handler = mock(MouseDownHandler.class);
+
+        view.addMouseDownHandler(handler);
+
+        verify(mockGridPanel,
+               never()).addMouseDownHandler(eq(handler));
```
jomarko: Looking on the changed code in implementation, you could also add:
```java
verify(mockGridPanel, never()).addDomHandler(eq(handler), eq(MouseDownEvent.getType()));
```

But I know, there is potentially infinite space of methods that could be asserted in similar way. I am fine if you keeps as it is.

PR #613, opened 2017-09-15
GUVNOR-3446: Test successful loading of scenario from xml

jomarko:
@Rikkola I will be pleased by your review.


jomarko:
jenkins please retest this

PR #615, opened 2017-09-15
GUVNOR-3447: Run scenario with possible transitive imports

jomarko:
@Rikkola Hi, the added test in this PR will fail. I would like to ask you if this is expected behavior. The scenario has import for **Person**. As the model field **Person** has attribute **Address**, the **Address** is added as dependent import during run. However model field **Address** has attributes **City** and **Street**, shouldn't be those **City** and **Street** added as dependent imports during scenario run?

### DIFF
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioTestEditorServiceImplTest.java
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioTestEditorServiceImplTest.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import org.appformer.project.datamodel.imports.Import;
 import org.appformer.project.datamodel.imports.Imports;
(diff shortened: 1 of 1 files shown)


Rikkola:
@jomarko I feel like they should all work. You are using full names with the package included. That should be enough to pick them up. So looks like this PR finds a bug.

jomarko:
@Rikkola feel free to close. I also don't like the feeling when PR related to me is opened for a long time. I generated also git patch of this commit and uploaded into the jira GUVNOR-3447, so it shouldn't be problem to reintroduce tests from this PR.

Rikkola:
@jomarko Thanks, but they don't trust me enough to access the close button ;-) You should be able to see it.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioTestEditorServiceImplTest.java
```diff
@@ -314,6 +314,37 @@ public void testCopyWithTargetThrowException() throws Exception {
     }
 
     @Test
+    public void testRunScenarioDependentImportsAndFactData() throws Exception {
+        final ArrayList<Fixture> fixtures = new ArrayList<Fixture>();
```
triceo: DRY, use diamond operator. (Also on line 325.)

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioTestEditorServiceImplTest.java
```diff
@@ -314,6 +314,37 @@ public void testCopyWithTargetThrowException() throws Exception {
...
+        when(scenario.getFixtures()).thenReturn(fixtures);
+        when(dataModelService.getDataModel(path)).thenReturn(modelOracle);
+        when(modelOracle.getProjectModelFields()).thenReturn(modelFields);
+        when(scenario.getImports()).thenReturn(imports);
+
+        testEditorService.runScenario("userName",
+                                      path,
+                                      scenario);
+
+        assertEquals(7,
```
triceo: `Assertions.assertThat(scenario.getImports().getImports()).hasSize(7);`

Also, wouldn't be bad to name the constant, so that the future generations know why 7.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioTestEditorServiceImplTest.java
```diff
@@ -314,6 +314,37 @@ public void testCopyWithTargetThrowException() throws Exception {
...
+        final ArrayList<Fixture> fixtures = new ArrayList<Fixture>();
+        final Imports imports = new Imports() {{
+            addImport(new Import("java.sql.ClientInfoStatus"));
+            addImport(new Import("Integer"));
+            addImport(new Import("org.kie.Person"));
+        }};
+
+        final Map<String, ModelField[]> modelFields = new HashMap<String, ModelField[]>() {{
+            put("java.sql.ClientInfoStatus",
+                new ModelField[]{modelField("java.sql.JDBCType")});
```
triceo: Looks like boilerplate.

Consider introducing `static ModelField[] toModelFields(String... m)` to hide the complexity.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioTestEditorServiceImplTest.java
jomarko: Sure, np.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioTestEditorServiceImplTest.java
jomarko: Will apply.

jomarko:
Comments incorporated, git patch recreated and uploaded into GUVNOR-3447. Closing for now, will be reopened when fix available.

PR #623, opened 2017-09-19
[7.3.x] GUVNOR-3474: grouping columns in management section (#602)

jomarko:
See https://issues.jboss.org/browse/GUVNOR-3474

This is cherry-pick of #602 

@karreiro please have a look when you have a moment.


karreiro:
@jomarko I think that we have some conflicts to solve. Could you please take a look? :-)

jomarko:
@karreiro I think I resolved it, lets wait for tests result for sure.

PR #636, opened 2017-09-29
GUVNOR-3200: [Guided Decision Table] Access control to prohibit business users from modifying columns

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3200

---

permissoes

---

Part of an ensemble:
- https://github.com/AppFormer/uberfire/pull/861
- https://github.com/kiegroup/kie-wb-common/pull/1190
- https://github.com/kiegroup/drools-wb/pull/636
- https://github.com/kiegroup/kie-wb-distributions/pull/628
- https://github.com/kiegroup/kie-docs/pull/402

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableModellerPresenter.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableModellerPresenter.java
@@ -413,16 +413,45 @@ public class GuidedDecisionTableModellerPresenter implements GuidedDecisionTable
         if (isGridPinned()) {
             view.getGridLayerView().flipToGridWidget(dtPresenter.getView());
         }
+
+        handlePermissions(dtPresenter);
+    }
+
+    void handlePermissions(final GuidedDecisionTableView.Presenter dtPresenter) {
+        if (dtPresenter.hasEditableColumns()) {
+            view.enableColumnOperationsMenu();
+        }
     }
 
     void refreshDefinitionsPanel(final GuidedDecisionTableView.Presenter dtPresenter) {
         final GuidedDecisionTable52 model = dtPresenter.getModel();
-        view.setEnableColumnCreation(dtPresenter.getAccess().isEditable());
+
+        view.setEnableColumnCreation(isColumnCreationEnabled(dtPresenter));
         view.refreshAttributeWidget(model.getAttributeCols());
         view.refreshMetaDataWidget(model.getMetadataCols());
         view.refreshConditionsWidget(model.getConditions());
         view.refreshActionsWidget(model.getActionCols());
         view.refreshColumnsNote(dtPresenter.hasColumnDefinitions());
+    }
+
+    boolean isColumnCreationEnabled(final GuidedDecisionTableView.Presenter dtPresenter) {
+
+        final boolean decisionTableIsEditable = !dtPresenter.isReadOnly();
+        final boolean decisionTableHasEditableColumns = dtPresenter.hasEditableColumns();
+
+        return decisionTableHasEditableColumns && decisionTableIsEditable;
+    }
+
+    @Override
+    public boolean isColumnCreationEnabledToActiveDecisionTable() {
+
+        final Optional<GuidedDecisionTableView.Presenter> decisionTable = Optional.ofNullable(getActiveDecisionTable());
+
+        if (!decisionTable.isPresent()) {
+            return false;
+        }
+
+        return isColumnCreationEnabled(decisionTable.get());
     }
 
     @Override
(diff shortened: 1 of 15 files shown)


manstis:
Read-only will also stop Users from editing cell values. The JIRA asks to restrict column definition  (presumably add, edit and delete. Not sure about move). Sorry. I think this needs more work :-(

karreiro:
@manstis Thank you for the comment, Mike.

I think that all restrictions are covered :-)
demo

manstis:
But the User cannot edit cell values either?

karreiro:
@manstis Ohh! I understand now. Thanks, Mike. I'm going to work this.

manstis:
Great :-) Sorry to "piss on your parade"

manstis:
It might be an idea to extend/use the Access class to support disabling column add, edit, delete operations (in addition to read-only).

karreiro:
Are you talking about the `WorkbenchFeatures` class?

...and when you say "extend/use the Access class", you mean having something like the following permissions there?

```java
String ADD_GUIDED_DECISION_TABLE_COLUMN = "guideddecisiontable.add";

String EDIT_GUIDED_DECISION_TABLE_COLUMN = "guideddecisiontable.edit";

String DELETE_GUIDED_DECISION_TABLE_COLUMN = "guideddecisiontable.delete";
```

Sounds like a good idea :-)

manstis:
I'm on my phone so not got the code at hand; but there's an inner class called Access that handles read-only and is IIRC a property of the dtable presenter that could be modified to handle "edit columns" permissions.

karreiro:
Cool, Mike. Thanks :-)

karreiro:
@manstis @jomarko Changes applied :-) See:

role-demo
projects-demo

Could you please take a look?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableModellerPresenter.java
```diff
@@ -413,18 +413,47 @@ public void execute(Collection<String> availableParentRuleNames) {
...
+        final Boolean decisionTableIsEditable = !dtPresenter.isReadOnly();
+        final Boolean decisionTableHasEditableColumns = dtPresenter.hasEditableColumns();
+
+        return decisionTableHasEditableColumns && decisionTableIsEditable;
+    }
+
+    @Override
+    public boolean isColumnCreationEnabledToActiveDecisionTable() {
+
+        final Optional<GuidedDecisionTableView.Presenter> decisionTable = Optional.ofNullable(getActiveDecisionTable());
```
manstis: I'm sure you could do something more funky with ```Optional``` (than check ``isPresent()```) but I'm old fashioned and am happy with this :-) Just adding a note as someone (younger!) somewhere might say something!

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
```diff
@@ -335,6 +345,19 @@ void initialiseContent(final ObservablePath path,
         initialiseAuditLog();
     }
 
+    void initialiseAccess(final boolean isReadOnly) {
+        getAccess().setReadOnly(isReadOnly);
+        getAccess().setHasEditableColumns(canEditColumns());
+    }
+
+    Boolean canEditColumns() {
```
manstis: ```Boolean``` for a reason (vs ```boolean```)? Just asking.. Use of ```Boolean``` in my mind always makes me think this is a tri-value ```boolean``` (```true```, ```false```, ```null```)..

jomarko:
@karreiro code changes looks good, please just incorporate Mike's findings with menu items.

karreiro:
@manstis @jomarko Changes applied :-) See:

demo

Thank you.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableModellerPresenter.java
karreiro: ;-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
karreiro: Personally, I think that avoiding primitives is more elegant, handling boolean values like any other object in the code base. Additionally (again personally) I try to use enums when something has more than two states. However, I don’t have a strong opinion about this, so I preferred to change it to primitive :-)

manstis:
@jomarko Regarding your question.

We're simply using the underlying AppFormer "permissions" framework to check whether a User has access to a _protected_ resource (the ability to edit columns). I guess, if a User belongs to multiple groups, one with permission and one without the underlying framework is finding a single "denied" and hence concluding the User does not have access.

I think conflict resolution issues (granted by one group/role; and denied by another) would be a possible JIRA for conflict resolution. That said it looks like the framework handles conflicts by allowing "priority" to be defined in the UI for different groups/roles; so I suspect the issue you found can be resolved by ensuring the "grant" group/role has higher priority than the "denied" group/role.

jomarko:
@manstis just letting you know. I discussed permissions with Tomas David. In cases when different roles assign different value for the same permission, the result permission is computed according to the role priority. It was my case, no new Jira needed.

PR #640, opened 2017-10-05
GUVNOR-3448: Use real KieSession during runnning scenario in tests

jomarko:
@Rikkola @manstis This PR is POC, I would like to add more similar tests. SImilar tests should replace selenium tests which just open prepared scenario in some project  and run the scenario. Definitely not all of them will be replaced.


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/pom.xml
```diff
@@ -172,6 +172,19 @@
...
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-workbench-models-guided-dtable</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.workbench.playground</groupId>
+      <artifactId>mortgages</artifactId>
+      <version>8.0.0-SNAPSHOT</version>
```
jomarko: @manstis @Rikkola how could/should I keep the proper version here? Any ideas?

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerServiceTest.java
```diff
@@ -17,45 +17,83 @@
...
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 @RunWith(MockitoJUnitRunner.class)
 public class ScenarioRunnerServiceTest {
 
+    private static final String MORTGAGES_HOME = "https://github.com/kiegroup/kie-wb-playground/tree/master/mortgages";
```
jomarko: @manstis @Rikkola how could/should I keep the up to date location here? Any ideas? I mean, the url now contains branch master, what could cause problems if this PR is cherry picked to other branch.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerServiceTest.java
```diff
@@ -65,10 +103,17 @@ public void setUp() throws Exception {
                                             projectService,
                                             scenarioLoader);
 
-        when(sessionService.newDefaultKieSessionWithPseudoClock(any(KieProject.class))).thenReturn(defaultPseudoClockKieSession);
-    }
+        final KieServices kieServices = KieServices.Factory.get();
+        final ReleaseId releaseId = kieServices.newReleaseId("org.kie.workbench.playground",
+                                                             "mortgages",
+                                                             "8.0.0-SNAPSHOT");
```
jomarko: Same question about version

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerServiceTest.java
manstis: @jomarko A JAR is created for https://github.com/kiegroup/kie-wb-playground which forms the examples that can be imported into the Workbench. It is versioned along with all other KIE artifacts. The Workbench includes the JAR as a dependency and unzips it to _load_ an example repository. You could consider doing the same for _mortgages_.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerServiceTest.java
manstis: Do you know if this has to match the version in the _mortgages_ project ```pom```?

I have a feeling it could be whatever you like; and therefore values ```test```/```test```/```1.0.0-SNAPSHOT``` could be fixed (and not matter.. like a lot of the KIE CI tests do.. e.g. https://github.com/kiegroup/drools/blob/master/kie-ci/src/test/java/org/kie/scanner/KieModuleBuilderTest.java#L61)

Rikkola:
These are probably good to do. If we ever get the simulation tests done I imagine this type of tests will help with making sure migration works.

manstis:
@jomarko I changed the PR title as I assume you do not want it merged at present.

jomarko:
@manstis totally right, thanks.

jomarko:
@manstis @Rikkola I would like to ask you for re-review, I changed the logic in that way, no versions management is needed and no whole kie projects needed.

manstis:
Jenkins please retest this.

PR #646, opened 2017-10-09
GUVNOR-3513: Guided Decision table V&V: Results are completely random

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3513


drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/relations/ConflictResolver.java
```diff
@@ -22,87 +22,89 @@
...
 
-        for ( final Object other : collection ) {
-            return isConflicting( isConflicting,
-                                  ( HasUUID ) other );
+        for (final Object other : collection) {
+            final Conflict conflicting = isConflicting(isConflicting,
+                                                       (HasUUID) other);
+            if (conflicting.foundIssue()) {
+                return conflicting;
+            }
```
Rikkola: The fix is here. We need to check all the items for an conflict.

Rikkola:
@manstis You might want to check my one liner. +1 from me for everything else.

Rikkola:
@jomarko Here is one fix @manstis and I worked together on.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-backend/src/test/java/org/drools/workbench/services/verifier/webworker/client/fromfile/DecisionTableAnalyzerFromFileTest.java
```diff
@@ -142,6 +142,24 @@ public void testFile3() throws
     }
 
     @Test
+    public void testCrazyConflictMessages() throws
```
manstis: I know it was volition but perhaps a better test name would be good?

Rikkola: Like "testReallyCrazyConlictMessages()" ?

Ok ok, I'll try to think something that describes this.

manstis:
@jomarko I assume you'd like to review too?

jomarko:
@manstis sure, I will have a look.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/relations/ConflictResolver.java
jomarko: I just wonder if the first `conflicting` that has `foundIssue() == true` is enough to return. As the method `getConflictingObjects` suggests me some list of conflicting objects should be returned? Shouldn't be the list `InspectorList collection` iterated always and collect all `conflicting` that has `foundIssue() == true`? Is my question a little bit clear?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-backend/src/test/java/org/drools/workbench/services/verifier/webworker/client/fromfile/DecisionTableAnalyzerFromFileTest.java
```diff
@@ -142,6 +142,24 @@ public void testFile3() throws
...
+
+        final GuidedDecisionTable52 table52 = GuidedDTXMLPersistence.getInstance()
+                .unmarshal(xml);
+
+        final Analyzer analyzer = analyzerProvider.makeAnalyser(table52);
+
+        analyzer.resetChecks();
+        analyzer.analyze();
+
+        analyzerProvider.getAnalysisReport().stream().forEach(i->System.out.println(i.getCheckType()));
```
jomarko: I think `System.out.println(i.getCheckType()` is really useful during fixing/investigating some bug, however then the test is just run automatically by a machine, without checking the output. Is there any chance for some debug level log or some assertion instead of the `System.out.println`?

drools-wb-services/drools-wb-verifier/drools-wb-verifier-client/src/main/java/org/drools/workbench/services/verifier/plugin/client/builders/ModelMetaDataEnhancer.java
```diff
@@ -17,37 +17,119 @@
...
             }
 
             columnIndex++;
         }
 
-        return new HeaderMetaData( map );
+        return new HeaderMetaData(map);
+    }
+
+    private String getFactType(final ActionSetFieldCol52 asf) {
```
jomarko: I think this should be placed in one of the following:
- `org.drools.workbench.services.verifier.plugin.client.builders.ColumnUtilitiesBase`
- `org.drools.workbench.services.verifier.plugin.client.builders.VerifierColumnUtilities`

drools-wb-services/drools-wb-verifier/drools-wb-verifier-client/src/main/java/org/drools/workbench/services/verifier/plugin/client/builders/PatternResolver.java
```diff
@@ -15,16 +15,13 @@
...
-import org.drools.workbench.models.guided.dtable.shared.model.ActionInsertFactCol52;
-import org.drools.workbench.models.guided.dtable.shared.model.ActionSetFieldCol52;
-import org.drools.workbench.models.guided.dtable.shared.model.Pattern52;
 import org.drools.workbench.services.verifier.api.client.configuration.AnalyzerConfiguration;
 import org.drools.workbench.services.verifier.api.client.index.Index;
 import org.drools.workbench.services.verifier.api.client.index.ObjectType;
 import org.drools.workbench.services.verifier.api.client.index.Pattern;
 import org.drools.workbench.services.verifier.api.client.index.Rule;
 import org.drools.workbench.services.verifier.plugin.client.api.HeaderMetaData;
+import org.drools.workbench.services.verifier.plugin.client.builders.ModelMetaDataEnhancer.PatternType;
```
jomarko: Unused import

drools-wb-services/drools-wb-verifier/drools-wb-verifier-backend/src/test/java/org/drools/workbench/services/verifier/webworker/client/fromfile/DecisionTableAnalyzerFromFileTest.java
```diff
@@ -142,6 +142,24 @@ public void testFile3() throws
     }
 
     @Test
+    public void testLHSConflictsArePickedUpForEachFieldOfAPattern() throws
```
jomarko: This test uses decision table with different characteristic as the one in the jira has. This table has two conditions like (WHEN OK THEN 10, WHEN AA THEN 20) while the table in jira has different kind of conditions (WHEN A THEN 1, WHEN B THEN 1, WHEN C THEN 1) shouldn't be such table part of tests?

manstis: @jomarko True (good spot)!

@Rikkola Probably does not hurt to include a test for the scenario in the JIRA.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/relations/ConflictResolver.java
Rikkola: The `conflicting` contains the two conflicting items, what ever they might be. To save time I just pick up the first issue.

Some additional information:  Then on reruns I check if this is still true. If it is then we skip the iteration all together and save some time. Conflicts are not always a bad thing. In this case we were checking for conflicts during subsumption checks. If there is a conflict the two items can not subsume each other. Subsumption is the heaviest operation, but it should also be the most rare situation in the table. Most tables should contain conflicts in the LHS, but this is different to conlicting rules. Where conflicting rules means two rules harm each other, conflicting LHS means the rules just fire with different Patterns.

You might have known all that. Just throwing as much info about this as possible :)

drools-wb-services/drools-wb-verifier/drools-wb-verifier-backend/src/test/java/org/drools/workbench/services/verifier/webworker/client/fromfile/DecisionTableAnalyzerFromFileTest.java
Rikkola: Yeah. Not a bad idea to make sure the table works. It is surprising how many bugs these dtable files are picking up and yet they are quite easy to add in.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-backend/src/test/java/org/drools/workbench/services/verifier/webworker/client/fromfile/DecisionTableAnalyzerFromFileTest.java
Rikkola: We can probably make a method that checks if the report is empty and then we only print the content if the test fails.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-client/src/main/java/org/drools/workbench/services/verifier/plugin/client/builders/ModelMetaDataEnhancer.java
Rikkola: True. We might need it elsewhere.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-backend/src/test/java/org/drools/workbench/services/verifier/webworker/client/fromfile/DecisionTableAnalyzerFromFileTest.java
jomarko: This sounds acceptable to me. thanks

drools-wb-services/drools-wb-verifier/drools-wb-verifier-api/src/main/java/org/drools/workbench/services/verifier/api/client/relations/ConflictResolver.java
jomarko: Thanks Toni, I see you understand the best your code, it is a little bit more clear to me after your explanation.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-client/src/main/java/org/drools/workbench/services/verifier/plugin/client/builders/ModelMetaDataEnhancer.java
Rikkola: And while actually doing the move, noticed this might still be the best place for now at least. VerifierColumnUtilities depends on ModelMetaDataEnhancer so moving this creates a cyclic dependency and it is probably not worth it to fix that before we actually need the getFactType method elsewhere.

jomarko: Hmm, not happy to hear, but well, I am fine if there will be no movement.

Rikkola:
Tests pass, but there is a bug during runtime that I tried to fix today. The fix was bad and had to revert the changes. I think I have the better fix in mind, but running out of time today so I'll fix it tomorrow.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-client/src/main/java/org/drools/workbench/services/verifier/plugin/client/builders/ModelMetaDataEnhancer.java
Rikkola: Spend some time on this area due to portable-bug I fixed in this PR and this needs some sort of clean up. The way the VerifierColumnUtilities and ColumnUtilities work are a bit messy. Tried few things, but the fix takes time. 

The fix might be to just do indexing in server side. We will see.

Rikkola:
Ok, so had to make `ModelMetaData` portable to make sure it works in the browser.

manstis:
@jomarko Could you please re-review?

Rikkola:
I believe this will go in as it is. Making the 7.4.x PR.

manstis:
Jenkins please retest this.

jomarko:

PR #649, opened 2017-10-11
GUVNOR-3431: [Guided Rule] Unable to set literal value of accumulated condition

karreiro:
See:
https://issues.jboss.org/browse/GUVNOR-3431

Before:
before

After:
after


jomarko:

drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/factPattern/PopupCreator.java
```diff
@@ -179,234 +179,259 @@ private boolean hasApplicableFields( final ModelField[] fields ) {
...
+    }
+
+    SingleFieldConstraintEBLeftSide makeSingleFieldConstraintEBLeftSide(final String factType) {
+
+        final SingleFieldConstraintEBLeftSide constraint = new SingleFieldConstraintEBLeftSide();
+        final ExpressionFormLine expression = makeExpressionFormLine(factType);
+
+        constraint.setConstraintValueType(SingleFieldConstraint.TYPE_UNDEFINED);
+        constraint.setExpressionLeftSide(expression);
+        constraint.setFactType(factType);
```
karreiro: @manstis the real change is here. The factType was not being set :-)

PR #651, opened 2017-10-12
GUVNOR-3556: [Guided Decision Table] Default Value hint stays shown after wizard is finished or cancelled

manstis:
See https://issues.jboss.org/browse/GUVNOR-3556

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/common/DecisionTableColumnViewUtils.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/common/DecisionTableColumnViewUtils.java
@@ -16,6 +16,8 @@
 
 package org.drools.workbench.screens.guided.dtable.client.wizard.column.pages.common;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -36,6 +38,9 @@ import org.jboss.errai.common.client.dom.Div;
 import org.jboss.errai.common.client.dom.HTMLElement;
 
 public class DecisionTableColumnViewUtils {
+
+    private static final List<Element> popoverElementRegistrations = new ArrayList<>();
+    private static final List<HTMLElement> popoverHTMLElementRegistrations = new ArrayList<>();
 
     public static int getCurrentIndexFromList(final String currentValue,
                                               final ListBox list) {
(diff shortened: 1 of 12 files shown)


jomarko:

manstis:
Jenkins execute full downstream build.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/common/DecisionTableColumnViewUtils.java
```diff
@@ -37,6 +39,9 @@
 
 public class DecisionTableColumnViewUtils {
 
+    private static final List<Element> popoverElementRegistrations = new ArrayList<>();
+    private static final List<HTMLElement> popoverHTMLElementRegistrations = new ArrayList<>();
+
```
karreiro: Why do we need to add these two lists? 🤔

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/common/DecisionTablePopoverUtils.java
```diff
@@ -0,0 +1,122 @@
...
+
+@ApplicationScoped
+public class DecisionTablePopoverUtils {
+
+    private final List<Element> popoverElementRegistrations = new ArrayList<>();
+    private final List<HTMLElement> popoverHTMLElementRegistrations = new ArrayList<>();
+
+    public void setupPopover(final Element e,
+                             final String content,
+                             final boolean register) {
```
karreiro: Personally, I would add two methods instead of using the `register` parameter. Something like:
```java
setupPopover(final Element e, final String content)
// and
setupAndRegisterPopover(final Element e, final String content)
```
However, feel free to keep it as is ;-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/common/DecisionTableColumnViewUtils.java
manstis: They don't extend a common ancestor (other than ```Object```) :-(

Some Popovers are on older ```Widget.getElement()``` and some on Errai's ```HTMLElement```

karreiro: @manstis Hmm... but.. I think that these two lists are never used. Maybe I'm missing something.

manstis: @karreiro Ahhhhhhhhhhhhh.. ok. I'll delete them.. (the ```Popover``` helper used to be in this class).

karreiro: :-)

manstis:
@karreiro Updates completed. Thanks for your review :-)

manstis:
Jenkins please retest this.

manstis:
Jenkins please retest this.

PR #652, opened 2017-10-12
GUVNOR-3557: [Guided Decision Table] User is able to bind two fields to same name

manstis:
See https://issues.jboss.org/browse/GUVNOR-3557

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/ValueOptionsPageView.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/ValueOptionsPageView.java
@@ -69,6 +69,9 @@ public class ValueOptionsPageView implements IsElement,
     @DataField("bindingContainer")
     private Div bindingContainer;
 
+    @DataField("fieldBindingWarning")
+    private Div fieldBindingWarning;
+
     private ValueOptionsPage<?> page;
 
     private TranslationService translationService;
@@ -86,6 +89,7 @@ public class ValueOptionsPageView implements IsElement,
                                 final Div defaultValueContainer,
                                 final Div limitedValueContainer,
                                 final Div bindingContainer,
+                                final Div fieldBindingWarning,
                                 final TranslationService translationService,
                                 final DecisionTablePopoverUtils popoverUtils) {
         this.valueListGroupContainer = valueListGroupContainer;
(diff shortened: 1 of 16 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/ValueOptionsPageView.java
```diff
@@ -95,6 +99,7 @@ public ValueOptionsPageView(final Div valueListGroupContainer,
         this.defaultValueContainer = defaultValueContainer;
         this.limitedValueContainer = limitedValueContainer;
         this.bindingContainer = bindingContainer;
+        this.fieldBindingWarning=fieldBindingWarning;
```
karreiro: Here ;-)

manstis:
@karreiro Updated. Thanks for your review :-)

manstis:
Jenkins please retest this.

manstis:
Jenkins please retest this.

jomarko:

jomarko:
screenshot from 2017-10-16 13-54-34

manstis:
@jomarko Fixed. Sorry for wasting your time.

karreiro:
Jenkins please retest this.

karreiro:
Jenkins please retest this.

karreiro:
Jenkins please retest this.

jomarko:
Also this, building locally, as jenkins ...

manstis:
@jomarko As commented on the JIRA the reported issue is not related to this PR; however appears to be long-standing. I have however included a fix in this PR to remove bound ```FactPattern``` in "BRL Conditions" from the pattern selector drop-down when adding new "simple conditions". The enhancement to allow the addition of "simple conditions" to "BRL Conditions" is quite a bit more extensive and should probably be a new JIRA requesting such an enhancement, if desirable.

jomarko:
@manstis sorry, I am reading this later, as jira, well I am fine if no new jira will be fired, if you mention the behavior here https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/resources/org/drools/workbench/screens/guided/dtable/client/resources/i18n/GuidedDecisionTableErraiConstants.properties#L123

manstis:
@jomarko I've updated PR to show different descriptions for "Pattern Page" depending on whether you're adding a condition (in which case only patterns on existing _Condition_ columns can be used); or whether you're adding an action (in which case the available patterns include those in both _Condition_ columns or _Condition BRL fragments_).

PR #659, opened 2017-10-17
AF-593: Decouple DMO from Drools. Add missing dependency

manstis:


karreiro:
Oh! I was about to open the same PR... You did it faster @manstis ;-)

manstis:
Thanks @karreiro

PR #661, opened 2017-10-18
Use XStreamUtils.createTrustingXStream(..) helper

manstis:
See BSIG emails regarding XStream.


jomarko:
Still good in my eyes, my approval stays in place.

PR #662, opened 2017-10-18
GUVNOR-3432: [Guided Decision Table] Formulas and Predicates cannot be set

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3432

Before:
before

After:
after

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/FieldPageTest.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/FieldPageTest.java
@@ -49,9 +49,9 @@ import org.uberfire.client.callbacks.Callback;
 import org.uberfire.ext.widgets.core.client.wizards.WizardPageStatusChangeEvent;
 import org.uberfire.mocks.EventSourceMock;
 
+import static org.drools.workbench.screens.guided.rule.client.util.ModelFieldUtil.modelField;
 import static org.junit.Assert.*;
 import static org.mockito.Mockito.*;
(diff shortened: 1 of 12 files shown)


jomarko:

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/FieldPageTest.java
```diff
@@ -374,22 +374,75 @@ public void testAsWidget() {
...
+
+        doReturn(factType).when(page).factType();
+
+        page.setupPatternWarningMessages();
+
+        verify(view).patternWarningToggle(false);
+    }
+
+    @Test
+    public void testSetupPatternWarningMessagesWhenFactTypeIsNilAndConstraintValueIsPredicate() {
```
jomarko: @karreiro This test is great, it brought me to this observation:
- If Calculation Type is Predicate, Field Page doesn't force the user to specify pattern
- However, the column wizard always force the user to specify pattern on the pattern page
- Second however, I am not sure, how to solve this

manstis: A Pattern always needs to be selected; therefore IMO even if the Calculation Type is Predicate the warning about Pattern not being selected should be shown on the Field page. When using a Predicate it is not possible to complete the wizard unless a Pattern has been selected.

jomarko: Thanks Michael, +1 from my site. Guilherme, please incorporate also this.

karreiro:
Changes applied. Thank you for the review @manstis and @jomarko.

---

demo

karreiro:
Jenkins please execute full downstream build.

PR #670, opened 2017-10-25
GUVNOR-3578: Disable or Enable both column manipulation buttons

jomarko:
Opening a table was setting column manipulation buttons state just for the Add Column button. As there is probably not case, when Add Collumn button could be enabled and Edit Column button disabled (and vice versa), the Table Modeller view API was simplified by removing method that is setting state just for Add Column button.


jomarko:

jomarko:
@karreiro @manstis I would be pleased with your review.

PR #676, opened 2017-10-27
DROOLS-2074: [Guided Decision Table] BRL Condition Update is Vetoed

manstis:
See https://issues.jboss.org/browse/DROOLS-2074

@jomarko The issue was that an "update" to a BRL Condition is effectively an "add" (of the updated) definition and then a "delete" (of the old). The veto exception was caused because the "delete" checks if any bindings in the fragment being deleted are used in the RHS (or other LHS elements). I've simply skipped the "delete" check when invoked as part of the update. 

This is safe as the "update" first checks if any deleted bindings (difference between original bindings and updated bindings) are used in the RHS (or other LHS elements) and fails fast if any are found.


jomarko:

jomarko:

PR #677, opened 2017-10-30
DROOLS-2078: Prepare drl-text-editor-api for static code analysis

jomarko:


jomarko:
@manstis I think it is ready to merge.

PR #678, opened 2017-10-30
DROOLS-2079: Prepare drl-text-editor backend and client modules for s…

jomarko:
…tatic code analysis


jomarko:
@manstis I have just fix the mistake in PR summary and commit message.

PR #680, opened 2017-10-31
Drools 2082 static code analysis

jomarko:
Prepare `drools-wb-verifier` module for the static code analysis that is already turned on in some other projects of kiegroup.

@manstis please have a look. 99% of changes is maked up from licences and imports organization. I think just in one file I have added `{}` for if statements.

### DIFF
--- a/drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/JavaCheckRunner.java
+++ b/drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/JavaCheckRunner.java
@@ -1,9 +1,24 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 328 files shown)


drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/JavaCheckRunner.java
```diff
@@ -1,9 +1,24 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
```
manstis: 2017 (if I was being nit-picking)

drools-wb-services/drools-wb-verifier/drools-wb-verifier-core/src/main/java/org/drools/workbench/services/verifier/core/checks/base/PriorityListCheck.java
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
```
manstis: 2017 (if I was being nit-picking)

jomarko:
looking on git history seems to me as 2016

manstis:
@jomarko Quite possibly, I did not check. Given (1) there are lots of classes in the module with a date of 2016, (2) it's a petty, trivial observation; I merged "as is"... but I don't doubt the creation date was 2016!

PR #685, opened 2017-11-01
DROOLS-2089: Prepare drools-wb-globals-editor module for static code …

jomarko:
…analysis


jomarko:
jenkins please retest this

PR #699, opened 2017-11-02
DROOLS-2107: Run Scenario against rules that use inheritance

jomarko:


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/resources/org/drools/workbench/screens/testscenario/backend/server/HelloEveryOne.gdst
```diff
@@ -83,7 +83,11 @@
     <entries/>
   </auditLog>
   <imports>
-    <imports/>
+    <imports>
+      <org.kie.soup.project.datamodel.imports.Import>
```
manstis: I know it's unrelated to this PR; but do you have an older (pre- ```kie-soup```) GDST with an import?

I thought we "aliased" the FQCN to the simple-name however if we do not; and older GDST contain the old FQCN we may have a problem loading such files in newer versions of the workbench. I had a look here but can only find "old" GDST without any imports.

jomarko:
@manstis I am going to add similar tests with different characteristics, see https://issues.jboss.org/browse/BXMSQE-258. Feel free to create task/s there for me.

manstis:
@jomarko I assume this can be merged now?

jomarko:
@manstis yes, I think so.

PR #705, opened 2017-11-07
DROOLS-2105: Add test for asserting rules which fire work item handler

jomarko:


jomarko:
@manstis hehe, sure, I am using the workbench, about my experience, sure, we can discuss it, I will ping you on irc when I have a moment

jomarko:
@manstis I updated also this, maybe add @Rikkola among reviewers also here.

jomarko:
@Rikkola the PR is rebased with latest changes form master, please review when you have a moment.

jomarko:
@manstis I think right time to merge.

PR #706, opened 2017-11-08
DROOLS-2102: Use Child objects in scenario to fire rule

jomarko:
@Rikkola @manstis Please have a look.


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/model/Acid.java
```diff
@@ -0,0 +1,54 @@
...
+
+package org.drools.workbench.screens.testscenario.backend.server.model;
+
+
+public class Acid implements java.io.Serializable {
+
+	static final long serialVersionUID = 1L;
+
+	private java.lang.String name;
+	private java.util.List<org.drools.workbench.screens.testscenario.backend.server.model.Molecule> molecules;
```
manstis: An ```import``` would have been nice.. but not essential.

manstis: (for both ```List``` and ```Molecule``` and later ```String```)

jomarko:
@manstis commit added, shout if anything else I could improve.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/model/Acid.java
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.drools.workbench.screens.testscenario.backend.server.model;
+
+import java.io.Serializable;
+import java.util.List;
+
+public class Acid implements Serializable {
```
manstis: Would you mind revisiting the formatting?

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/model/Acid.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public List<Molecule> getMolecules() {
+        return this.molecules;
+    }
+
+    public void setMolecules(
```
manstis: Formatting here...

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/model/Acid.java
```diff
@@ -0,0 +1,55 @@
...
+            List<Molecule> molecules) {
+        this.molecules = molecules;
+    }
+
+    public Acid(
```
manstis: Formatting here...

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/test/java/org/drools/workbench/screens/testscenario/backend/server/model/Acid.java
jomarko: Should be done now, I copied the java model classes from workbench data editor. It is where comes these weird things from. Also weird after that I applied code style in IDE and not all of them were corrected.

PR #707, opened 2017-11-14
DROOLS-2076: Add reproducer test for the guided rule editor

jomarko:
This PR adds reproducer test that assert DSL sentences are loaded
correctly into given popups:
- Colum Wizard in Guided decision table editor
- Guided Rule editor
- Guided Rule Template editor

@manstis I would like to ask you for review.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-backend/src/test/java/org/drools/workbench/screens/guided/rule/backend/server/GuidedRuleEditorServiceImplTest.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-backend/src/test/java/org/drools/workbench/screens/guided/rule/backend/server/GuidedRuleEditorServiceImplTest.java
@@ -16,17 +16,27 @@
 
 package org.drools.workbench.screens.guided.rule.backend.server;
 
-import java.util.HashMap;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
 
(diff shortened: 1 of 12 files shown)


drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-backend/src/test/java/org/drools/workbench/screens/guided/rule/backend/server/GuidedRuleEditorServiceImplTest.java
```diff
@@ -63,27 +73,30 @@
...
+    @Mock
+    private DSLSentence dslSentence;
+
     @InjectMocks
     GuidedRuleEditorServiceImpl service = new GuidedRuleEditorServiceImpl(sessionInfo,
                                                                           mock(Instance.class));
 
     @Test
-    public void checkConstructContentPopulateProjectCollectionTypes() {
+    public void checkConstructContentPopulateProjectCollectionTypesAndDSLSentences() throws Exception {
```
jomarko: @triceo I was trying to cover this service with integration test, that would not mock `PackageDataModelOracle` data, however I gave up after like 8 hours of constructing setup for the `GuidedRuleEditorServiceImpl` class. There was like 20 complex constructors called before the test and setup was still failing. Moving to another issue rather than spend more hours here and not be sure if I will succeed.

jomarko:
jenkins please retest this

jomarko:
@manstis feel free to re review. I had to add some dependency exclusions to get green result.

PR #712, opened 2017-11-20
DROOLS-2076: Keep listbox multi select to show correctly

jomarko:
Before fix:
screenshot from 2017-11-20 14-57-17
After fix:
screenshot from 2017-11-20 14-55-54


jomarko:
jenkins please retest this

PR #714, opened 2017-11-23
[DO NOT MERGE] DROOLS-1798: Show RuleModel imports in FromAccumulate FactType selector

jomarko:
Ensemble with:
- kiegroup/kie-wb-common#1306
- kiegroup/kie-soup#17

PR #720, opened 2017-11-29
DROOLS-1798: Show RuleModel imports in FromAccumulate FactType selector

jomarko:
Ensemble with:
- kiegroup/kie-wb-common#1306
- kiegroup/kie-soup#17


jomarko:
@triceo @manstis please re-review. This is replacement of #714

PR #723, opened 2017-12-04
DROOLS-2160: Convert delete buttons to gwt-bootstrap3

jomarko:


jomarko:

jomarko:

manstis:
@Rikkola Fancy a look?

jomarko:
@manstis thanks for review, sure, @Rikkola your review is welcome, here is video showing result

https://webmshare.com/KwPmB

jomarko:
@karreiro well, I am open to this approach, could you please point to to proper button element I should use?

manstis:
@karreiro @jomarko Well... my 2c is this: This editor is erm, how can I put it, not one of the better ones. It is set to be replaced (one day.. development even started recently.. and has not stopped and been cancelled). Anything quick and easy we can do to make it "fit" better in the WB then great however IMO anything more than a trivial tickle isn't probably worth the effort. Same for the Guided Rule Editor and Guided Rule Template Editor (and hence BRL Fragment editor).

karreiro:
@manstis Yeah, I agree. It's important to be pragmatic, and some places don't deserve to be refactored for now.

@jomarko Answering the question that you made earlier on IRC -  "Why it is better to use components from Elemental2 rather than from GWT Bootstrap 3?"

The GWT Bootstrap 3 is a wrapper for GWT Widgets with the Bootstrap style applied. It's pretty handy, but we probably won't have GWT Widgets in a near future.

Knowing that, it's a good idea to not couple the new features (or refactorings) that we're writing with the old GWT Widgets. It's a big challenge because we have a lot of features depending on GWT Widgets, but with Errai + Elemental 2, this is tangible.

Regarding the use of Elemental 2 API instead of the deprecated [Errai elements API](https://github.com/errai/errai/tree/master/errai-common/src/main/java/org/jboss/errai/common/client/dom), see:

> Elemental 2 is a project that exposes built-in web APIs in Java. Like the first Elemental, the project is automatically generated from external definitions of web APIs. Unlike its predecessor, Elemental 2 uses JS Interop.
>
> Back when we began work on Errai 4, Elemental 2 was not yet available. We didn't want that to slow down our work on HTMLElement-based user interfaces and JS Interop support, and so we created a collection of hand-written element wrappers in Errai.
>
> But now that Elemental 2 is available, it is in the best interests of the GWT community to standardise around these. As such, Errai's DOM wrappers have been deprecated and we have added full support to templating and data-binding for classes from Elemental 2.
>
> We will not be removing the Errai element wrappers, but we will also not be updating them, and you are highly encouraged to use Elemental 2 going forward.
>
> Source: http://errai-blog.blogspot.com.br/2017/11/errai-410final-released.html

Besides, it's much more readable writing templates using pure HTML than an `ui.xml` or even by instantiating elements inside  Java classes :-)

jomarko:
@manstis @karreiro  thanks for comments, especially for the blog where I found *JsInterop v1.0:
Nextgen GWT/JavaScript Interoperability* document.

manstis:
@jomarko Hi.. was you wanting a green full downstream build before merge?

jomarko:
@manstis to be 100 percent sure, I will try to get green.

jomarko:
@manstis green

PR #724, opened 2017-12-05
DROOLS-1811: [Guided Decision Table] Column Wizard usability improvements

karreiro:
See: https://issues.jboss.org/browse/DROOLS-1811

---

drools-1811


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/wizard/column/pages/OperatorPage.java
```diff
@@ -76,13 +78,29 @@ public void prepareView() {
...
+        setupCurrentField();
     }
 
-    private void setupOperator() {
+    void setupCurrentField() {
+        view.setCurrentField(currentField());
+        view.currentFieldToggle(hasFactField());
+    }
+
+    String currentField() {
```
manstis: A great example of an alternative way to achieve "a : b" construction (with null/empty checks) I've seen :-)

karreiro:
@jomarko Thank you.

I just applied the fix :-)

PR #725, opened 2017-12-07
Cover validation of dslr file

jomarko:
@manstis @jhrcek I would like to ask you for a review.


jomarko:
@manstis good point, I thought I removed all not necessary, thanks for spotting. Will amend in a minute.

jomarko:
@manstis please recheck.

drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/pom.xml
```diff
@@ -162,6 +162,30 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
jhrcek: I understand that it might seem confusing but you probably don't need both `uberfire-testing-utils` and `uberfire-test-utils`

[ubefire-test-utils](https://github.com/kiegroup/appformer/tree/master/uberfire-test-utils) is a module that was recently moved from guvnor's guvnor-test-utils and renamed. Looking at the sources it mainly contains utilities to make setup of CDI (weld-se -based) tests easier. In the long run I think it would be best to refactor that module a bit by:
1. Renaming it to *ubefire-backend-test-utils* to prevent confusion
2. Simplifying the weld setup a bit (along the lines of [this PR](https://github.com/kiegroup/kie-wb-common/pull/1307))


[uberfire-testing-utils](https://github.com/kiegroup/appformer/tree/master/uberfire-testing-utils) is a module that's been in uberfire for a long time and contains utility classes primarily for client-side testing with mockito (e.g. CallerMock etc). I think you don't need to add that as a test dependency in this case. In the long run I think this should be renamed to *uberfire-client-test-utils*

@ederign would you agree with renaming 
uberfire-test-utils -> uberfire-baskend-test-utils
uberfire-testing-utils -> uberfire-client-test-utils

drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/src/test/java/org/drools/workbench/screens/drltext/backend/server/DRLTextEditorServiceImplCDITest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        // disable git and ssh daemons as they are not needed for the tests
+        System.setProperty("org.uberfire.nio.git.daemon.enabled", "false");
+        System.setProperty("org.uberfire.nio.git.ssh.enabled", "false");
+        System.setProperty("org.uberfire.sys.repo.monitor.disabled", "true");
+
+        //Bootstrap WELD container
+        weldContainer = new Weld().initialize();
```
jhrcek: For now it's ok like this, but in the future we might want to reuse [this](https://github.com/kiegroup/appformer/blob/master/uberfire-test-utils/src/main/java/org/guvnor/test/CDITestSetup.java) to share this repeated piece of setup (after we simplify it a bit, because [this piece](https://github.com/kiegroup/appformer/blob/master/uberfire-test-utils/src/main/java/org/guvnor/test/CDITestSetup.java#L54-L60) can be simplified to `weldContainer.select(SERVICE_YOU_WANT_TO_INSTANTIATE).get();`).

drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/pom.xml
jomarko: Great, understood, will preserve just the correct one.

drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/src/test/java/org/drools/workbench/screens/drltext/backend/server/DRLTextEditorServiceImplCDITest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.uberfire.backend.server.util.Paths;
+import org.uberfire.backend.vfs.Path;
+import org.uberfire.java.nio.fs.file.SimpleFileSystemProvider;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(MockitoJUnitRunner.class)
```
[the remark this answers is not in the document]
jomarko: Good catch, will remove the annotation `@RunWith`

drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/src/test/java/org/drools/workbench/screens/drltext/backend/server/DRLTextEditorServiceImplCDITest.java
jomarko: Actually, I remember I sent you link to this, I do not know why I have not used :)

jomarko:
@jhrcek thanks for valuable feedback. changes applied.

PR #730, opened 2017-12-13
DROOLS-2149: Row and column can be added to read only table

manstis:
See https://issues.jboss.org/browse/DROOLS-2149

The problem was that when the editor was opened the menus' state is set based on the "selected decision table". At the point of opening there has been no request for the "lock state" and hence the file appears unlocked and the menus enabled. However when the async call to retrieve the "lock state" completes the file becomes locked and the menus should be disabled.

This PR introduces a new ```RefreshMenusEvent``` that is fired when the "lock state" has been retrieved causing the menus to refresh themselves. Tried and tested locally (what is nice is that when a file is locked, as per the JIRA, and then becomes unlocked the menus in the other users browser update to become enabled).

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
@@ -68,6 +68,7 @@ import org.drools.workbench.screens.guided.dtable.client.widget.table.events.cdi
 import org.drools.workbench.screens.guided.dtable.client.widget.table.events.cdi.RefreshActionsPanelEvent;
 import org.drools.workbench.screens.guided.dtable.client.widget.table.events.cdi.RefreshAttributesPanelEvent;
 import org.drools.workbench.screens.guided.dtable.client.widget.table.events.cdi.RefreshConditionsPanelEvent;
+import org.drools.workbench.screens.guided.dtable.client.widget.table.events.cdi.RefreshMenusEvent;
 import org.drools.workbench.screens.guided.dtable.client.widget.table.events.cdi.RefreshMetaDataPanelEvent;
 import org.drools.workbench.screens.guided.dtable.client.widget.table.lockmanager.GuidedDecisionTableLockManager;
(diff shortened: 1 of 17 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTablePresenter.java
```diff
@@ -625,6 +629,7 @@ void onUpdatedLockStatusEvent(final @Observes UpdatedLockStatusEvent event) {
                 access.setLock(NOBODY);
             }
             refreshColumnsPage();
+            refreshMenus();
```
manstis: This is the crux of the fix.

manstis:
@jomarko Thanks for the review.

> Table is not marked as locked/read only if the user defines new column / edit an existing one.

Not at all and is a bug (possibly introduced by the change to how columns are maintained).

> Assume there is a graph with a tableA and a tableB....

The graph and individual tables are locked IIRC.. so UserA can modify TableA and UserB, TableB in the same graph. However the graph is locked too and IIRC UserA will be prevented from doing somethings to the graph "as a whole" if UserB has it locked... I'd need to run the WAR locally to verify.

jomarko:
@manstis thanks.
- I filed the first issue [here](https://issues.jboss.org/browse/RHDM-218)
- About the second, one let me know your findings

manstis:
@jomarko This is my findings:

- UserA and UserB open Graph containing TableA and TableB

- No locks acquired.

- UserA selects/clicks TableA -> TableA is locked by UserA
- UserB selects/clicks TableA -> TableA is read-only cannot be modified. 
  - "Columns" tab is disabled.
  - "Save", "Delete", "Rename" are disabled.
  - UserB can make changes on Overview/Data Objects but cannot save it.

- UserB can open TableB -> TableB is locked by UserB.

- UserA closed Graph.

- UserB's UI updates as lock is released:
  - "Columns" tab is enabled.
  - "Save", "Delete", "Rename" are enabled.
  - UserB can save changes.

That said I did find behaviour a bit, erm at best, "flakey". Sometimes when UserB selected/clicked TableA a Notification appeared stating TableA was locked by UserA. In this case the "Columns" tab did not disable. When the Notification did not appear I found the "Columns" tab was disabled. So I suspect there are to observes of either _Lock_ or _Decision Table Selected_ events and different things are happening for different observes.. or perhaps even some explicit code that is being called and "cancelled" by an event... but I think it is worthy of some investigation.

PR #737, opened 2017-12-18
6.5.x DROOLS-2189 : Business central shows redundant warning messages from validation.

Rikkola:
Updated the existing unit tests for this.

https://issues.jboss.org/browse/DROOLS-2189

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/DecisionTableAnalyzer.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/DecisionTableAnalyzer.java
@@ -59,38 +59,38 @@ public class DecisionTableAnalyzer
     private final GuidedDecisionTable52 model;
     private final EventManager eventManager = new EventManager();
 
-    public DecisionTableAnalyzer( final PlaceRequest place,
-                                  final AsyncPackageDataModelOracle oracle,
-                                  final GuidedDecisionTable52 model,
-                                  final EventBus eventBus ) {
+    public DecisionTableAnalyzer(final PlaceRequest place,
+                                 final AsyncPackageDataModelOracle oracle,
+                                 final GuidedDecisionTable52 model,
+                                 final EventBus eventBus) {
         this.place = place;
         this.model = model;
 
-        cache = new RowInspectorCache( oracle,
-                                       model,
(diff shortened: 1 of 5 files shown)


Rikkola:
@tomasdavidorg @karreiro I would like to request reviews.

@mareknovotny I could use a merge once the reviews are complete.

Rikkola:
@karreiro Nice, you broke the bug!

I'll double check this.

Rikkola:
Good job @karreiro. So your comment made me test this. If I edit the table fast enough I get the duplicate messages. I even managed to get both conflict and redundancy for the same 2 rows.
Redundancy and conflict are the exact opposites and can never be on two lines. So what is happening here is the list is showing results from more than one verification runs. Need to investigate more.

Rikkola:
@tomasdavidorg @karreiro New try! This one gave me some gray hairs. First finding reproducer was tricky, then the fix was even harder. But works now.

karreiro:
Thank you for the update @Rikkola :-)

Unfortunately, I'm getting the same results :-( see:

Before (commit: 23e6e5b4f38406f847759aa35938b7fd6b8274fb):
before

After (commit: 7d4df927e1546860b87dc974f7e9b1834521114f):
after

Am I missing something?

Thank you.

Rikkola:
@karreiro My bad. I forgot to copy the reproducer here. Took me while to figure it out.

How to reproduce.

Make two rows:
LHS | RHS
age == 10 | approved = true
age == 10 | approved = true

change to:

age == 10 | approved = true
age == 10 | approved = false

change to:

age == 10 | approved = false
age == 10 | approved = false

Both conflict and redundancy are present. Same can be done to get duplicate Conflict report.

karreiro:
@Rikkola Thank you for the steps to reproduce :-)

Unfortunately, I think that I found a problem :-( see:

problem

Rikkola:
@karreiro Yep. That indeed is a problem. Could be introduced by these changes.

Rikkola:
Ok so the original issue is fixed. What @karreiro found is an issue, but I can only reproduce with bool and int. There might be more, but at least with Strings it is not an issue. Can't reproduce in unit tests either, tried the exactly same setup karreiro had.

manstis:
@mareknovotny Why did you merge this when only 1 of three people have approved it?

Rikkola:
@tomasdavidorg started the review while covering @jomarko when he was on PTO.
@karreiro found a bug, but we are making a new ticket to cover it. The original issue is fixed and the new bug is less likely to happen. It probably already exists even without these changes. Still need to check.

And we are running out of time. Deadline is today.

mareknovotny:
@manstis as @Rikkola mentioned above, we talked about if that is ok to merge on IRC, the original issue was fixed with a side bug for int and boolean row values, this will be fixed in new created issue.

PR #738, opened 2018-01-04
Increase the TableWizard coverage

jomarko:
This PR focus on a case when the TableWizard informs the user about no available pattenrs which could be used in the table actions. In other words, user can not change a field of a pattern because there is no pattern defined yet.

This PR is part of migration form internal selenium test suite into community. See BAQE-314.

@manstis please have a look.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/table/pages/ActionSetFieldsPageTest.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/table/pages/ActionSetFieldsPageTest.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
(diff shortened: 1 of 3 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/wizard/table/pages/ActionSetFieldsPageTest.java
```diff
@@ -0,0 +1,84 @@
...
+        model.setConditionPatterns(Collections.singletonList(
+                new Pattern52() {{
+                    setChildColumns(Collections.singletonList(new ConditionCol52()));
+                }})
+        );
+
+        actionSetFieldsPage.prepareView();
+        verify(actionSetFieldsPageView).setAvailablePatterns(patternsCaptor.capture());
+
+        assertEquals(1, patternsCaptor.getValue().size());
```
manstis: It might be worth checking the pattern set is that you set in the model?

jomarko:
@manstis thanks for feedback, incorporated now.

manstis:
Please let me know when green.

jomarko:
jenkins please retest this

jomarko:
@manstis green

PR #739, opened 2018-01-04
[7.5.x] Increase the TableWizard coverage

jomarko:
This PR focus on a case when the TableWizard informs the user about no available pattenrs which could be used in the table actions. In other words, user can not change a field of a pattern because there is no pattern defined yet.

This PR is part of migration form internal selenium test suite into community. See BAQE-314.

@manstis please have a look. This is corollary of #738 .


jomarko:
@manstis feedback incorporated also here.

manstis:
Please let me know when green.

jomarko:
jenkins please retest this

jomarko:
@manstis green

PR #746, opened 2018-01-10
DROOLS-2228: [Guided Decision Table] Rename Set the field value with Work Item parameter column

manstis:
See https://issues.jboss.org/browse/DROOLS-2228

i18n will not take effect until the next round of translations from Zanata.


manstis:
Jenkins please retest this.

manstis:
@jomarko Who knows!?!? Engineering _push_ the _source_ files (```EN_us```) to Zanata however the translation process is, AFAIK, no longer free and needs requesting rather than what used to be a continual process. I'd like to think translations are requested and paid for whenever a product release is made.. but to be honest IDK. @etirelli Any ideas?!

jomarko:
@manstis could we cherry-pick this into 7.5.x please?

PR #749, opened 2018-01-11
BAQE-317: Cover more properly XLS decision table validation

jomarko:


jomarko:
@manstis tests passed, I think we can merge.

jomarko:
@manstis are we fine if I cherry pick these tests into 7.5.x?

manstis:
@jomarko I am fine with that; it is only additional tests.

PR #751, opened 2018-01-12
BAQE-322 Cover validation of the guided rule with inheritance

jomarko:
This PR cover validation of the guided rule asset while focus on the guided rule that inherit from another guided rule.

PR #752, opened 2018-01-15
[7.5.x] BAQE-322 Cover validation of the guided rule with inheritance

jomarko:
This PR cover valiudation of the guided rule asset while focus on the guided rule that inherit from another guided rule.

This is corollary of #751 .


manstis:
@jomarko I've approved (as it's a cherry-pick to 7.5.x) however I did notice (and it's probably the same in the orginal PR) that the paths for "Guided Rules" reference "dtables". I wonder whether you want to improve the separation and move RDRL files to different packages/projects?

jomarko:
@manstis makes sense, will update also master

jomarko:
jenkins please retest this

jomarko:
@manstis ready for merge when you have free hand.

PR #758, opened 2018-01-18
BAQE-322: Rename tested project from dtables to rules

jomarko:
Corollary of #752 review comments.


jomarko:
@manstis ready for merge when you have free hand.

PR #767, opened 2018-01-23
[6.5.x] DROOLS-2263 : Unexpected results in GDST when using enumerations with commas

Rikkola:

This PR has a friend:
https://github.com/kiegroup/drools/pull/1716

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/utils/GuidedDecisionTableUtils.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/utils/GuidedDecisionTableUtils.java
@@ -22,6 +22,7 @@ import org.drools.workbench.models.datamodel.rule.DSLSentence;
 import org.drools.workbench.models.datamodel.rule.FactPattern;
 import org.drools.workbench.models.datamodel.rule.IAction;
 import org.drools.workbench.models.datamodel.rule.IPattern;
+import org.drools.workbench.models.datamodel.util.ListSplitter;
 import org.drools.workbench.models.guided.dtable.shared.model.ActionCol52;
 import org.drools.workbench.models.guided.dtable.shared.model.ActionInsertFactCol52;
 import org.drools.workbench.models.guided.dtable.shared.model.ActionSetFieldCol52;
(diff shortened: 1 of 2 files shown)


Rikkola:
And it needs acks.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/utils/GuidedDecisionTableUtils.java
```diff
@@ -309,25 +310,30 @@ private String getTypeFromDataOracle( final String factType,
...
     private String[] getValueList( final ActionInsertFactCol52 col ) {
         if ( col.getValueList() != null && !"".equals( col.getValueList() ) ) {
-            return col.getValueList().split( "," );
+            return getSplit(col.getValueList());
         }
         return new String[ 0 ];
     }
 
+    private String[] getSplit(final String valueList) {
+        return ListSplitter.split(valueList);
```
jomarko: Due to changes in `RuleModelDRLPersistenceImpl`, shouldn't we call `ListSplitter.split("\"",workingValue);` here?

Rikkola: I used the single ' quote for the outer quotes. This way " can be used inside. Here any " gets ignored and included.

Now I see a test in master says` "a, b b b  ,c"` should become `"a", " b b b  ", "c"`. Not sure if the ' is needed then.

Rikkola: Thinking thinking. Need to figure out how master and this branch are compatible.

jomarko: I mean, I understand that implementation in this PR, assumes *'* as spanner of options like *'John, Peter'*, and it works in UI as implemented, but I am not sure if it then affects somehow backend where *"* is used as a spanner of such options. I mean Imagine in UI *'"A", "B"'* will this backend store correctly?

Rikkola:
Ok, the acks are in.

manstis:
@Rikkola Need @jomarko to approve... is this for EoD today?

PR #770, opened 2018-01-25
DROOLS-2160: Add spacing between table cells

jomarko:
This PR is corollary of Getting started experience requirement, where @ibek asked for more spaces between elements in scenario designer.

screenshot from 2018-01-25 09-13-16

PR #771, opened 2018-01-25
[7.5.x] DROOLS-2160: Add spacing between table cells

jomarko:
This PR is corollary of Getting started experience requirement, where @ibek asked for more spaces between elements in scenario designer.

**Cherry-pick** of #770


manstis:
@jomarko advised this is a cherry-pick...

@jomarko Do you think, in the future, you could state in the PR whether it's a cherry-pick?

jomarko:
@manstis definitely I can, in this one I put short sentence "This is corollary of ...", but I admit, It can be easily overseen, please what do you think should be ideal to make clear and easily distinguishable which PRs are just cherry-picks?

manstis:
@jomarko I saw your original comment about "corollary" however to me that simply means this PR is for the same reason as that with which it is correlated but not necessarily exactly the same code changes as those already reviewed and approved... Sometimes we do a fix for one branch that contains quite different code to the other depending on how the two branches may have diverged. In those cases we'd need to review and approve the different PRs individually although they may be for the same JIRA,

jomarko:
@manstis lesson learned, I will use **cherry-pick** vs **corollary** more carefully, sorry for inconvenience.

jomarko:
@manstis it can be merged I believe.

PR #775, opened 2018-01-26
Add reproducer tests into guided decision table backend

jomarko:
This PR adds reproducer tests for:
- RHBA-370
- RHDM-329


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/test/resources/org/drools/workbench/screens/guided/dtable/backend/server/rhba370/src/main/resources/META-INF/kie-deployment-descriptor.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
manstis: Do we need this file?

jomarko: Well spotted, removed now.

jomarko:
jenkins please retest this

jomarko:
@manstis jenkins says, green.

PR #777, opened 2018-01-26
DROOLS-2263 : Unexpected results in GDST when using enumerations with commas

Rikkola:

The fix is different than the one for 6.5.x. Still doing the same thing in the end. Just on different locations.

Bundle:
https://github.com/kiegroup/kie-soup/pull/29
https://github.com/kiegroup/drools/pull/1727
https://github.com/kiegroup/drools-wb/pull/777

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/ValueListUiColumn.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/ValueListUiColumn.java
@@ -18,6 +18,8 @@ package org.drools.workbench.screens.guided.dtable.client.widget.table.columns;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.ait.lienzo.client.core.shape.Text;
 import org.drools.workbench.screens.guided.dtable.client.widget.table.GuidedDecisionTablePresenter;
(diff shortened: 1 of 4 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/ValueListUiColumn.java
```diff
@@ -28,79 +28,104 @@
...
+                                       final GridBodyCellRenderContext context) {
+        t.setText(getLabel(value.toString()));
     }
 
+    private String getLabel(final String value) {
+        if (!isMultipleSelect) {
+            return getValue(value);
+        }
+        final StringBuilder sb = new StringBuilder();
+        final String[] values = value.split(",");
```
manstis: I know it's old code but should I be worried about the split here?

Rikkola: Double checking, perhaps the fix is not the best.

Rikkola: @manstis Ok, I think this is the best way to do it. Unless you know why the list is empty when the value is list. For example with a cell value "a,b" the a or b can not be found from the `valueListLookup` map.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/utilities/ColumnUtilities.java
```diff
@@ -142,7 +143,7 @@ public static boolean canAcceptOtherwiseValues(final BaseColumn column) {
     private String[] parseValueList(String fieldType,
                                     String valueList) {
 
-        final String[] values = valueList.split(",");
+        final String[] values = ListSplitter.split(valueList);
```
jomarko: Could we test this change in `org.drools.workbench.screens.guided.dtable.client.widget.table.utilities.ColumnUtilitiesGetValueListTest` class please?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/ValueListUiColumn.java
```diff
@@ -28,79 +28,104 @@
...
+                                       final String value,
+                                       final GridBodyCellRenderContext context) {
+        t.setText(getLabel(value.toString()));
     }
 
+    private String getLabel(final String value) {
+        if (!isMultipleSelect) {
+            return getValue(value);
+        }
+        final StringBuilder sb = new StringBuilder();
```
jomarko: I believe we don't need the `StringBuilder sb` and we can implement the `for` cycle below as:
```java
Stream.of(values).collect(Collectors.joining(","));
```

Rikkola: Changing, the code is more clear this way.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/utilities/ColumnUtilities.java
Rikkola: Tried with a break point and this does get hit with several different setups where the character , is in the string.

jomarko:
@manstis this is ready for merge.

Ensemble 1/4.

PR #779, opened 2018-01-29
BAQE-336: Increase coverage of drl file validation

jomarko:
This PR increases validation coverage of:
- DRL file with global varaible
- DRL file with explicit import


jomarko:
@manstis tests passed.

PR #780, opened 2018-01-29
[7.5.x] BAQE-336: Increase coverage of drl file validation

jomarko:
This PR increases validation coverage of:
- DRL file with global varaible
- DRL file with explicit import

This is **cherry-pick** of #779


jomarko:
@manstis tests passed.

PR #781, opened 2018-01-30
Increase coverage of DSLR files validation

jomarko:
This test is consequence of the BXMSDOC-2067 review.


jomarko:
@manstis ready for merge

PR #787, opened 2018-01-31
DROOLS-2263: Unexpected results in GDST when using enumerations with commas - info

Rikkola:

@manstis @jomarko 
I was planning to add this to the docs, but there is no good spot for it there and we already had the info panel in the Guided DTable editor.


jomarko:
@Rikkola maybe worth to cherry-pick into 7.5.x also.

Rikkola:
@jomarko Sure, waited for approvals on this one.

manstis:
@Rikkola Please let me know when this is "good to go"

Rikkola:
@manstis It is good to go. I was waiting for your and @jomarko's approvals :)

PR #788, opened 2018-01-31
BAPL-839 (for upstream master): Updated Template Key help text in GDT and guided rule designers

sterobin:
Details in [JIRA](https://issues.jboss.org/browse/BAPL-839).

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/resources/i18n/Constants.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/resources/i18n/Constants.java
@@ -278,6 +278,8 @@ public interface Constants
 
     String TemplateKey();
 
+    String TemplateKeyTip();
+
     String TemplateData();
 
     String LoadTemplateData();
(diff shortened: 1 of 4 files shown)


manstis:
@sterobin You'll need to add ```String TemplateKeyTip();``` to [this](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/resources/i18n/Constants.java) file.

sterobin:
@manstis, thanks. Done. We'll see how she builds.

drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/resources/i18n/Constants.java
```diff
@@ -278,6 +278,8 @@ String Append0ToList1( String varName,
 
     String TemplateKey();
 
+    String TemplateKeyTip();
```
manstis: :+1:

manstis:
@sterobin Let's see what Jenkins makes of your programming prowess!

sterobin:
@jomarko , my vote is to merge and submit a JIRA to widen the tooltip. It's a problem with all of the help icons in the window anyway. Plus it's better to have a hard-to-read help tip that essential, then not have it at all (in this case).

PR #791, opened 2018-02-01
[7.5.x] DROOLS-2263: Unexpected results in GDST when using enumerations with commas - info

Rikkola:
(cherry picked from commit f68c96ca7432a3787f9dadb9387f109bdc8227ea)

And the same for 7.5.x


Rikkola:
@barboras7 Something you might want to check.

Rikkola:
@manstis @jomarko You approved the original change to master. This is the same 1:1. Reviewes are welcomed.

manstis:
@Rikkola I assume this can be merge? Could you however please confirm.

Rikkola:
@manstis It is ok.

PR #794, opened 2018-02-02
DROOLS-2218: Catch NPE in CDI tests cleanup

jomarko:
@manstis please could you have a look?

This could "fix" [the problem](https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/drools-wb-pullrequests-7.5.x/46/testReport/org.drools.workbench.screens.dtablexls.backend.server/DecisionTableXLSServiceImplCDITest/testValidateColumnsNotInStandardOrder/), however I admit "fix" is not the right word. It is like hiding symptoms. I am sorry but I am not able to reproduce locally so this is the most I was able to do. Any hints?


jomarko:
jenkins please retest this

PR #796, opened 2018-02-02
RHDM-360: Increase coverage of enum indexing

jomarko:
@manstis please have a look


jomarko:
@manstis yes agreed, In my eyes, `EnumFileIndexer` is written in so "private" approach, that I can not do much more? Any suggestions welcome.

manstis:
@jomarko Using ```indexer.fillIndexBuilder(..).build()``` will give a ```Set<KProperty<?>>``` that should contain entries for both the Type (```Car```) and Fields (hopefully both ```colour``` and ```price``` however due to RHDM-360 this may not be the case).

jomarko:
@manstis would you mind re-review? I enhanced the test.

manstis:
Thanks @jomarko

PR #801, opened 2018-02-06
AF-980: Indexing: Improve on false negatives during parallel Unit Test execution

manstis:
See https://issues.jboss.org/browse/AF-980

Further to https://github.com/kiegroup/kie-wb-common/pull/1395 this is the PR for ```drools-wb```.

### DIFF
--- a/drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/src/test/java/org/drools/workbench/screens/drltext/backend/server/indexing/IndexRuleAttributeNameAndValueCompositionTest.java
+++ b/drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-backend/src/test/java/org/drools/workbench/screens/drltext/backend/server/indexing/IndexRuleAttributeNameAndValueCompositionTest.java
@@ -22,6 +22,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.concurrent.NotThreadSafe;
+
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.search.Query;
 import org.drools.workbench.screens.drltext.type.DRLResourceTypeDefinition;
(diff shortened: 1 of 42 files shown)


manstis:
@jomarko An interesting question! 

I think any test that relies on indexing should probably be annotated. New commit added.

manstis:
Successful x1... trying again.

Jenkins please retest this.

manstis:
Success x2.

Jenkins please retest this.

manstis:
Success x3.

Jenkins please retest this.

manstis:
Jenkins please retest this.

manstis:
Success x4.

Jenkins please retest this.

manstis:
Jenkins please retest this.

yurloc:
@manstis I think this change has no effect on parallel execution stability because [Surefire](https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html) looks for [the `net.jcip.annotations.NotThreadSafe` annotation](https://github.com/apache/maven-surefire/blob/surefire-2.19.1/surefire-providers/surefire-junit47/src/main/java/org/apache/maven/surefire/junitcore/pc/ParallelComputerBuilder.java#L240) and not `javax.annotation.concurrent.NotThreadSafe`, which is used in this PR.

We can either fix the imports or revert this PR. Since I believe the current annotations really have no effect I'd go for reverting this. On the other hand, the indexing tests stopped failing after this PR according to [random failure analysis](http://janhrcek.cz/random-failures/).

Wdyt?

manstis:
@yurloc I agree to revert (along with https://github.com/kiegroup/kie-wb-common/pull/1395)

I'll submit some PRs.

jhrcek:
@yurloc @manstis Just FYI: the reason why indexing tests stopped failing is more likely because of Max Barkley's [changes in indexing implementation](https://github.com/kiegroup/appformer/pull/255).

yurloc:
@jhrcek Thanks. I thought there existed something like that but I only look for this changed close to 2018-02-05 where the last failure of one of the indexing tests was recorded.

One more thing: Even if the imports were correct, the annotation would have no effect because we're not using [Surefire parallel execution](https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html). The paralellism that is currently in place is on the Maven build level.

PR #802, opened 2018-02-06
BAQE-201: Increase coverage for rdrl with abbreviated condition

jomarko:
This PR cover case when the rdrl file contains abbreviated conditions. It means condition like:
p : Person(age != 18 && > 15 && < 25)

Reproducer for RHBRMS-2854


jomarko:
@manstis just friendly reminder, this is ready for merge.

PR #803, opened 2018-02-06
BAQE-339: Cover loading action plugins into rule modeller actions

jomarko:

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/RuleModeller.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/RuleModeller.java
@@ -219,8 +219,7 @@ public class RuleModeller extends Composite
             addAction.addClickHandler(new ClickHandler() {
 
                 public void onClick(ClickEvent event) {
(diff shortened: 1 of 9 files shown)


drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/RuleModeller.java
```diff
@@ -481,16 +479,20 @@ protected void showConditionSelector(Integer position) {
...
-                                                                                    actionPlugins,
-                                                                                    position,
-                                                                                    getDataModelOracle());
+    protected void showActionSelector(final Integer position) {
+        final RuleModellerActionSelectorPopup popup = ruleModellerActionSelectorPopup(position, actionPlugins);
         popup.show();
     }
 
+    protected RuleModellerActionSelectorPopup ruleModellerActionSelectorPopup(final Integer position,
+                                                                              final Collection<RuleModellerActionPlugin> testableActionPlugins) {
```
manstis: Do you really want to call this ```testableActionPlugins``` since this is not test code?

jomarko: Updated to standard name.

jomarko:
@manstis green

PR #805, opened 2018-02-07
[6.5.x] DROOLS-2306 : Decision Table Verification does not update correctly

Rikkola:
https://issues.jboss.org/browse/DROOLS-2306

@jomarko @manstis Hi, you two are probably the best two to check this. If you could please review.


Rikkola:
And ping @barboras7

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/action/FieldActionInspector.java
```diff
@@ -68,6 +69,44 @@ private boolean isValueRedundant(final DTCellValue52 other) {
...
+            case DATE:
+                return nullSafeEquals(value.getDateValue(), that.getDateValue());
+            case BOOLEAN:
+                return nullSafeEquals(value.getBooleanValue(), that.getBooleanValue());
+            default:
+                return false;
+        }
+    }
+
+    private boolean nullSafeEquals(final Object value1,
```
manstis: I believe this can be replaced with ```Objects.equals(..)``` as I've seen @jomarko do similar changes in my code!

Rikkola: The name is a bit misleading. What it does it compares if the value is equal when `valueTypes` are the same.

The reason for this was that equals failed when both had the `valueType` boolean and the value was "true", but another one had `stringValue` as null and other one at "". 

This is the actual bug, the other changes actually fix other bugs that kept on draining memory as the dtable changed.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/action/FieldActionInspector.java
```diff
@@ -68,6 +70,31 @@ private boolean isValueRedundant(final DTCellValue52 other) {
         }
     }
 
+    private boolean valuesAreEqual(DTCellValue52 value, DTCellValue52 that) {
```
jomarko: Well, looking on `org.drools.workbench.models.guided.dtable.shared.model.DTCellValue52.equals()` implementation, I still think this `valuesAreEqual`method could be just `Objects.equals(value, that);`, however when I tried, the new added test fails, so probably keep as it is.

Rikkola: The reason why it can not is that sometimes even when the value type is boolean (or anything, but String). If either object also has the stringValue field filled, it can be "" or null, the equals fails. Even though the real values are equal.

This is the original bug. Other having "" and other having null. So even if we detected this with boolean it does apply to everything, but the String.

Rikkola:
@manstis or who ever merges this one. I just added @barboras7 in to keep her up to date with my work. So I think this is good to merge.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/analysis/action/FieldActionInspector.java
jomarko: Yes, I noticed this during debug that boolean value and empty string was stored at the same time, I decided to not invest more time as it is on 6.5.x branch.

Rikkola:
Ok. Acks are in

PR #808, opened 2018-02-09
BAQE-339: Cover validation of rule/gdst file with scoreHolder inside

jomarko:
@manstis please have a look


jomarko:
@manstis I will try to do.

jomarko:
Closing as this is not possible without adding optaplanner dependencies

PR #811, opened 2018-02-16
RHDM-251 - Wildfly 11 upgrade

mareknovotny:
This repeated effort of https://github.com/kiegroup/drools-wb/pull/765


mareknovotny:
Please verify also GWT SDM as this could fail with ClassCastException while trying to use JBoss logging.
@mbarkley created for that a special extension which can disable that JBoss logging, but he tried it and it seems to not needed. See https://github.com/kiegroup/kie-wb-common/pull/1431

mareknovotny:
ok, last commit from @mbarkley fixes the server logging which was not showing properly.

PR #814, opened 2018-02-21
[DO NOT MERGE] DROOLS-2341: Ignore tests while the issue is not fixed

jomarko:
Once the issue DROOLS-2341 tests ignored in this commit should be unignored and possibly enhanced to check the chatched error messages belongs to the correct asset.

Please have a look @manstis @hasys . This is needed to get F(ull)D(ownstream)B(uild) working.


hasys:

manstis:
This may not be necessary. See https://github.com/kiegroup/kie-wb-common/pull/1443

PR #820, opened 2018-02-28
BAQE-339: Cover Guided Rule service for modify score action in rule

jomarko:
@manstis please have a look.


jomarko:
@manstis can be merged

PR #821, opened 2018-03-01
Removing dead code

Rikkola:
Doing some changes into AuthoringPerspectives and I think the only ones we actually still use are in kie-wb-distributions. So these two can go.


manstis:
@Rikkola I assume then these perspectives are no longer used and Library offers all we need?

Rikkola:
@manstis Yes, these were just in -webapp for drools-wb needs. Library now runs the "Authoring" views. The AuthoringPerspectives in kie-wb-distributions are still used by embed editors, but they have no dependencies on these.

PR #822, opened 2018-03-05
BAQE-376: Cover constraint operator

jomarko:
@manstis please have a look

Sorry this time a slightly more complicated than just adding tests

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/widget/FactPatternWidget.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/widget/FactPatternWidget.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import com.google.gwt.core.client.GWT;
 import com.google.gwt.event.dom.client.ChangeEvent;
 import com.google.gwt.event.dom.client.ChangeHandler;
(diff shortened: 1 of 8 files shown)


drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/widget/FactPatternWidget.java
```diff
@@ -664,145 +660,26 @@ public void execute() {
...
-                                col,
-                                new Callback<CEPOperatorsDropdown>() {
-                                    @Override
-                                    public void callback(CEPOperatorsDropdown result) {
-                                        hp.add(result);
-                                    }
-                                });
+            final SingleFieldConstraintOperatorSelectorBuilder operatorSelectorBuilder =
+                    GWT.create(SingleFieldConstraintOperatorSelectorBuilder.class);
+            operatorSelectorBuilder.hasConstraint(constraint);
```
manstis: I think the builder pattern is great for when you're constructing a complex object with optional **configurations** however in this case they all seem **mandatory** and IDK what value having separate "has" setters has - in fact it make construction error prone as for example if I don't call ```hasPlaceholderForOperatorsDropdown(..)``` then I will get a runtime NPE error. Please either consider replacing all "has" setters with a single "configure(..)" method that takes all mandatory parameters. Or, at least in the ```build()``` method validate all mandatory parameters have been set.

jomarko:
@manstis thanks for review, regarding builder pattern, definitely good point, I changed the code, please let me know if there is anything I should still improve.

PR #823, opened 2018-03-07
DROOLS-2362: Trim whitespace in valueList string

jomarko:
@manstis please have a look


jomarko:
@manstis ready for merge.

PR #824, opened 2018-03-07
DROOLS-2371: Unable to complete your request. The following exception occurred: (TypeError) in Guided Decision Table Graph

manstis:
See https://issues.jboss.org/browse/DROOLS-2371

The problem I found was that with a GDT open and then opening a GDTG(raph) and adding the same GDT a ```DecisionTableSelectedEvent``` was raised. The event held a different instance of the first GDT and hence the singular GDT editor set its "active decision table" to ```null```. However the same event also causes the "Columns" tab to refresh itself assuming the "active decision table" is not null... <bang!>

This PR fixes that and also safe guards against similar (undetected) issues by making "active decision table" a ```java.util.Optional``` and refactoring the old use of it accordingly. There will no doubt need to be some tests written (there are no new ones in this PR at the moment) but I thought it worth checking with you both that this does indeed fix the issues you encountered.


manstis:
Jenkins please execute full downstream build (this should give a WAR to test somewhere..)

jomarko:

jomarko:
@manstis I build locally and can confirm issue is gone, let me know when you are done with tests.

manstis:
@jomarko This now has Unit Tests and is "good for review". Thanks in advance.

IDK if we'll need @tkobayas review if you've built and tested locally.

PR #830, opened 2018-03-13
RHBA-520: Cover validation of huge spreadsheet table

jomarko:
@Rikkola @manstis please have a look.


manstis:
@jomarko Given https://issues.jboss.org/browse/RHBA-520 do you want this merged? 

Doesn't it run the risk of OOM'ing Jenkins?

jomarko:
OOM occurs during indexing, validation should be fine, we can run tests here (10 times lets say) and check if OOM occurs. WDYT?

manstis:
@jomarko No, if this build is green then I am OK with merging as it is.

jomarko:
@manstis hehe, you was right: `Caused by: java.lang.OutOfMemoryError: Java heap space`, so rpobably this can't be merged.

PR #834, opened 2018-03-23
RHDM-390,RHDM-391: Workaround for ERRAI-1101

yurloc:
Related PRs:
kiegroup/appformer/pull/258
kiegroup/kie-wb-common/pull/1535
kiegroup/drools-wb/pull/834
kiegroup/optaplanner-wb/pull/266
kiegroup/jbpm-wb/pull/1011
kiegroup/kie-wb-distributions/pull/716

Works around: https://issues.jboss.org/browse/ERRAI-1101
Fixes: https://issues.jboss.org/browse/RHDM-390, https://issues.jboss.org/browse/RHDM-391


manstis:
Hi @yurloc have you checked these changes with our Zanata [integration](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L1134)?

You might want to ask @mbiarnes to check whether the CI jobs to push to Zanata work OK with these modifications. Furthermore you might want to check Zanata integration when we pull back works OK too as we rename some files. These files' names will be based on the "base" file name (i.e. that now contains ```_default``` and hence current [configuration](https://github.com/kiegroup/kie-wb-distributions/blob/master/kie-wb-parent/kie-wb-webapp/pom.xml#L2051) likely to break). 

There will no doubt be more places to fix than these I list :-(

yurloc:
@manstis Thanks for bringing this up! I have got another approach that might be safer with respect to Zanata integration. Instead of committing the renamed files, I could do this on-the-fly during maven build. I just need to copy MyConstants.properties to MyConstants_default.properties. The original MyConstants.properties must be present and it doesn't matter if it's empty or has the original content.

What do you think? Is this a better approach to avoid risk of breaking Zanata integration?

manstis:
@yurloc It could be the safer option; we rename i18n files used by the i18n taglib for the "login page" [already](https://github.com/kiegroup/kie-wb-distributions/blob/master/kie-wb-parent/kie-wb-webapp/pom.xml#L2051).

yurloc:
@manstis are you sure the renaming you just mentioned is needed? I can't see the source files.

PR #847, opened 2018-04-12
DROOLS-2371: Increase coverage

jomarko:
@manstis please have a look

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/lockmanager/GuidedDecisionTableLockManagerImplTest.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/lockmanager/GuidedDecisionTableLockManagerImplTest.java
@@ -138,6 +138,24 @@ public class GuidedDecisionTableLockManagerImplTest {
     }
 
     @Test
+    public void testFireChangeTitleEvent_LockInfoUpdateForActiveDecisionTableMissingCurrentPath() {
+        lockManager.init(mock(LockTarget.class),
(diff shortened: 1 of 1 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/lockmanager/GuidedDecisionTableLockManagerImplTest.java
```diff
@@ -137,6 +137,24 @@ public void testFireChangeTitleEvent_LockInfoUpdateForActiveDecisionTable() {
...
+    @Test
+    public void testFireChangeTitleEvent_LockInfoUpdateForActiveDecisionTableMissingCurrentPath() {
+        lockManager.init(mock(LockTarget.class),
+                         modellerPresenter);
+
+        final GuidedDecisionTableView.Presenter dtPresenter = mock(GuidedDecisionTableView.Presenter.class);
+        final PathFactory.PathImpl lockInfoPath = makePathImpl("filename");
+
+        when(lockInfo.getFile()).thenReturn(lockInfoPath);
+        when(dtPresenter.getCurrentPath()).thenReturn(null);
```
manstis: @jomarko Whilst this test does no harm; it should be impossible to have a ```GuidedDecisionTableView.Presenter``` instance without "current path" set. It might be worth checking/adding there's a test for that too?

jomarko: @manstis I am lost, I think this test checks what happens if current path would be null, what test should I look for please?

jomarko:
Closing after discussion

PR #851, opened 2018-04-16
DROOLS-2477: Update column wizard hints due to ui changes

jomarko:
@manstis please have a look


jomarko:
@manstis this is ready for merge

PR #859, opened 2018-05-09
DROOLS-2477: Align hint with latest UI

jomarko:
@manstis please have a look when you have a moment


jomarko:
@manstis ready for merge

PR #867, opened 2018-05-31
RHDM-600: latest Zanata translation changes in drools-wb from 2018-05-31

kiereleaseuser:


mbiarnes:
@manstis can you please merge this

PR #869, opened 2018-06-06
DROOLS-2617: Cover adding imports tab to guided designers

jomarko:
This PR is a corollary of internal selenium tests migration into community.

@manstis would you mind to have a look?


jomarko:
@manstis I think this can be merged.

PR #870, opened 2018-06-07
[do not merge] DROOLS-2623: Stabilize xls decision table tests

jomarko:
Unify drools date format initialization to get tests not affecting each other


jomarko:
@manstis @karreiro @Rikkola anybody interested, please have a look.

Rikkola:
Sounds and looks like a good change. +1

jomarko:
Well jenkins obviously behave differently than local machine, argh. Keeping this open a and will try to investigate.

jomarko:
@manstis will try, as you can see, it fails on jenkins, it is fine on local machine.

jomarko:
Replaced with #1037

PR #875, opened 2018-06-15
DROOLS-1906: Translate frozen_conditions and frozen_actions metadata

jomarko:
Guided rule editor supports two specific metadata (see the title) that tells the editor the conditions or actions can not be modified. This PR translates those metadata to more human readable text.

@Rikkola would you mind to have a look as a reporter of the [issue](https://issues.jboss.org/browse/DROOLS-1906) ?


jomarko:
@Rikkola @manstis would you mind to have a look please?

PR #879, opened 2018-06-28
DROOLS-2662: Prevent NPE when no dropDownData available

jomarko:
For more details, see the issue [DROOLS-2662](https://issues.jboss.org/browse/DROOLS-2662)

@manstis @Rikkola I will be pleased with your review, especially I am interested if the added tests reproduce/verify the issue enough.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/ConstraintValueEditor.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/ConstraintValueEditor.java
@@ -628,7 +628,7 @@ public class ConstraintValueEditor extends Composite {
                 final Button bindingButton = new Button(GuidedRuleEditorResources.CONSTANTS.BoundVariable());
 
                 //This Set is used as a 1flag to know whether the button has been added; due to use of callbacks
-                final Set<Button> bindingButtonContainer = new HashSet<Button>();
+                final Set<Button> bindingButtonContainer = new HashSet<>();
 
                 for (String var : bindingsInScope) {
                     helper.isApplicableBindingsInScope(var,
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/ConstraintValueEditor.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/ConstraintValueEditor.java
@@ -131,6 +131,11 @@ public class ConstraintValueEditor extends Composite {
 
         setUpConstraint();
 
+        // DROOLS-2662
+        // Call refresh() before ConstraintValueEditorHelper is created
+        // This ensures dropDownData is not null
+        refresh();
+
         helper = new ConstraintValueEditorHelper(model,
                                                  oracle,
                                                  factType,
@@ -138,8 +143,6 @@ public class ConstraintValueEditor extends Composite {
                                                  constraint,
                                                  fieldType,
                                                  dropDownData);
-
-        refresh();
         initWidget(panel);
     }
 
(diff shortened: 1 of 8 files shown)


drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/util/ConstraintValueEditorHelper.java
```diff
@@ -239,7 +240,12 @@ private void isLHSFieldTypeEquivalent(final String boundVariable,
     }
 
     public static boolean isEnumEquivalent(final String[] values, final DropDownData dropDownData) {
-        return Arrays.equals(values, dropDownData.getFixedList());
```
jomarko: This was the root cause, the `dropDownData` was null.

jomarko:

drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/util/ConstraintValueEditorHelper.java
manstis: @jomarko Do you know why the ```dropDownData``` is ```null``` for ```java.time.LocalDate``` but not, for example, when ```java.util.Date``` (the JIRA states the issue is only evident when using ```LocalDate```)? I have no doubt this PR fixes the symptom but I'd like to know why ```dropDownData``` is ```null``` in the first place.

manstis: @jomarko I see ```ConstraintValueEditor``` instantiates ```ConstraintValueEditorHelper``` in its constructor and, at this point in time, ```dropDownData``` is guaranteed to always be ```null```. It is only when ```ConstraintValueEditor.refresh()``` is called (after ```ConstraintValueEditorHelper``` has been instantiated) that the dropdown data is actually loaded from the DMO.

IMO we should probably also look at correcting the initialisation order in ```ConstraintValueEditor```.

jomarko: @manstis I added commit according your suggestion.

jomarko:

drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/util/ConstraintValueEditorHelper.java
```diff
@@ -239,7 +240,12 @@ private void isLHSFieldTypeEquivalent(final String boundVariable,
...
+        if (dropDownData != null) {
+            return Arrays.equals(values, dropDownData.getFixedList());
+        } else {
+            // droopDownData is null, so check if also values are null
```
manstis: Spelling: "dropDownData" not "droopDownData".

drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/ConstraintValueEditor.java
```diff
@@ -131,15 +131,18 @@ public ConstraintValueEditor(BaseSingleFieldConstraint con,
 
         setUpConstraint();
 
+        // DROOLS-2662
+        // Call refresh() before ConstraintValueEditorHelper is created
+        // This ensures dropDownData is not null
+        refresh();
```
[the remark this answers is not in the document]
manstis: @Rikkola @jomarko ```Mockito.InOrder()```

jomarko:
Just for record, the reason why `dropDownData` was null for `java.time.LocalDate` field is that `AsyncPackageDataModelOracleImpl` assumes `java.time.LocalDate` fields have type `Comparable` while `java.util.Date` fields have type `Date`. 

This difference caused that different branches of code were executed for different field types and we didn't spot the issue sooner.

If we decide to support `java.time.LocalDate` we will have to  update [converter](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-datamodel/kie-wb-common-datamodel-backend/src/main/java/org/kie/workbench/common/services/datamodel/backend/server/builder/projects/JavaTypeSystemTranslator.java)

jomarko:
@manstis this can be merged

PR #881, opened 2018-07-03
[6.5.x] GUVNOR-3244 (RHBRMS-3103): [GSS] (6.4.z) "Parameter named 'fieldName' should be not null" when indexing a GDST with BRL fragment eval

manstis:
See https://issues.jboss.org/browse/RHBRMS-3103

The fix for ```6.5.x``` is to skip constraints that cannot be indexed (vs throwing the NPE).

This is different to ```master``` as the codebases have diverged and ```master``` does not _visit_ the model.


jomarko:
I checked manually, in server log is not the error more, instead of it there is logged this info:
```
12:11:14,725 INFO  [org.drools.workbench.screens.guided.dtable.backend.server.indexing.GuidedRuleModelIndexVisitor] (EJB default - 9) Unable to index constraint FactType[null], FieldName[null], FieldType[]. Skipping.
```

manstis:
@jomarko Log message updated in new commit.

PR #886, opened 2018-07-11
DROOLS-2690 : Create a Scenario asset

Rikkola:
https://issues.jboss.org/browse/DROOLS-2690

This works is based on what @gitgabrio started. The goal is to get the modules for both GWT and Maven in to avoid conflicts with master and to make rebasing easier.

The actual codes in the modules are still experimental, but only visible to drools-wb-webapp. Before we expose this in the actual wars there needs to be other commits that clean up the setup.

Bundle:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/774
https://github.com/kiegroup/kie-wb-common/pull/1972
https://github.com/kiegroup/drools-wb/pull/886


Rikkola:
Need to update the years in license headers before @manstis sees this!

Rikkola:
@gitgabrio I updated the headers so the licenses are correct and reformatted each class to the default code format.

Rikkola:
@kkufova You might also want to check this.

drools-wb-webapp/src/main/resources/org/drools/workbench/DroolsWorkbench.gwt.xml
```diff
@@ -88,10 +93,16 @@
 
   <!-- Specify the paths for translatable code -->
   <source path='client'/>
+  <source path="org.drools.workbench.screens.scesim.client" />
```
Rikkola: This actually needs to be changed.

Rikkola: Changing it.

Rikkola:
Jenkins runs more tests against the code. Let's see what it thinks.

gitgabrio:
+1

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scesim/backend/server/SceSimServiceImpl.java
```diff
@@ -0,0 +1,310 @@
...
+import org.uberfire.ext.editor.commons.service.CopyService;
+import org.uberfire.ext.editor.commons.service.DeleteService;
+import org.uberfire.ext.editor.commons.service.RenameService;
+import org.uberfire.io.IOService;
+import org.uberfire.java.nio.file.FileAlreadyExistsException;
+import org.uberfire.rpc.SessionInfo;
+import org.uberfire.workbench.events.ResourceOpenedEvent;
+
+/**
+ *
```
kkufova: Empty Javadoc?

drools-wb-scenario-simulation/drools-wb-scenario-simulation-webapp/src/main/java/org/drools/workbench/scenario/simulation/showcase/client/screens/editor/SessionDiagramEditorScreen.java
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    private void validateAndSave() {
+        GWT.log("validateAndSave");
+        final Command save = this::save;
+    }
+
+    private void save() {
+        GWT.log("save");
+        /*diagramService.save(getSession(),
```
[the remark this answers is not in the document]
Rikkola: Just cleaning it up with the next PR. I could add it into this if needed.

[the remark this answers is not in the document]
Rikkola: Yes there is. The code section is still a playground for @gitgabrio. He will be working on everything that is under Scenario Simulation Presenter.

I already removed some of the commented codes. Currently adding save/load with proper unit test support.

drools-wb-scenario-simulation/drools-wb-scenario-simulation-webapp/src/main/java/org/drools/workbench/scenario/simulation/showcase/client/screens/editor/SessionDiagramEditorScreen.java
```diff
@@ -0,0 +1,417 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
```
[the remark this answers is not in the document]
Rikkola: This was actually dead code. Well spotted.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scesim/backend/server/SceSimServiceImpl.java
```diff
@@ -0,0 +1,310 @@
...
+    public List<ValidationMessage> validate(final Path path,
+                                            final String content) {
+        return doValidation(path,
+                            content);
+    }
+
+    private List<ValidationMessage> doValidation(final Path path,
+                                                 final String content) {
+        try {
+            final KieModule module = moduleService.resolveModule(path);
```
manstis: Presumably this was copied from the existing Enumeration editor? 

I assume this method code will be replaced in the future.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scesim/backend/server/SceSimServiceImpl.java
```diff
@@ -0,0 +1,310 @@
...
+        try {
+            Metadata currentMetadata = metadataService.getMetadata(resource);
+            ioService.write(Paths.convert(resource),
+                            content,
+                            metadataService.setUpAttributes(resource,
+                                                            metadata),
+                            commentedOptionFactory.makeCommentedOption(comment));
+
+            //Invalidate Package-level DMO cache as SceSims have changed.
+            invalidateDMOPackageCache.fire(new InvalidateDMOPackageCacheEvent(resource));
```
manstis: Presumably this was copied from the existing Enumeration editor? 

There's probably no need to invalidate the DMO cache when a Simulation is saved?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scesim/backend/server/SceSimInvalidateDMOPackageCacheDeleteHelper.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.drools.workbench.screens.scesim.type.SceSimResourceTypeDefinition;
+import org.guvnor.common.services.project.builder.events.InvalidateDMOPackageCacheEvent;
+import org.kie.workbench.common.services.backend.helpers.AbstractInvalidateDMOPackageCacheDeleteHelper;
+
+/**
+ * DeleteHelper for SceSimerations to invalidate LRUDataModelOracleCache entries when an SceSimeration is deleted.
+ */
+@ApplicationScoped
+public class SceSimInvalidateDMOPackageCacheDeleteHelper extends AbstractInvalidateDMOPackageCacheDeleteHelper<SceSimResourceTypeDefinition> {
```
manstis: Presumably this was copied from the existing Enumeration editor? 

There's probably no need to invalidate the DMO cache when a Simulation is deleted?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scesim/backend/server/indexing/SceSimFileIndexer.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public boolean supportsPath(final Path path) {
+        return type.accept(Paths.convert(path));
+    }
+
+    @Override
+    public DefaultIndexBuilder fillIndexBuilder(final Path path) throws Exception {
+        final String scesimDefinition = ioService.readAllString(path);
```
manstis: Presumably this was copied from the existing Enumeration editor?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scesim/backend/server/indexing/SceSimIndexVisitor.java
```diff
@@ -0,0 +1,128 @@
...
+import org.kie.workbench.common.services.refactoring.service.PartType;
+import org.kie.workbench.common.services.refactoring.service.ResourceType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.uberfire.java.nio.file.Path;
+
+/**
+ * Visitor to extract index information from a DataEnumLoader
+ */
+public class SceSimIndexVisitor extends ResourceReferenceCollector {
```
manstis: Presumably this was copied from the existing Enumeration editor?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scesim/client/editor/SceSimEditorPresenter.java
```diff
@@ -0,0 +1,192 @@
...
+import org.uberfire.ext.editor.commons.service.support.SupportsSaveAndRename;
+import org.uberfire.lifecycle.OnClose;
+import org.uberfire.lifecycle.OnMayClose;
+import org.uberfire.lifecycle.OnStartup;
+import org.uberfire.mvp.Command;
+import org.uberfire.mvp.PlaceRequest;
+import org.uberfire.workbench.model.menu.Menus;
+
+/**
+ * Enum Editor Presenter
```
manstis: Yes it was!! (copied from the existing Enumeration editor)

manstis:
FWIW I don't think this should be merged into ```kie-group\drools-wb``` at all.

It looks a candidate for merge into @gitgabrio fork of ```kie-group\drools-wb``` if he's having problems getting started. I don't _buy_ the argument that it'd make conflict resolution and rebasing easier as new editors are invariably completely isolated from existing code anyway.

Still.. I'm not owner etc of this work so will just vent my frustration here and move on.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scesim/backend/server/SceSimServiceImpl.java
Rikkola: Yep I'm already cleaning the Presenter backend Service pipeline.

Rikkola:
The modules are not isolated. Thinking about dependencies that we have and version number changes. Then we also will start depending on services that might change. 

Buuuut....

I do think you are right. We should not rush this in as it is now. No test coverage, commented code, misleading names like "enum". The commits need to be clean.

I'm cleaning this up right now. We do however need to commit a mocked up editor where you can't really edit anything. This is because of the fact that this is what makes the sprint a success.

Rikkola:
@kkufova Cleaned up and deleted temp code I can not unit test. Making unit tests for the rest. Will take a while and we do need the right icon for this to go in ( or a ticket that says "add the correct icon" ).

jomarko:
Just my comment about merging here. I think merging here is is not problem if it is tested and doesn't throw error by every second click in that editor. From my point of view it can be really simple for now, just open and store empty file with the new **.scesim** extension. Then improving by small PRs is more preferable than opening PR with thousands lines of code.

Rikkola:
@jomarko Yep. It wasn't tested so I think that was one of the biggest problems. I'm still trying to figure out these incremental commits that we should now do.

jomarko:
@Rikkola I think everyone of us is figuring out for now.

Rikkola:
@kkufova @gitgabrio @manstis @jomarko Ok. The PR updated. 

* Everything added should have unit tests. 
* SceSim written out to ScenarioSimulation
* scesim package to scenariosimulation
* References to enumerations was removed

The editor doesn't show anything, but I don't think adding mocks makes any sense.

Rikkola:
Thanks @kkufova. Yep I think this quality is more closer to what the incremental merges should be.

So if everyone is happy this can be merged. I'm working on the next thing on top of this.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/pom.xml
```diff
@@ -0,0 +1,210 @@
...
+
+    <dependency>
+      <groupId>org.jboss.spec.javax.servlet</groupId>
+      <artifactId>jboss-servlet-api_3.1_spec</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+  </dependencies>
+
+  <dependencyManagement>
```
manstis: To be moved to ```droolsjbpm-build-bootstrap``` (don't you already have a PR for that repository?)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/pom.xml
```diff
@@ -0,0 +1,210 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.spec.javax.servlet</groupId>
```
manstis: Do you really need this for tests?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/pom.xml
```diff
@@ -0,0 +1,184 @@
...
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>lienzo-tests</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+  </dependencies>
+
+
+  <dependencyManagement>
```
manstis: Should this be included in a PR for ```droolsjbpm-build-bootstrap```?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/resources/images/ScenarioSimulationEditorImageResources.java
```diff
@@ -0,0 +1,28 @@
...
+package org.drools.workbench.screens.scenariosimulation.client.resources.images;
+
+import com.google.gwt.resources.client.ClientBundle;
+import com.google.gwt.resources.client.ImageResource;
+
+public interface ScenarioSimulationEditorImageResources
+        extends
+        ClientBundle {
+
+    @Source("BPM_FileIcons_scesim.png")
```
manstis: Might be worth renaming from ```BPM_FileIcons_scesim.png```?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/pom.xml
Rikkola: I might be trusting the pre-set dependencies too much here. I'll comment everything out and add everything back based on needs.

Rikkola: Hehe. Looks like every other import is not needed.

Rikkola:
@manstis Thank you for the reviews. Managed to remove a lot of unwanted deps.

I wonder how many of those we have in the poms in general. I remember hearing that the enforcer plugin would at some point also figure out unused imports.

jomarko:
@Rikkola unused imports should be caught already, see [1].

[1]
https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L623

Rikkola:
@jomarko Those do get picked up, but I'm talking about a dependency jar and the situation when nothing is used from that jar. We have the opposite, if an import is used from a transitive dependency we get a warning.

Rikkola:
Ok. I think we can merge this in.

Rikkola:
@manstis Hi, I tried to reach others to merge this, but I hear you might be the best. If you could please merge the changes in.

PR #893, opened 2018-07-18
RHPAM-1401: test the issue on backend side

jomarko:
This PR adds the reported scenario to the backend tests

@manstis @Rikkola please have a look, this just add tests to the backend. It doesn't fix the issue. The problem is in the client module, especially on this [line](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioEditorPresenter.java#L171).


Rikkola:
What is wrong with me? I keep leaving words out of my sentences. Too hot here. :-)

jomarko:
@Rikkola better message now please?

jomarko:
@Rikkola just reminder, is the commit message better now?

PR #896, opened 2018-07-25
RHDM-693: Load table using different timezones

jomarko:
See [RHDM-693](https://issues.jboss.org/browse/RHDM-693) for more details.

@manstis please have a look.

My assumption was the problem is table model loading. But this test says it is ok using different timezones.


manstis:
@jomarko I'd have assumed the same (I had a quick, cursory look yesterday and thought since we store the ```2018-07-31 21:00:00.0 UTC``` as the ```String``` value in the XML it was converting differently for different time-zones). Test seems OK so will approve.. obviously it needs some more looking into :-(

I've mentioned to @etirelli that there are a couple of JIRAs (other than DMN) that may need my attention in this sprint. He's not said otherwise, so after I'm done with my current DMN JIRA I'll take a look.

jomarko:
@manstis probably ready for merge.

PR #898, opened 2018-07-27
DROOLS-2781 : [Guided Rule] Can't be opened if constraint to bound field present

Rikkola:
https://issues.jboss.org/browse/DROOLS-2781


Rikkola:
@jomarko If you could please take a look. I think this was introduced with DROOLS-2662. I tried myself and could not reproduce DROOLS-2662 or DROOLS-2781.

manstis:
So the initialisation order broke things... that'd be my fault [then](https://github.com/kiegroup/drools-wb/pull/879#discussion_r198831288). Good now however that we have tests to ensure the initialisatin order.

Rikkola:
@manstis the code is a bit of a spider web. Each method call in the InOrder test should really work individually, but afraid to do a bigger clean up. Maybe we hit one more issue with this.

Rikkola:
Ok. This can now be merged in.

PR #900, opened 2018-07-30
[DROOLS-2672] Create static Scenario grid based on Uberfire Wires Core Grids

gitgabrio:
This PR contains the first working implementation of the Grid to be used by the editor - based on the uberfire wires core grid. The grid is shown inside a lienzo panel, i.e. as a canvas.
For the moment being it just show the header with three statically defined columns.

To be viewed by  @manstis , @Rikkola and @kkufova.

Please Note: currently the grid is using the style copied from DMN editor. We should define and use a specific style.


kiereleaseuser2:

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/metadata/ScenarioHeaderMetaData.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
```
Rikkola: 2018

kkufova: @gitgabrio, please fix all occurrences. :-)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -68,13 +68,12 @@ public ScenarioSimulationEditorPresenter() {
...
-    public ScenarioSimulationEditorPresenter(final ScenarioSimulationView baseView,
-                                             final Caller<ScenarioSimulationService> service,
+    public ScenarioSimulationEditorPresenter(final Caller<ScenarioSimulationService> service,
                                              final ScenarioSimulationResourceType type,
                                              final ImportsWidgetPresenter importsWidget,
                                              final AsyncPackageDataModelOracleFactory oracleFactory) {
-        super(baseView);
-        this.view = baseView;
+        super();
+        this.baseView = newScenarioSimulationView();   // Indirection added for test-purpose
```
Rikkola: You should use the super(baseView) to provide the view for base and then set the view for this class so that you don't need to later cast for the actual view.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/ScenarioSimulationViewProvider.java
```diff
@@ -0,0 +1,55 @@
...
+import org.drools.workbench.screens.scenariosimulation.client.editor.ScenarioSimulationViewImpl;
+import org.drools.workbench.screens.scenariosimulation.client.handlers.ScenarioSimulationGridPanelContextMenuHandler;
+import org.drools.workbench.screens.scenariosimulation.client.models.ScenarioGridModel;
+import org.drools.workbench.screens.scenariosimulation.client.renderers.ScenarioGridRenderer;
+import org.drools.workbench.screens.scenariosimulation.client.widgets.ScenarioGrid;
+import org.drools.workbench.screens.scenariosimulation.client.widgets.ScenarioGridLayer;
+import org.drools.workbench.screens.scenariosimulation.client.widgets.ScenarioGridPanel;
+
+/**
+ * Class used to instantiate a <code>ScenarioSimulationViewImpl</code> with all the contents/handlers required, avoiding CDI
```
Rikkola: Is there a need to skip CDI?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/metadata/ScenarioHeaderMetaData.java
Rikkola: And there are several more. For legal reason each one has to be updated to this year.

gitgabrio: Ok I'll do immediately

Rikkola:
@gitgabrio One more thing. The commit message should have the Jira ticket number and title.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/metadata/ScenarioHeaderMetaData.java
gitgabrio: Fixed with [DROOLS-2672] Create static Scenario grid based on Uberfire Wires Core Grids - Fixed copyright year

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
gitgabrio: I've removed the injection of baseview to reduce CDI usage (as you mention on next comment). Then, I've add  "newScenarioSimulationView" for testing purposes.
I've just restored usage of ScenarioSimulationView view to avoid casting with

[DROOLS-2672] Create static Scenario grid based on Uberfire Wires Core Grids - Restoring usage of "ScenarioSimulationView view" property to avoid casting.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/ScenarioSimulationViewProvider.java
Rikkola: You should still be able to create that one element in the ViewImpl and still be able to inject the View here. There are few widgets like that, but for consistency you should inject the Views.

And about CDI usage. If we change the way we use it here then we need to change it everywhere. Email in bsig might be the best tool for that.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/BaseExpressionGridTheme.java
```diff
@@ -0,0 +1,186 @@
...
+                .setTextUnit(TextUnit.PT)
+                .setListening(false)
+                .setTextBaseLine(TextBaseLine.MIDDLE)
+                .setTextAlign(TextAlign.CENTER);
+    }
+
+    @Override
+    public Rectangle getBodyBackground(final GridColumn<?> column) {
+        final Rectangle background = new Rectangle(0, 0);
+       /* if (column instanceof NameColumn) {
```
Rikkola: If these are not needed, please remove them.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/ScenarioSimulationViewProvider.java
gitgabrio: Well, maybe I was not clear, but I am talking only about CDI usage inside this module, for that specific component. Of course general usage of CDI is not in discussion - I am just referring to this specific case

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/widgets/ScenarioGrid.java
```diff
@@ -0,0 +1,75 @@
...
+    private void setHeaderColumns(Map<Integer, String> headersMap) {
+        headersMap.forEach((columnIndex, columnTitle) ->
+                                   model.insertColumn(columnIndex, getScenarioGridColumn(columnTitle, scenarioGridPanel, scenarioGridLayer)));
+
+
+    }
+
+    private void appendRows(Map<Integer, Map<Integer, String>> rowsMap) {
+        if (!rowsMap.isEmpty()) {
+            rowsMap.forEach((rowIndex, cellValueMap) -> {
```
jomarko: I believe using `forEach` ensures nothing is done if collection is empty. We should remove the outer `if`.

jomarko: Thanks!

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/BaseExpressionGridTheme.java
gitgabrio: Fixed with
[DROOLS-2672] Removing commented code from BaseExpressionGridTheme

gitgabrio:
@kkufova 
Point 1 - currently there is not any "persistence" - so every time you open the grid it just shows up like the first time; moreover, this "moving around" is just a consequence of using lienzo, but I am not sure it is a wanted, nor needed, feature - I mean: does it make any sense to move that table around ? If so, we have to take in account this in the data model, otherwise probably it is better to block this option
Point 2 - Yes, there are a lot of new classes, but for the most part they are UI-related, with basically no behaviour; in our tests the ui part is actually mocked, so I am not so sure about what  to test - I'll take a look

gitgabrio:
@kkufova 
About test coverage, do you have any suggestion ? Is there any specific class/method you think should be tested the way we are currently testing everything else ?

Rikkola:
@kkufova Is right. I assumed the following classes were copy pasted and would be removed at some point, but due to the new agile "commit as often as possible" method anything that goes in should have tests or they never get done. I believe they are copy pasted and perhaps edited after that. So there might already be a test that can be copied and edited to fit the needs.

I took a quick look and at least the following have logic in them that should be covered:
- ScenarioGridModel
- BaseGridRenderer
- ScenarioGrid

kkufova:
@gitgabrio
1. In my opinion, moving the table is not needed if there will be only one table on the canvas. However, if it will be possible to add more tables (I am not completely sure, but I think I saw it somewhere in the code), moving them on the canvas should be an option to allow users to reposition them based on their needs. WDYT?
2. I agree with @Rikkola, we should be adding as many tests as possible. The classes suggested by Toni are exactly the ones I was thinking about, especially the `BaseGridRenderer` (446 lines of code and no test).

gitgabrio:
@Rikkola @kkufova Actually that most infamous BaseGridRenderer is not necessary - I have removed it. About other tests, fixed with:
[DROOLS-2672] Implemented tests. Removed unused BaseGridRenderer class.

gitgabrio:
@kkufova Now grid does not move with:

[DROOLS-2672] Making grid unmovable. Updated test to check it.

kkufova:
@gitgabrio, could you please squash the commits?

kkufova:
@manstis, could you please merge the PR?

manstis:
@kkufova Do we need @jomarko to approve?

jomarko:
@manstis no, my comments were incorporated, I am fine with merging.

PR #903, opened 2018-07-30
Untrack jacoco file

jomarko:


jomarko:
@Rikkola @manstis please have a look, we add that file by #898

jomarko:
@manstis please merge when you have a moment.

PR #904, opened 2018-08-06
Drools 2780

gitgabrio:
Implemented a two-tabbed RightPanel. Currently tabs showing hardcoded text.
The RightPanel may be closed with its own "x" button or may be opened/closed with a specifi "Show/Hide" button on the editor toolbar.
The panel automatically shows up when creating/opening a scenariosimulation asset.
The panel automatically close when the scenariosimulation asset closes or goes in the background (e.g. when clicking on the bradcrumbs items)

To be viewed by @jsoltes, @manstis , @Rikkola and @kkufova.


kiereleaseuser3:

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelViewImpl.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @EventHandler("editorTab")
+    public void onEditorTabClick(ClickEvent event) {
+        showTab(editorTab, editorTabContent);
+        hideTab(cheatSheetTab, cheatSheetTabContent);
+    }
+
+    @EventHandler("cheatSheetTab")
+    public void onCheatSheetTabClick(ClickEvent event) {
```
Rikkola: This should call a method in Presenter and that should call showTab/hideTab based on the actions. You can also remove the test that targets this ViewImpl and test the Presenter instead.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -174,6 +220,13 @@ protected void loadContent() {
             baseView.hideBusyIndicator();
             view.setContent(model.getHeadersMap(), model.getRowsMap());
             createOriginalHash(model.hashCode());
+//            if (PlaceStatus.CLOSE.equals(placeManager.getStatus(RightPanelPresenter.IDENTIFIER))) {
+//                placeManager.goTo(RightPanelPresenter.IDENTIFIER);
+//            }
```
Rikkola: Commented code can be removed.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelView.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
```
Rikkola: 2018

Rikkola:
@kkufova @gitgabrio We needs docks for editors if we want to use docks. Any editor that uses docks at the moment is broken outside of Library.

gitgabrio:
There has been some exchange here https://issues.jboss.org/browse/DROOLS-2730
At the end of the day, there are a couple of remarks from Eder about the fact that somehow Docks should not be bound to the editor shown (as for resource navigation, for example) and that in some editor/component docks are not used as they are meant to. This, together with the documentation I found on Uberfire, makes me thinking that for the right panel they are the wrong choice. I do not know why others have used docks in a different way/scope for what they are meant, but surely whenever you use a framework in a different way as it was designed to work, sooner or later issues will come.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelViewImpl.java
gitgabrio: Hi @Rikkola - 
maybe you're right. But if those fields/methods are private (as they should) how could we test that the different tab are actually shown/hidden depending on the fired event ?  EventHandler is managed by framework (errai) but what we do inside it is done by us....

Rikkola: @gitgabrio Here is what I think it should look like. Made a gist out of it. Perhaps easier to read than a comment.

https://gist.github.com/Rikkola/fa097bb4097536ca0cd17c7adaa213d2

gitgabrio:
About the button user friendly: immediately to the left of show/hide right panel there is a show/hide alert button that works exactly the same way - I've just followed the logic. It should not be too hard to implement a dynamic show/close arrow on the middle of the left border of the scenario simulation screen, not sure if it would be more user friendly to have to do different actions for the same result (hide/show a panel). Anyway, I also have the impression that we could spend years talking about this with UXP experts 😄  , so maybe it is not too bad to keep it as it is for the moment being, and eventually change it later - in pure "agile" way.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelViewImpl.java
Rikkola: So each action gets passed back to Presenter so we can test it and ViewImpl is kept as stupid as possible. It might even make sense to add hideCheatSheetTabTab and hideEditorTab methods into the view. So you can test they are hidden.

Rikkola: @gitgabrio Ok I updated the gist with the hide methods.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
gitgabrio: Fixed with
[DROOLS-2780] Refactored RightPanelView/RightPanel interaction. Removing access modifier for testing purpose. Renaming mocked instances.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelView.java
gitgabrio: Fixed with
[DROOLS-2780] Refactored RightPanelView/RightPanel interaction. Removing access modifier for testing purpose. Renaming mocked instances.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelViewImpl.java
gitgabrio: Fixed with
[DROOLS-2780] Refactored RightPanelView/RightPanel interaction. Removing access modifier for testing purpose. Renaming mocked instances.

kkufova:
**RE button**:

@gitgabrio, @Rikkola, I don't know what would be the best approach, maybe we could ask Liz to give us her opinion? I see a difference between the "alerts button" and our "panel button"---the alerts are used on multiple places within Business Central, while I haven't seen such button for side panel anywhere else (but I might be wrong). That's why I'm suggesting an approach that would be unified as much as possible with the rest of the editors and perspectives. Should I create a Jira for Liz?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.java
```diff
@@ -45,4 +45,14 @@
     String remove();
 
     String invalidDefinitionDisabled();
+
+    String ShowRightPanel();
```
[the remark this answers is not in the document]
gitgabrio: Fixed with last push 
DROOLS-2780 Fixed merge

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelView.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.gwt.user.client.ui.IsWidget;
+import org.uberfire.client.mvp.HasPresenter;
+
+public interface RightPanelView
+        extends IsWidget,
+                HasPresenter<RightPanelView.Presenter> {
+
+    void showEditorTab();
+
+    void hideCheatSheetTabTab();
```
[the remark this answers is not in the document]
gitgabrio: Doesn't sounds funny? 😄 
Anyway, fixed with:
DROOLS-2780 Fixed merge

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelViewImpl.java
```diff
@@ -0,0 +1,102 @@
...
+        presenter.onCheatSheetTabActivated();
+    }
+
+    @Override
+    public void showEditorTab() {
+        showTab(editorTab, editorTabContent);
+    }
+
+    @Override
+    public void hideCheatSheetTabTab() {
```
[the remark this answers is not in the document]
gitgabrio: Fixed with
DROOLS-2780 Fixed merge

Rikkola:
ok to test

Am I an admin? :)

Rikkola:
jenkins please test this

jsoltes:
Also @danielezonca how can you merge without full downstream build?

kkufova:
@jsoltes, I would definitely leave such decision on the UX team; there are a few different opinions and many different solutions, so somebody with more experience should resolve this. And we can make the decision and implement the changes in the next sprint, as it does not affect the functionality of the panel.

I created an issue for Liz: [DROOLS-2861](https://issues.jboss.org/browse/DROOLS-2861).

Rikkola:
Maybe for now we can remove the buttons and trust on the automation of the editor and on the fact that it opens and closes with the editor. We do this with dtables and verification for example. And then we can add a ticket for solving how to manage open and close in trouble situations. The ideal situation is to use docks, but at this point it would mean the editor is not usable in embed containers or anywhere else where the perspective is not setup for docks. This again comes down to if the docks get fixed or not.

@jsoltes Do we need downstream builds for merges?

jsoltes:
@kkufova I didn't know about the DROOLS-2861. I think we can agree that this pr can be merged, so that we don't block the other prs, but the feature (user story) cannot be put to done until the design is discussed (I set the design jira to blocker). In the future the design should be clear to everybody before eh implementation is started so that we don't waste time on redesigns.

@Rikkola In forms we used to run full downstream build for everything, but I mean maybe we don't have to if no other repos are affected.

manstis:
Waiting for @jsoltes approval before merging.

Rikkola:
@jsoltes Yep this is only visible to drools-wb. The other repositories do not depend on these modules yet.

PR #905, opened 2018-08-06
[DROOLS-2693] Scenario data model persistance

danielezonca:
Created a data model to contain simulation descriptor and scenario data. Added simple API to provide access by index.
Next step will be to bind table content with this data model (DROOLS-2694)

@Rikkola @kkufova @gitgabrio @manstis


danielezonca:
jenkins retest this please

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/Scenario.java
```diff
@@ -0,0 +1,88 @@
...
+    public FactMappingValue addMappingValue(FactIdentifier factIdentifier, ExpressionIdentifier expressionIdentifier, String value) {
+        String factName = factIdentifier.getName();
+        if (getFactMappingValue(factIdentifier, expressionIdentifier).isPresent()) {
+            throw new IllegalArgumentException(
+                    new StringBuilder().append("A fact value for expression '").append(expressionIdentifier.getName())
+                            .append("' and fact '").append(factName).append("' already exist").toString());
+        }
+        FactMappingValue factMappingValue = new FactMappingValue(factIdentifier, expressionIdentifier, value);
+        factMappingValues.add(factMappingValue);
+        return factMappingValue;
```
Rikkola: Should be useful to test the normal add and the situation where the exception happens.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/Scenario.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public Optional<FactMappingValue> getFactMappingValue(FactIdentifier factIdentifier, ExpressionIdentifier expressionIdentifier) {
+        return factMappingValues.stream().filter(e -> e.getFactIdentifier().getName().equalsIgnoreCase(factIdentifier.getName()) &&
+                e.getExpressionIdentifier().equals(expressionIdentifier)).findFirst();
```
Rikkola: Also the searches.

I see there are few more "meaty" methods in this and few other @Portable classes. Anything with a bit of logic could use at least a simple test.

Rikkola:
@jomarko @danielezonca The StringBuilders do not help in the client side ( for speed at least, for readability yes. Probably not why there are two approaches, but thought I'll bring this up an sound like a nerd that knows something about something.

@danielezonca The code looks really good. Just few tests that could be added.

danielezonca:
@jomarko 
There is no real reason why I used two different approaches for string concatenation, I am lazy :-P
Now I moved all of them to `StringBuilder`

@Rikkola 
Added a couple of tests to cover `Scenario` and `SimulationDescriptor` methods with exception. Let me know if you think I should cover other methods otherwise I think we can wait @kkufova and/or @jsoltes reivew and then ask to merge

manstis:
@danielezonca Does this PR need any others merged at the same time; or can it be merged on its own?

danielezonca:
@manstis 
This PR can be merged by its own.
If you like follow the "order" you can merge #904 before (also @jsoltes approved it now) and then this or the opposite as you prefer :)

PR #906, opened 2018-08-07
DROOLS-2785 : Create run scenario test pipeline

Rikkola:
https://issues.jboss.org/browse/DROOLS-2785

Bundle:
https://github.com/kiegroup/appformer/pull/430
https://github.com/kiegroup/drools-wb/pull/906


Rikkola:
@kkufova @danielezonca @gitgabrio Here are two PRs that could use your attention.

Rikkola:
I think it better to merge this after https://github.com/kiegroup/drools-wb/pull/904, but putting it out here for reviews and test runs.

kkufova:
@Rikkola, the build failed because of these, could you please take a look?

> `[ERROR] src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/ScenarioSimulationViewProvider.java:[18,8] (imports) UnusedImports: Unused import - org.drools.workbench.screens.scenariosimulation.client.editor.ScenarioSimulationView.`
`[ERROR] src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/ScenarioSimulationViewProvider.java:[19,8] (imports) UnusedImports: Unused import - org.drools.workbench.screens.scenariosimulation.client.editor.ScenarioSimulationViewImpl.`

Rikkola:
Failures and merge commits fixed.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioRunnerServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public void runAllTests(final String identifier,
+                            final Path path) {
+
+        defaultTestResultMessageEvent.fire(
+                new TestResultMessage(
+                        identifier,
+                        1,
```
kkufova: Can we use two variables instead of just values (here and in the following methods)?

danielezonca: Current test result is mocked because the runner has not been implemented yet. As soon as the runner will be implemented, all these fake numbers will be replaced by the real ones.
That's also why you see `1 test(s) ran in 0 minutes 0 seconds.`: if you change here the number to 0 or to 10 you can show whatever you want

gitgabrio:
Fine for me 👍

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioRunnerServiceImpl.java
Rikkola: Yes. This is the reason they are as they are. Giving some result to the UI. We should try to keep the number of magic numbers on minimum and use static variables for numbers like this, but this to me is the cleanest mock we can do without investing too much on a mock.

ederign:
@Rikkola can you please fix the conflicts?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioRunnerServiceImpl.java
kkufova: @Rikkola, @danielezonca, alright, thanks for explaining!

PR #909, opened 2018-08-09
[DROOLS-2694] Bind real data model to Scenario Simulation asset

danielezonca:
Replaced mocked (map-based) data model behind simulation table with the real one.

@kkufova @gitgabrio @Rikkola @manstis


danielezonca:
jenkins ok to test

Rikkola:
@manstis Evening, this can go in.

PR #911, opened 2018-08-15
DROOLS-2777: Guided Decision Table is changing date field value based on the timezone

karreiro:
See:
https://issues.jboss.org/browse/DROOLS-2777

---

Demo:
- First, we need to set the `user.timezone` property:

- ..and then, we can change the timezone to **"Europe/Vilnius"** in the server (`systemsetup -settimezone Europe/Vilnius`):

- ..or even change to **"Europe/Prague"** (`systemsetup -settimezone Europe/Prague`):

..the client side is always using the timezone from the server (from `user.timezone`), so the displayed value will always be consistent.

cc @etirelli 

---

Part of an ensemble:
- https://github.com/kiegroup/kie-wb-common/pull/2038
- https://github.com/kiegroup/drools-wb/pull/911

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/DateUiColumn.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/DateUiColumn.java
@@ -20,20 +20,17 @@ import java.util.List;
 import java.util.function.Consumer;
 
 import com.ait.lienzo.client.core.shape.Text;
-import com.google.gwt.i18n.client.DateTimeFormat;
 import org.drools.workbench.screens.guided.dtable.client.widget.table.GuidedDecisionTablePresenter;
 import org.drools.workbench.screens.guided.dtable.client.widget.table.columns.dom.datepicker.DatePickerDOMElement;
 import org.drools.workbench.screens.guided.dtable.client.widget.table.columns.dom.datepicker.DatePickerSingletonDOMElementFactory;
-import org.kie.workbench.common.services.shared.preferences.ApplicationPreferences;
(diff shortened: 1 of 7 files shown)


manstis:
@karreiro Looks good to me.

Just to clarify; the ```user.timezone``` System Property needs to be set to the time-zone that all dates in the Guided Decision Table editor will be represented with no matter what locale the client/browser is actually using. I wonder whether it'd be worth falling back to ```TimeZone.getDefault()``` if the System Property is not set (which should be the servers time-zone).

manstis:
@karreiro ... and, I wonder, what about all the other editors that use ```Date``` :-(

karreiro:
@manstis Thank you for your comment. I just amended the `kie-wb-common` PR, to consider the `TimeZone.getDefault()` in the offset property, when the `user.timezone` is not defined.

Regarding other places that use Date. For the `6.5.x`, I don't think that we should touch anything unless some important/specific scenario appears. For the `master` branch, I think that we can raise another JIRA to refactor all places (editors or not) that are using Date (since the DROOLS-2777 explicit refers to the GDT editor). wdyt? :-)

manstis:
@karreiro Raising new JIRA sounds good to me :-)

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/columns/DateUiColumn.java
```diff
@@ -33,37 +34,40 @@
...
+              width,
+              isResizable,
+              isVisible,
+              access,
+              factory);
+    }
+
+    static String format(final Date value) {
+        return dateTimeFormat.format(value, TimeZoneUtils.getTimeZone());
     }
```
jomarko: I think we can reuse `GWTDateConverter`

```java
static String format(final Date value) {
    return GWTDateConverter.getInstance().format(value);
}
```

karreiro:
@jomarko Changes applied.

PR #914, opened 2018-08-21
PLANNER-1159: Improve categories

yurloc:
- Move Test Scenario resource type to Decision category.
- ~Move WorkItem resource type to Process category.~

Related PRs:
kiegroup/appformer/pull/438
kiegroup/kie-wb-common/pull/2048
kiegroup/drools-wb/pull/914
kiegroup/optaplanner-wb/pull/296

## Asset List
kie workbench - google chrome_070
kie workbench - google chrome_069

## Add Asset
screenshot from 2018-08-20 16-53-16
screenshot from 2018-08-20 16-53-07

### DIFF
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-api/pom.xml
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-api/pom.xml
@@ -32,6 +32,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.uberfire</groupId>
+      <artifactId>uberfire-project-api</artifactId>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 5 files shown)


manstis:
@yurloc I'd agree with "Work Items" being considered "Process" however:-
1) WorkItemHandlers can also be used from the Guided Decision Table designer
2) The original classifications came from, IIRC, Ivo Beck and hence we should probably defer to him.

@adrielparedes Might recall where the original classifications came from and hence who we should probably ask for approval/agreement (rather than engineering making an executive decision, only to be revoked when the original requester hears of the change).

yurloc:
@manstis Good point. I didn't know that WIDs can be used in decision tables. In that case it would make sense to have the ability to assign multiple categories to a resource type but this is becoming way out of scope of the original request. Let's hear @ibek's opinion but I'm ready for reverting this change.

PR #918, opened 2018-08-28
refactor the junit result panel to use the new errai

jsoltes:
This is first part of DROOLS-2903 Update test result panel with new UX/UI.

@Rikkola can you please review?


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/reporting/TestRunnerReportingViewImpl.java
```diff
@@ -20,120 +20,130 @@
...
+        resultStats.textContent = TestScenarioConstants.INSTANCE.XTestsRanInYMinutesZSeconds(runCount,
+                                                                                             minutesFormat.format(date),
+                                                                                             secondsFormat.format(date));
+    }
 
-        stats.setText( TestScenarioConstants.INSTANCE.XTestsRanInYMinutesZSeconds( runCount, minutesFormat.format( date ), secondsFormat.format( date ) ) );
+    @Override
+    public Widget asWidget() {
+        return ElementWrapperWidget.getWidget(resultPanel);
     }
```
jomarko: I think our `TestRunnerReportingView` interface should extend `org.uberfire.client.mvp.UberElemental` and then we can omit this `asWidget()`

See [this class](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/popups/validation/ValidationPopupView.java) for example.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/reporting/TestRunnerReportingViewImpl.html
```diff
@@ -0,0 +1,33 @@
...
+  -->
+
+<div data-field="resultPanel" class="container-fluid">
+    <div class="row">
+        <span data-field="resultTitle"></span>
+    </div>
+
+    <div class="row" style="text-align: center;">
+        <div class="col-md-12">
+            <span data-field="resultStats"></span>
```
jhrcek: You're declaring `resultStats` as span in the template, but your view has `HTMLDivElement resultStats`. This will lead to errai warnings in the browser console. I'd recommend using div in the template instead.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/reporting/TestRunnerReportingViewImpl.java
Rikkola: This is true. The reason for this is that the Presenter should not have GWT Widgets or components in it and we hide them behind interfaces like IsWidget or in the new parts of the code we should use UberElemental. The presenter should be testable without the GWT mockito runner.

jsoltes: @jomarko on the other hand, you have to override getElement and init methods. But if it is the "new" way, I can do this.

jsoltes:
@jomarko making the changes you suggested actually broke the component and also it made the code much longer, so I decided to revert it.

jomarko:
@jsoltes Sorry for confusion. I didn't want to say that just change of the interface will do the trick. Probably that would require changes at more places.

jsoltes:
@manstis can you please merge?

PR #923, opened 2018-09-06
[AF-1528] Servlet ContextListener deployer 

desmax74:
Enabled the Servlet listener to deploy jars during the startup of the webapp


desmax74:
https://github.com/kiegroup/appformer/pull/450

@porcelli @jhrcek @domhanak

porcelli:
Servlet container automatically picks the `@WebListener` annotation, you don't need declare it in web.xml

desmax74:
@porcelli we are using a web.xml with servlet 2.5

porcelli:
its likely just a left over from an old upgrade, we should just update the servlet reference, but quite sure WF/EAP will pickup based on annotation

PR #926, opened 2018-09-10
Fix for ECJ - GWT conflict

Rikkola:


Rikkola:
Testing the hosted mode

drools-wb-webapp/pom.xml
```diff
@@ -1240,6 +1240,22 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
manstis: Is this confirmed from CLI (```mvn gwt:run```) and IDEA?

I had to add it as the top-most dependency (or probably just before any other dependency that has ```drools-compiler``` as a transitive dependency). You've added last.

Rikkola:
@manstis Had to move it up. Idea SDM took a while to get working due to me running out of disk space. Happens everyday at some point due to Idea not cleaning temps.

Rikkola:
So I tested Idea and hoping @kkufova could test CLI. Since that doesn't normally work for me.

gitgabrio:
@manstis @porcelli @Rikkola 
I think the proper place to put this modification should be the drools-wb/pom.xml or higher, if possible.
All other projects (screens/webapp/etc) extend it, so this modification should be inherited by any of them.
As I already said elsewhere, I have the idea that having only the drools-wb-webapp to test the screens is a mistake from an architectural point of view : the dependency mess and the unwanted cross-dependency between modules is the demonstration.
Also, always from a good-design point of  view, IMHO these modification should be available to all GWT projects, not only the ones inside drools

manstis:
@gitgabrio Can you try placing the fix in ```drools-wb pom.xml```?

If you can identify a higher place to put it that'll fix all webapps then we'd be more than happy to do that too. @Rikkola, @wmedvede, @porcelli and I just happened to be the first to try to fix our development environments. Feel free to get involved!

As for your comment about ```drools-wb-webapp``` I'd disagree. Having a _lighter-weight_ (lighter than ```kie-wb```) workbench, not just to test drools screens but also their integration in a (near complete) workbench has proved to pay dividends in the past. However that said, if you want to cleanse what we have that still provides a platform to test integration with the workbench (project builds, dock integration etc) then I'd happily review your PR :-)

gitgabrio:
@manstis @Rikkola @porcelli 
Currently I've already put that on drools-wb/pom.xml (I would like to have it there - in my machine -  because most of the time I use my lighter showcase and when I need all the assets I use drools-wb-webapp).
My only "concern" is that I do not exactly how to fire a "drools" compilation that might eventually fail if there is a wrong version - beside that my code runs as usual.

ederign:
Thanks @Rikkola

manstis:
@gitgabrio This PR has been merged to expedite the fix for other developers.

Please submit another PR if you can improve upon this solution. All improvements are welcome.

gitgabrio:
@manstis  
@Rikkola I agree with your comment.
Last night I begun a drools-wb  refactoring to address this. I may be wrong, so I have to go through and test it, but I think the main issue is that currently client/gwt modules are not separated from backends one; instead, they are all under "screens", grouped by the asset. All the "best practices" of GWT are somehow against it (GWT documentation, mojo maven plugin, tbroyer plugin).
What I'm doing is
1) create two "sub-module" roots: drools-wb-backend and drools-wb-client
2) put all "client" modules under the latter, and the others under the former
That way, I think it would be much easier to manage the different needs of this two groups of modules.
In the same time, this would allow me to declare all dependencies in on place (drools-wb-client/pom) while now they are duplicated inside all the client poms. Moreover, it will be clearly written out  which "client" depends specifically on other one (that dependency will be declared only in the requiring module).
After that, my plan would be to have drools-wb-webapp simply depending on drools-wb-backends and drools-wb-clients.
If everything works as I hope, the maven "fix" could be put in one place and inherited in all the modules that needs it.

gitgabrio:
@Rikkola 
I thought it was for that reason. I would call it a "maven" approach and overall I would agree. On the other side, in my experience to work well with GWT you have to follow its rules. All the troubles about Intellij plugin, SDM, compilation times, dependencies, etc etc are related exactly to that - and if you think it over they are extremely costly. Of course after a while we all find our "way" to overcome some issues and we get used to things as they are (i.e. we all develop our "black" rituals), but IMHO we are just fighting against the framework - and this is something that I would put in higher priority than "comfort" of having related codes under the same path.
As I side note, I worked in other complex multi-module GWT projects with the client/share/server layout, and all the people working on that where perfectly at easy: it is just a matter of getting used, just like here we are getting used to a lot of different workarounds.

Rikkola:
@gitgabrio is it possible that the split would speed up compile times? In theory the backend becomes a black box for client side, but then again it also still has to be all put together in the webapp module.

gitgabrio:
@Rikkola 
Well, now we are in the "hypothesis" world. 
My idea is that we should have a modular test environment - i.e. it should be easy to create specific "showcases" for each screen - I would like to make an archetype that simply create a "showcase" for a specific "screen"; showcases made that way could be dropped and created easily, so no need to version/maintain them. In that situation, developing time would be greatly reduced, because as developer I would not have the hassle of managing tons of unneeded dependencies/inherited modules.
drools-wb-webapp would then be just an "integration" environment, that could be also skipped by developers - so build time would not be an issue (this is exactly what I'm currently doing).

Anyway, I am not even sure that my idea would actually work.
My approach is simple: 
I see something that does not works as I know it should; 
I see that something has been done differently on what I would have done; 
I try to see if re-doing things as I would have done may fix the issues.
I have to actually test it to verify everything - I think it should work

Rikkola:
@gitgabrio I think we could generate the showcases. Just tick the screens you need and you are good to go. This would remove the need for most of the showcases and speed up the build I think, but based on these ECJ fixes and issues with the same fix not working for every showcase I'm not sure how easy something like this would be to do.

gitgabrio:
@Rikkola 
Well, we could have different issues, some of them manageable only at "showcase" level, others more up. In that situation, I just modified the pom of my personal showcase - maybe 30 seconds to copy and paste ? - and it worked again.
Anyway, your last comments makes me ask: is this not due to the fact that every module/showcase is done somehow in a different way? What if (not saying is it possible, do not know, but just as hypotesis) all the showcases would extend a common parent:  would have not possible to make the fix in only one point? My university teacher always repeated: "Structure is function" (meaning the final behaviour depends on the structure you use): well, I still think that a lot of issues are generally consequent to a wrong structure - but that unfortunately instead of fixing the structure we add patches, increasing the overall entropy and the potentiality for further issues.

porcelli:
@gitgabrio based on our call, we're waiting for a working proposal from you, so everyone will be able to evaluate and point pros/cons... as I said in our call: code always win.

Rikkola:
@gitgabrio Yes the fix was an easy copy paste for most parts, but some kie-wb-common showcases needed more, but perhaps they are indeed doing something differently. 

Another issue with these workspaces is for example the fact that css-styles ( and probably many other things ) are not the same for all. For example some buttons looks different in drools-wb-showcase and in the production wars. So a shared base for all would be needed.

And still even when the workspaces are different, they still are really redundant with copy pasted code.

But hoping you have figured out a better way. For anyone following, the speed differences I tested with drools-wb-showcase and @gitgabrio's custom showcase for the Scenario Simulation editor were:
* drools-wb-showcase 4 minutes to start SDM for the first time
* scenario-simulation-showcase 2 minutes 30 seconds

And this is on my own setup that is way faster than the company standard laptop. If I used the official work laptop drools-wb-showcase would be closer to 10 minutes.

PR #927, opened 2018-09-11
[DROOLS-2982] Create backend methods to pair with "duplicate row"

danielezonca:
Added methods to clone Scenario and create Scenario at a specific index

@jsoltes @Rikkola


danielezonca:
@manstis can you merge (squash) also this PR?

Thanks :)

PR #928, opened 2018-09-12
Drools 2904 - Show menu to add row and column

gitgabrio:
Scope of this review is to show a contextual menu (right-click) to allow insert/deleting of columns/rows.
Contextual menus will have different elements depending on the element clicked.

Please verify
1) correct behaviour
2) adherence with UX proposal; if not met, let's talk if modifications are needed


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/Simulation.java
```diff
@@ -49,11 +61,31 @@ public Scenario getScenarioByIndex(int index) {
...
         return scenario;
     }
 
+    public Scenario cloneScenario(int sourceIndex, int targetIndex) {
+        if (sourceIndex < 0 || sourceIndex >= scenarios.size()) {
+            throw new IllegalArgumentException(new StringBuilder().append("SourceIndex out of range ").append(sourceIndex).toString());
+        }
+        if (targetIndex < 0 || targetIndex > scenarios.size()) {
+            throw new IllegalArgumentException(new StringBuilder().append("TargetIndex out of range ").append(targetIndex).toString());
+        }
```
jomarko: Do we really need these index checks in this class? I had a look on [javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#add-int-E-) and it seems to me as not needed.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -105,30 +109,28 @@ public ScenarioSimulationEditorPresenter() {
...
         view.init(this);
 
-        rightPanelRequest = new DefaultPlaceRequest(RightPanelPresenter.IDENTIFIER);
-        rightPanelRequest.addParameter("ScenarioSimulationEditorPresenter", this.toString());
-
-        rightPanelMenuItem.init(rightPanelRequest);
+//        rightPanelRequest = new DefaultPlaceRequest(RightPanelPresenter.IDENTIFIER);
+//        rightPanelRequest.addParameter("ScenarioSimulationEditorPresenter", this.toString());
+//
+//        rightPanelMenuItem.init(rightPanelRequest);
```
jomarko: If not needed, please remove.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/menu/AbstractHeaderGroupMenuPresenter.java
```diff
@@ -0,0 +1,47 @@
...
+    PrependColumnEvent prependColumnEvent;
+
+    /**
+     * This method set <b>group-specific</b> menu items and common <b>SCENARIO</b> menu items
+     */
+    public void initMenu() {
+        addMenuItem(HEADERCONTEXTMENU_GROUP, HEADERCONTEXTMENU_LABEL, HEADERCONTEXTMENU_I18N);
+        addExecutableMenuItem(HEADERCONTEXTMENU_INSERT_COLUMN_LEFT, constants.insertColumnLeft(), "insertColumnLeft", prependColumnEvent);
+        addExecutableMenuItem(HEADERCONTEXTMENU_INSERT_COLUMN_RIGHT, constants.insertColumnRight(), "insertColumnRight", appendColumnEvent);
+        //        addExecutableMenuItem(HEADERCONTEXTMENU_DELETE_COLUMN, constants.deleteNewColumn(), "deleteNewColumn", () -> GWT.log(HEADERCONTEXTMENU_DELETE_COLUMN));
```
jomarko: Probably save to remove

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
jomarko: There are more similar occurrences, I didn't comment all.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/Simulation.java
gitgabrio: These code comes from https://github.com/kiegroup/drools-wb/pull/927 - that has been approved. Anyway, ask @danielezonca about it.

danielezonca: I created these checks to avoid a general IOOB exception so the user/developer will see which index is wrong at least

jomarko: Thanks for answers, will not dig deeper into it. Leaving review for scenario experts.

gitgabrio:
@kkufova @danielezonca 
About "insert row above/below" on green header: completely agree: they are just labels and I need the exact ones to put there; when clicked on header, they actually mean "insert as first" or "insert as last", while when clicking on a grid row, the mean "insert above or below the clicked row"
About "insert column left/right" on GIVEN/HEADER: the behavour is different than in GIVEN-1/EXPECTED-1: in GIVEN/EXPECTED -> insert the column to the leftmost or rightmost of the **group**, while in GIVEN-1/EXPECTED-1 it insert the column to the left or right of the clicked one.
About "2. Secondly": yes - that's a bug
About "3. The duplicate row": I saw your video a couple of time, part of the behaviour is expected  - the collapsing effect is absolutely misleading. I will look about the duplication of the first row

kkufova:
@gitgabrio, thank you for the answers!

> About "insert column left/right" on GIVEN/HEADER: the behavour is different than in GIVEN-1/EXPECTED-1: in GIVEN/EXPECTED -> insert the column to the leftmost or rightmost of the group, while in GIVEN-1/EXPECTED-1 it insert the column to the left or right of the clicked one.

In that case, could we please rename the options for the **GIVEN** and **EXPECTED** to something like _insert leftmost column_ and _insert rightmost column_ to be more precise?

gitgabrio:
@kkufova 
Yup - I'll put that strings.

Rikkola:
@danielezonca @gitgabrio I don't think the distribution wars have GWT.log setup on the client side, so it doesn't matter if we use it, but I need to check if just using it gives us trouble.

manstis:
@gitgabrio Please be aware @danielezonca is moving Stunner's generic "Command" infrastructure (execute/undo etc etc) to a more re-usable module. You should strive to re-use the existing infrastructure rather re-invent the wheel. See https://issues.jboss.org/browse/DROOLS-2898 and feel free to talk to @danielezonca about its usage.

manstis:
@gitgabrio @danielezonca Ah, I now see https://issues.jboss.org/browse/DROOLS-2911 :-)

danielezonca:
@manstis 
Yes we decided to start with a simple local command executor because the discussion about Stunner command were still in progress when he started this implementation and then migrate (essentially just replace the executor and the interfaces) in a second step :)

gitgabrio:
Fixed comments. Please verify

danielezonca:
@manstis 

PR #930, opened 2018-09-13
DROOLS-2973 : Revisit the XML and XStream format and setup for Scenario Simulation

Rikkola:
https://issues.jboss.org/browse/DROOLS-2973

@kkufova @danielezonca This cleans up the packages from the XML. Let me know if you want a different format or additional tests, not sure what tests make sense here.


Rikkola:
@danielezonca Ok. The version number is added.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationXMLPersistence.java
```diff
@@ -28,17 +40,34 @@
 
     private ScenarioSimulationXMLPersistence() {
         xt = XStreamUtils.createTrustingXStream(new DomDriver());
+
+        xt.autodetectAnnotations(true);
+
+        xt.alias("ExpressionElement", ExpressionElement.class);
```
jomarko: Just idea, couldn't we use 
```java
xt.alias(ExpressionElement.class.getSimpleNmae(), ExpressionElement.class);
```

danielezonca: In general I agree with you but in this specific use case I suggest to use the string to avoid that a refactor will break existing scenario loading: the IDE will update class name without any warn but then you will be not able to reload an existing file

jomarko: thanks

PR #935, opened 2018-09-17
DROOLS-2940 : It is not possible to run all scenarios in the old test scenarios.

Rikkola:
https://issues.jboss.org/browse/DROOLS-2940

Bundle:
https://github.com/kiegroup/appformer/pull/463
https://github.com/kiegroup/drools-wb/pull/935


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioEditorPresenter.java
```diff
@@ -33,7 +33,7 @@
 import org.drools.workbench.screens.testscenario.model.TestScenarioResult;
 import org.drools.workbench.screens.testscenario.service.ScenarioTestEditorService;
 import org.guvnor.common.services.shared.metadata.model.Metadata;
-import org.guvnor.common.services.shared.test.TestService;
```
jomarko: Is `org.guvnor.common.services.shared.test.TestService` used somewhere else in `kiegroup` codebase?

Rikkola:
@manstis Hello, if you could be kind enough to merge these thank you.

PR #937, opened 2018-09-19
DROOLS-2994 Rename Scenario Simulation to Test Scenarios (Preview) and update icon

kkufova:
According to [this document](https://docs.google.com/presentation/d/1uBgHnBOzmsXvRUs7fFToLsjVLs3ag6lCnc6XFVUk9Ao/edit?ts=5b96322e#slide=id.g3efa8f3c6c_7_130), the official name of the Scenario Simulation asset is **Test Scenarios (Preview)**. I renamed the asset in the UI and updated its icon (based on a UX decision, the icon should be the same as for the old Test Scenarios). I followed the "naming convention" of the old Test Scenarios.

However, I am not sure I managed to update all the right places. For example, in all `pom.xml` files, I left the name Scenario Simulation (all components in the code are called Scenario Simulation anyway), as it does not really make sense to me to rename it there.

@Rikkola, @danielezonca, could you please review?


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/BaseExpressionGridTheme.java
```diff
@@ -67,7 +67,7 @@
 
     @Override
     public String getName() {
-        return "Scenario Simulation Editor";
+        return "Test Scenarios (Preview)";
```
Rikkola: Since this is a client side name is it visible to user? If so, then it would also need to be i18nized.

danielezonca: I think this value will be visible only if we want to support multiple themes with the possibility for the user to switch. For now I think it is enough to change it here without i18n

kkufova: @Rikkola, thank you for noticing, I don't think this is visible at the moment (as @danielezonca wrote). From the code:

```
/**
 * Definition of themes used by a render for the pluggable rendering mechanism.
 */
public interface GridRendererTheme {

    /**
     * Returns a display name for the theme.
     * @return A name for the theme
     */
    String getName();
...
```

manstis:
@Rikkola @kkufova @danielezonca Theme "names" are not shown to the User (in the workbench at least). It is therefore, IMO, OK to not i18n it.

kkufova:
@jomarko, can you take a look as well? Juraj's not available at the moment for a review.

PR #938, opened 2018-09-19
Drools 2902 - Link Scenario grid header cells with expression builder component

gitgabrio:
Scope of these PR is to provide ability to modify a selected column' type with one choosen from firght panel.
Requiremnts (to verify)
1) clicking on a 2nd level header should select the colum and *eventually* deselect a previously selected one
2) clicking outside the grid should deselect the column
3) Editor tab in right panel should be disabled by default
4) the editor tab should be enabled only when a column is selected
5) if there is no selected column, editor tab should be disabled
6) when double click on a property of a type inside the  editor tab, the header of the selected column should show the full "path" for this property
7) after saving/closing/and reopen the scenario, the modification should be present


gitgabrio:
@kkufova 

About the bugs:
I've fixed _clicking outside the grid should deselect the column_

After that, I tried to recreate the "_remove column_" one, but I could not.

About UX:
_when double click on a property of a type_.. : underlying/changing color sounds great for me; I would stick to double click, though (it is easy to fire a click by mistake); anyway, I would prefer to have the feedback from the actual user (next tuesday @danielezonca will show it)
 
_double click the second-level header_ : there is a pending PR from @danielezonca about these editing, and it should have fixed that

gitgabrio:
@kkufova 
Fixed with [DROOLS-2902] Fixing column selection/deselection on lost focus/delete column. CHanged property style inside RightPanel

danielezonca:
@manstis 
Build is green now :)
Squash and merge is fine

Thanks

PR #939, opened 2018-09-19
[DROOLS-3001] Add support for simple values mapping, add support to not equal operator

danielezonca:
This PR contains
- Fix for the exception found by Toni today when the user runs a scenario with column mapped to entire DO (for instance an empty scenario contains only 6 default String columns)
- Add `NOT_EQUALS` operator (I forgot to add it before from the playground project)
- Minor refactor of raw values parsing

@Rikkola @kkufova


danielezonca:
jenkins retest this please

gitgabrio:
@danielezonca 
Inside ScenarioRunnerHelper
primitive type are casted to boxed equivalents.
I do not know if it could be an issue; anyway it is possible to use _isAssignableFrom_ to primitive classes:

_Class<?> clazz = int.class;
clazz.isAssignableFrom(int.class);_

With that in place, it would be possible to return the exact same type of original variable (not the boxed one):

 _Object value = 4;
 int toReturn = ((Integer) value).intValue();_

danielezonca:
@gitgabrio 
I updated primitive handling to cover also null values and limit the usages of boxing/unboxing. I think for now it is enough because we still need to define how manage collections so I expect quite big refactor of these methods in the future.

The same for operators symbols, I would like to keep multiple choices for now until we will define the expression language that we want to support (FEEL, MVEL, etc)

Let me know if you have other questions

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/FactMappingValue.java
```diff
@@ -72,6 +78,34 @@ FactMappingValue cloneFactMappingValue() {
...
+        FactMappingValueOperator operator = FactMappingValueOperator.findOperator(value);
+        Optional<String> first = operator.getSymbols().stream().filter(value::startsWith).findFirst();
+        if (first.isPresent()) {
+            symbolToRemove = first.get();
+        }
+
+        int index = value.indexOf(symbolToRemove);
+
+        return value.substring(index + symbolToRemove.length()).trim();
+    }
```
jomarko: Not sure how often `symbolToRemove` is empty, but if in most cases, I think we should implement like this:
```java
public static Object cleanValue(Object rawValue) {
        if (!(rawValue instanceof String)) {
            return rawValue;
        }
        String value = ((String) rawValue).trim();

        FactMappingValueOperator operator = FactMappingValueOperator.findOperator(value);
        Optional<String> first = operator.getSymbols().stream().filter(value::startsWith).findFirst();
        if (first.isPresent()) {
            String symbolToRemove = first.get();
            int index = value.indexOf(symbolToRemove);
            value = value.substring(index + symbolToRemove.length());
        }

        return value.trim();
    }
```

danielezonca: Done, thanks for the suggestion :)
Can you please approve the PR now that you have commented?

danielezonca:
@manstis 

Thanks

PR #941, opened 2018-09-20
[DROOLS-2981] Fix collapsing cell/label default naming

gitgabrio:
@kkufova @danielezonca 
Scope of this PR is to remove the "collapsing cell" behaviour.
Beside that, the second header level label generation has been modified, so that the number appearing to the right is always incremented, regardless of where the column is actually put.
This PR should be quick to check and has a minimal impact, so if everything is ok it could be merged before 2902

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/ScenarioGridColumnRenderer.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/ScenarioGridColumnRenderer.java
@@ -15,8 +15,29 @@
  */
 package org.drools.workbench.screens.scenariosimulation.client.renderers;
 
+import java.util.List;
+import java.util.function.BiFunction;
+
+import org.uberfire.ext.wires.core.grids.client.model.GridColumn;
+import org.uberfire.ext.wires.core.grids.client.widget.context.GridBodyColumnRenderContext;
+import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.columns.impl.ColumnRenderingStrategyFlattened;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.columns.impl.StringColumnRenderer;
+import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.grids.GridRenderer;
+import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.grids.impl.BaseGridRendererHelper;
 
 public class ScenarioGridColumnRenderer extends StringColumnRenderer {
 
+    @Override
+    public List<GridRenderer.RendererCommand> renderColumn(final GridColumn<?> column,
+                                                           final GridBodyColumnRenderContext context,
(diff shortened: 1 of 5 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/ScenarioGridColumnRenderer.java
```diff
@@ -15,8 +15,29 @@
...
 import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.columns.impl.StringColumnRenderer;
+import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.grids.GridRenderer;
+import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.grids.impl.BaseGridRendererHelper;
 
 public class ScenarioGridColumnRenderer extends StringColumnRenderer {
 
-    }
\ No newline at end of file
+    @Override
+    public List<GridRenderer.RendererCommand> renderColumn(final GridColumn<?> column,
```
manstis: Overriding the ```renderColumn(..)``` method works; however you'd probably be better off setting the model data to "not merged" in the first instance: https://github.com/kiegroup/drools-wb/blob/097b08c64547f84bcf5f8a96729d072b27d8f4c3/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java#L56 (use ```super(false)```)

manstis: The default constructor for ```BaseGridData``` sets ```merged``` to true; so you could either have ```ScenarioGridModel()`` call ```super(false)``` or change https://github.com/kiegroup/drools-wb/blob/bb81464fdafb5cebb72157c38a1e5b31430cb603/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/producers/ScenarioGridPanelProducer.java#L48 to use ```false``` in the constructor.

manstis: @gitgabrio I have no problem with your approach; I was just explaining an alternative.

_Some_ uses of the grid allow the User to toggle merged/unmerged state (the Guided Decision Table Editor supports this). The [JavaDocs](https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/GridData.java#L230) may not be the best (but arguably better than most of our code).

gitgabrio: @manstis 
👍Thx!

PR #943, opened 2018-09-24
[DROOLS-3008] It should not be possible to delete the GIVEN/EXPECTED header from a Test Scenario (Preview) table

gitgabrio:
Scope of this PR is to fix column deletion policy:
1) column deletion should be allowed only for GIVEN/EXPECTED columns
2) when deleting the last GIVEN/EXPECTED column, a new autogenerated one should be put in place, so that there will always be at least one GIVEN/EXPECTED column


Rikkola:
Good day. Is there a test for item number 2?

gitgabrio:
@Rikkola 
Add that test with last commit

gitgabrio:
@kkufova 
Fixed "deselection" and pom

gitgabrio:
@manstis 
Could you merge this, please?

gitgabrio:
@manstis 
please hold on this merge

gitgabrio:
@manstis
Could you merge this, please?

PR #944, opened 2018-09-25
[DROOLS-3033] Fixing dependency declaration of drools-verifier-api

gitgabrio:
@Rikkola @manstis @porcelli @danielezonca 
Scope of this PR is to fix the dependency declaration of drool verifier api inside drools-wb-webapp

drools-wb-verifier-api -> drools-verifier-api

Currently, enforcer find "drools-verifier-api" as transitive dependency, while in the pom the drools-wb-verifier-api one is declared.


gitgabrio:
@manstis 
please merge this, thanks!

PR #945, opened 2018-09-26
Drools 3007 Add the leftmost ID column to the Test Scenario (Preview) table

gitgabrio:
@jomarko  @kkufova @danielezonca @Rikkola 
Scope of this PR is to show an "index" column to the leftmost of the grid.
The number on that columns should also reflect the actual position of each row, regardless of inserting/deleting row.

With that PR the following issues should also be fixed:
https://issues.jboss.org/browse/DROOLS-3036
https://issues.jboss.org/browse/DROOLS-3042


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationViewImpl.java
```diff
@@ -76,9 +67,7 @@ public void setContent(Simulation simulation) {
 
     @Override
     public void clear() {
-        if (handlerRegistrations != null) {
-            handlerRegistrations.forEach(HandlerRegistration::removeHandler);
-        }
+        //
```
Rikkola: Sometimes it is better to say nothing at all.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/test/java/org/drools/workbench/screens/scenariosimulation/model/ScenarioTest.java
```diff
@@ -80,6 +80,17 @@ public void getDescriptionTest() {
...
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void getIndexTestFail() {
+        scenario.getIndex();
+    }
+
+    @Test
+    public void getIndexTest() {
+        scenario.addMappingValue(FactIdentifier.INDEX, ExpressionIdentifier.INDEX, 1);
```
danielezonca: Why are you using the "internal" implementation instead of `setIndex` ?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/handlers/ScenarioSimulationGridPanelClickHandler.java
```diff
@@ -113,14 +116,26 @@ public void setEventBus(EventBus eventBus) {
...
+                    scenarioGrid.clearSelections();
+                    if (manageLeftClick(canvasX, canvasY, isShiftKeyDown, isControlKeyDown)) {
+                    } else {
+                        eventBus.fireEvent(new DisableRightPanelEvent());
+                    }
+                }
+                clickReceived.set(0);
+            }
+        };
+        t.schedule(300);
```
danielezonca: Please add a comment to explain the reason of this (= manage single click and double click on the same element)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/ScenarioSimulationModel.java
```diff
@@ -38,22 +38,24 @@ public ScenarioSimulationModel() {
         simulation = new Simulation();
         SimulationDescriptor simulationDescriptor = simulation.getSimulationDescriptor();
 
+        simulationDescriptor.addFactMapping(FactIdentifier.INDEX.getName(), FactIdentifier.INDEX, ExpressionIdentifier.INDEX);
         simulationDescriptor.addFactMapping(FactIdentifier.DESCRIPTION.getName(), FactIdentifier.DESCRIPTION, ExpressionIdentifier.DESCRIPTION);
 
         Scenario scenario = simulation.addScenario();
         int row = simulation.getUnmodifiableScenarios().indexOf(scenario);
-        scenario.setDescription(FactMappingValue.getPlaceHolder(0, 0));
+        scenario.setIndex(1);
```
danielezonca: Haven't we decided to keep index column empty in the model and update it only at UI level?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/Scenario.java
```diff
@@ -106,6 +106,17 @@ public FactMappingValue addOrUpdateMappingValue(FactIdentifier factIdentifier, E
         return factMappingValues.stream().filter(e -> e.getFactIdentifier().equals(factIdentifier)).collect(toList());
     }
 
+    public void setIndex(int index) {
```
danielezonca: We discussed to have index only as UI element so I think we can remove completely `setIndex` and `getIndex` from `Scenario`. WDYT?
If you prefer to keep it I suggest to consider it not like a field but more like a property: `withIndex()` and `hasIndex()` with internal fixed value like `null`

PR #947, opened 2018-09-28
[DROOLS-3044] Update ScenarioGridModel inherited methods to align Simulation model

gitgabrio:

Scope of this pr is to refactor ScenarioGridModel methods names to reflect following contract: "overrided" methods works on both basegrid and scenariosimulation models, "GridOnly" ones works only on basegrid model.

This modification has impact on all the grid rendering, but it should be absolutely transparent for the final user.

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
@@ -23,6 +23,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.stream.IntStream;
 
+import com.google.gwt.core.client.GWT;
 import com.google.gwt.event.shared.EventBus;
 import org.drools.workbench.screens.scenariosimulation.client.events.ScenarioGridReloadEvent;
 import org.drools.workbench.screens.scenariosimulation.client.metadata.ScenarioHeaderMetaData;
@@ -76,18 +77,11 @@ public class ScenarioGridModel extends BaseGridData {
     }
 
     /**
-     * This method <i>append</i> a new column to the grid <b>and</b> to the underlying model
-     */
-    public void appendNewColumn(final GridColumn<?> column) {
-        checkSimulation();
-        commonAddColumn(-1, column);
-    }
-
-    /**
      * This method <i>append</i> a new row to the grid <b>and</b> to the underlying model
      * @param row
      */
-    public void appendNewRow(GridRow row) {
+    @Override
+    public void appendRow(GridRow row) {
         checkSimulation();
         super.appendRow(row);
         int rowIndex = getRowCount() - 1;
@@ -98,17 +92,17 @@ public class ScenarioGridModel extends BaseGridData {
      * This method <i>insert</i> a row to the grid and populate it with values taken from given <code>Scenario</code>
      * @param row
      */
-    public void insertRow(final int rowIndex,
-                          final GridRow row, final Scenario scenario) {
+    public void insertRowGridOnly(final int rowIndex,
+                                  final GridRow row, final Scenario scenario) {
         checkSimulation();
-        insertRow(rowIndex, row);
+        super.insertRow(rowIndex, row);
         scenario.getUnmodifiableFactMappingValues().forEach(value -> {
             FactIdentifier factIdentifier = value.getFactIdentifier();
             ExpressionIdentifier expressionIdentifier = value.getExpressionIdentifier();
             if (value.getRawValue() instanceof String) {
                 String stringValue = (String) value.getRawValue();
                 int columnIndex = simulation.getSimulationDescriptor().getIndexByIdentifier(factIdentifier, expressionIdentifier);
-                setCell(rowIndex, columnIndex, () -> new ScenarioGridCell(new ScenarioGridCellValue(stringValue)));
+                setCellGridOnly(rowIndex, columnIndex, () -> new ScenarioGridCell(new ScenarioGridCellValue(stringValue)));
             } else {
                 throw new UnsupportedOperationException("Only string is supported at the moment");
             }
@@ -119,7 +113,8 @@ public class ScenarioGridModel extends BaseGridData {
      * This method <i>insert</i> a new row to the grid <b>and</b> to the underlying model
      * @param row
      */
-    public void insertNewRow(int rowIndex, GridRow row) {
+    @Override
+    public void insertRow(int rowIndex, GridRow row) {
         checkSimulation();
         super.insertRow(rowIndex, row);
         commonAddRow(rowIndex);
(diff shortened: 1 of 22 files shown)


gitgabrio:
@jomarko 
Fixed merge conflicts

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
```diff
@@ -100,21 +93,21 @@ public void appendNewRow(GridRow row) {
...
-                setCell(rowIndex, columnIndex, () -> new ScenarioGridCell(new ScenarioGridCellValue(stringValue)));
+                setCellGridOnly(rowIndex, columnIndex, () -> new ScenarioGridCell(new ScenarioGridCellValue(stringValue)));
             } else if (value.getRawValue() instanceof Integer) {
                 String stringValue = ((Integer) value.getRawValue()).toString();
                 int columnIndex = simulation.getSimulationDescriptor().getIndexByIdentifier(factIdentifier, expressionIdentifier);
-                setCell(rowIndex, columnIndex, () -> new ScenarioGridCell(new ScenarioGridCellValue(stringValue)));
+                setCellGridOnly(rowIndex, columnIndex, () -> new ScenarioGridCell(new ScenarioGridCellValue(stringValue)));
             } else {
                 throw new UnsupportedOperationException("Only string is supported at the moment");
             }
```
jomarko: I had look on this method and think we could simplify slightly, let me know what do you think:
```java
    public void insertRowGridOnly(final int rowIndex,
                                  final GridRow row, final Scenario scenario) {
        checkSimulation();
        super.insertRow(rowIndex, row);
        scenario.getUnmodifiableFactMappingValues().forEach(value -> {
            FactIdentifier factIdentifier = value.getFactIdentifier();
            ExpressionIdentifier expressionIdentifier = value.getExpressionIdentifier();
            if (value.getRawValue() instanceof String || value.getRawValue() instanceof Integer) {
                String stringValue = String.valueOf(value.getRawValue());
                int columnIndex = simulation.getSimulationDescriptor().getIndexByIdentifier(factIdentifier, expressionIdentifier);
                setCellGridOnly(rowIndex, columnIndex, () -> new ScenarioGridCell(new ScenarioGridCellValue(stringValue)));
            } else {
                throw new UnsupportedOperationException("Only string and integer are supported at the moment");
            }
        });
        updateIndexColumn();
    }
```

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
```diff
@@ -149,20 +144,20 @@ public Range deleteNewRow(int rowIndex) {
...
     }
 
     /**
      * This method <i>insert</i> a new column to the grid <b>without</b> modify underlying model
      * @param index
      * @param column
      */
-    @Override
-    public void insertColumn(final int index, final GridColumn<?> column) {
+    //@Override
```
jomarko: Please remove this comment.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
```diff
@@ -313,25 +322,14 @@ public void updateHeader(int columnIndex, int rowIndex, String value) {
...
-        return super.deleteCell(rowIndex, columnIndex);
-    }
-
     public void clear() {
         // Deleting rows
         int to = getRowCount();
         IntStream.range(0, to)
                 .map(i -> to - i - 1)
-                .forEach(this::deleteRow);
+                .forEach(super::deleteRow);
```
jomarko: If the reverse order is really so important, I think we could use the code below, let me know what do you think.
```java
IntStream.range(0, to - 1)
                .boxed()
                .sorted(Collections.reverseOrder())
                .forEach(super::deleteRow);
```

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
gitgabrio: I would prefer keep things separated - I'm not sure we will actually manage integer columns, in which case I would just delete the "else"

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
gitgabrio: The scope of that code was actually to avoid boxing and reverting order - with that implementation I loop just once; with your's, two (beside the boxing)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
jomarko: ok

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
jomarko: ok

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
jomarko: @gitgabrio please remove the comment and I can approve

gitgabrio:
@manstis 
Hullo! This could be merged - please squash commits. With @danielezonca we'll begin a different strategy to avoid these proliferation of "small" PRs.

manstis:
@gitgabrio Please let me know when it's green and I'll happily merge. Thanks.

-- and I don't mind the "..proliferation of 'small' PRs.." keep them coming (takes me no time to merge them).

PR #951, opened 2018-10-02
Drools 3051 - Implement placeholder for ScenarioGridModel

gitgabrio:
Scope of this PR is to Implement a HTML-like placeholder mechanism for ScenarioGridModel. More specifically, empty cells should show a "placeholder-like" string  (Insert value).
After deleting the value of a cell, the "placeholder" should be shown.
Cell of new rows/columns, or soon after asset creation, should also show the placeholder.
Verify also inserted values (if not empty) are correctly saved.

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/ScenarioCellTextBoxDOMElement.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/ScenarioCellTextBoxDOMElement.java
@@ -17,8 +17,8 @@
 package org.drools.workbench.screens.scenariosimulation.client.factories;
 
 import org.drools.workbench.screens.scenariosimulation.client.models.ScenarioGridModel;
+import org.drools.workbench.screens.scenariosimulation.client.values.ScenarioGridCellValue;
 import org.gwtbootstrap3.client.ui.TextBox;
-import org.uberfire.ext.wires.core.grids.client.model.impl.BaseGridCellValue;
 import org.uberfire.ext.wires.core.grids.client.widget.dom.impl.TextBoxDOMElement;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.GridWidget;
 import org.uberfire.ext.wires.core.grids.client.widget.layer.GridLayer;
@@ -33,13 +33,9 @@ public class ScenarioCellTextBoxDOMElement extends TextBoxDOMElement {
     public void flush(final String value) {
         final int rowIndex = context.getRowIndex();
         final int columnIndex = context.getColumnIndex();
-        if (value == null || value.trim().isEmpty()) {
-            ((ScenarioGridModel) gridWidget.getModel()).deleteNewCell(rowIndex,
-                                                                      columnIndex);
-        } else {
-            ((ScenarioGridModel) gridWidget.getModel()).setNewCellValue(rowIndex,
-                                                                        columnIndex,
-                                                                        new BaseGridCellValue<String>(value));
-        }
+        String actualValue = (value == null || value.trim().isEmpty()) ? null : value;
+        ((ScenarioGridModel) gridWidget.getModel()).setNewCellValue(rowIndex,
+                                                                    columnIndex,
+                                                                    new ScenarioGridCellValue(actualValue));
     }
 }
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/ScenarioGridColumnRenderer.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/ScenarioGridColumnRenderer.java
@@ -15,8 +15,35 @@
  */
 package org.drools.workbench.screens.scenariosimulation.client.renderers;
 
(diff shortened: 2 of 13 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
```diff
@@ -28,6 +28,7 @@ testTools=Test Tools
 testEditor=Test Editor
 scenarioCheatSheet=Scenario Cheatsheet
 runScenarioSimulation=Run Test
+insertValue=(Insert value)
```
Rikkola: Are we sure we want to use the brackets and second question is, are we sure we want to add them as part of the String here?
Just checking the brackets are actually needed.

gitgabrio:
@Rikkola 
Removed brackets :)

jomarko:
@gitgabrio Hi, worth to review from my site? Or are you going to apply @manstis's proposal?

manstis:
@gitgabrio I only commented as @danielezonca [asked](https://github.com/kiegroup/drools-wb/pull/951#pullrequestreview-160672497) whether we should move support for placeholders to `uberfire-wires-core-grids`. I'm happy with the way you've done it here.

gitgabrio:
@manstis 
Yup, understood. I've implemented your suggestion to ease moving the "placeholder" concept on upper level, in the future; actually, I like more to have it on GridCellValue. 
For the moment being I'm working on our subclasses but I think the idea is to move the code to upper ones (probably slight modifications will be needed)
BTW, maybe it would be better to avoid the possibility to have a gridcell without a gridcellvalue (for sure there is a reason it is so,  but I do not know, and I do not know if it is possible).  Thx! 👍 
@jomarko 
You could review this, whenever you can 😄

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/ScenarioCellTextBoxDOMElement.java
```diff
@@ -33,13 +33,9 @@ public ScenarioCellTextBoxDOMElement(TextBox widget, GridLayer gridLayer, GridWi
     public void flush(final String value) {
```
jomarko: Would you mind to add coverage for this method?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/renderers/ScenarioGridColumnRenderer.java
```diff
@@ -15,8 +15,44 @@
...
+import org.drools.workbench.screens.scenariosimulation.client.widgets.ScenarioGridCell;
+import org.uberfire.ext.wires.core.grids.client.model.GridCell;
+import org.uberfire.ext.wires.core.grids.client.widget.context.GridBodyCellRenderContext;
 import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.columns.impl.StringColumnRenderer;
+import org.uberfire.ext.wires.core.grids.client.widget.grid.renderers.grids.GridRenderer;
 
 public class ScenarioGridColumnRenderer extends StringColumnRenderer {
 
+    @Override
+    public Group renderCell(final GridCell<String> cell,
```
jomarko: Would you mind to cover this method?

danielezonca:
@manstis
Thanks for the feedback, I created this JIRA to track this activity
https://issues.jboss.org/browse/DROOLS-3074

gitgabrio:
@jomarko 
Hullo. Add tests.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/renderers/ScenarioGridColumnRendererTest.java
```diff
@@ -0,0 +1,102 @@
...
+        verify(scenarioGridColumnRenderer, never()).renderPlaceholderCell(eq(scenarioGridCell), eq(contextMock));
+        scenarioGridCell = new ScenarioGridCell(new ScenarioGridCellValue(null, "PLACEHOLDER"));
+        retrieved = scenarioGridColumnRenderer.renderCell(scenarioGridCell, contextMock);
+        assertNotNull(retrieved);
+        verify(scenarioGridColumnRenderer, times(1)).renderPlaceholderCell(eq(scenarioGridCell), eq(contextMock));
+    }
+
+    @Test
+    public void renderPlaceholderCell() {
+    }
```
jomarko: @gitgabrio thanks for added tests! could you please remove this empty one?
```java
    @Test
    public void renderPlaceholderCell() {
    }
```

danielezonca:
@jomarko 
No it is not correct if you need to save and reopen to see the placeholder. As soon as you lost the focus from the cell and the input box disappear the placeholder should be already there. If not I think it is a bug

gitgabrio:
@jomarko 
Fixed behaviour and add placeholder test

gitgabrio:
@manstis
Hi! Could you merge this and squash the commits? Thx!

manstis:
@Rikkola Could you please approve before I merge. Thanks.

PR #952, opened 2018-10-04
Drools 3053 Warning popup of column type change in Scenario grid

gitgabrio:
Scope of this PR is to show a warning popup when user try to change the column type of a not empty column.
*Empty column* is a column whose cells are all empty.

**Acceptance criteria**

If the column contains no value, no confirmation is needed
If old and new types are the same the user has to decide if the values should be removed or not
If old and new types are not the same the user has to accept that old values will be removed
All warning popups should contains also a "Cancel" command to abort the edit
If the user try to update the column type with the same column type no warning should be raised and data should stay untouched


gitgabrio:
@Rikkola 
HI Tony. I started extending it, actually. Unfortunately all the properties access modifiers are "package" or "private", without getter/setter (IMHO not a good design for widgets or classes that potentially could/should be extended) so actually it is impossible to extend it (I do not consider the option to recreate the same package inside scenario sources). So, at the end of the day I had to copy all the original code/html and add mine.
Beside that, anyway the original version or mine extension could not cope with the last UX proposal (still under discussion, though).

gitgabrio:
@jomarko 
Created ticket for future implementation:
https://issues.jboss.org/browse/DROOLS-3081

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/CommandExecutor.java
```diff
@@ -187,7 +197,21 @@ public void handle(ScenarioGridReloadEvent event) {
...
+        } else if (model.isSameSelectedColumnProperty(event.getValue())) {
+            return;
+        } else if (model.isSameSelectedColumnType(event.getValueClassName())) {
+            Command yesCommand = () -> commonExecute(new SetColumnValueCommand(model, String.valueOf(new Date().getTime()), event.getFullPackage(), event.getValue(), event.getValueClassName(), scenarioGridPanel, scenarioGridLayer, false));
+            Command noCommand = () -> commonExecute(new SetColumnValueCommand(model, String.valueOf(new Date().getTime()), event.getFullPackage(), event.getValue(), event.getValueClassName(), scenarioGridPanel, scenarioGridLayer, true));
+            yesNoConfirmPopupPresenter.show(ScenarioSimulationEditorConstants.INSTANCE.updateColumn(), ScenarioSimulationEditorConstants.INSTANCE.dataInColumnKeepOrDelete(), InlineNotification.InlineNotificationType.WARNING, CommonConstants.INSTANCE.Delete(), ScenarioSimulationEditorConstants.INSTANCE.keep(), Button.ButtonStyleType.DANGER, Button.ButtonStyleType.DANGER, null, yesCommand, noCommand);
```
jomarko: @gitgabrio I see this exceeding of recommended amount of chars per line in scenario-simulation often and I am not asking for fix as it is ont official requirement for code style but in this case I will make an exception, could you plase format as bellow? I know it is caused by inapproparite `show` method signature but will really facilitate reading and reviewing code.
```java
yesNoConfirmPopupPresenter.show(ScenarioSimulationEditorConstants.INSTANCE.updateColumn(),
                                            ScenarioSimulationEditorConstants.INSTANCE.dataInColumnKeepOrDelete(),
                                            InlineNotification.InlineNotificationType.WARNING,
                                            CommonConstants.INSTANCE.Delete(),
                                            ScenarioSimulationEditorConstants.INSTANCE.keep(),
                                            Button.ButtonStyleType.DANGER,
                                            Button.ButtonStyleType.DANGER, 
                                            null, 
                                            yesCommand,
                                            noCommand);
```

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/CommandExecutor.java
```diff
@@ -187,7 +197,21 @@ public void handle(ScenarioGridReloadEvent event) {
...
+        } else if (!model.isSameSelectedColumnType(event.getValueClassName())) {
+            Command okCommand = () -> commonExecute(new SetColumnValueCommand(model, String.valueOf(new Date().getTime()), event.getFullPackage(), event.getValue(), event.getValueClassName(), scenarioGridPanel, scenarioGridLayer, false));
+            yesNoConfirmPopupPresenter.show(ScenarioSimulationEditorConstants.INSTANCE.updateColumn(), ScenarioSimulationEditorConstants.INSTANCE.dataInColumnWillDelete(), InlineNotification.InlineNotificationType.WARNING, CommonConstants.INSTANCE.OK(), Button.ButtonStyleType.DANGER, null, okCommand);
```
jomarko: same comment as above.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModel.java
```diff
@@ -203,6 +207,11 @@ public void updateColumnType(int columnIndex, final GridColumn<?> column, String
         IntStream.range(1, elements.length)
                 .forEach(stepIndex ->
                                  factMappingByIndex.addExpressionElement(elements[stepIndex], lastLevelClassName));
+        if (keepData) {
+            IntStream.range(0, getRowCount())
+                    .forEach(rowIndex ->
+                                     setCellValue(rowIndex, columnIndex, originalValues.get(rowIndex)));
```
jomarko: Please format to one line:
```
.forEach(rowIndex -> setCellValue(rowIndex, columnIndex, originalValues.get(rowIndex)));
```

gitgabrio:
@manstis 
There is still an ongoing discussion from UX about the final look of it. But from what I saw of their proposal I fear I will have to do completely different, new widget; so before putting it as it is in appformer I want to see how it ends. Beside the final look, the implementation is done, and in the meantime the code can be reviewed and tested (the panel itself is only a by-side product of all the PR).

jomarko:
@gitgabrio generally I agree with @manstis about DROOLS-3081 to be resolved as part of this, however didn't wanted to force as I am not so aware of all tasks and its priorities for `scesim` team.

jomarko:
@gitgabrio we put comment at the same time, yes, I think your explanation about totally new widget is valid reason for DROOLS-3081 separately, thanks.

gitgabrio:
@jomarko 
Done formatting

danielezonca:
jenkins please retest this

PR #956, opened 2018-10-08
[DROOLS-2687] Implement support for range conditions and list of values into scenario

danielezonca:
This PR add support for range conditions ( <, >, <=, >= ) and list of values ( [ 1, 2, ...  ] ) to scenario to specify expected values.

**Detailed specification**
GIVEN columns:
- Only equals is supported on GIVEN columns ( = ). It can be used to specify empty strings because by default an empty string is considered as a null value. If the user tries to use other operators into GIVEN column scenario run fails.

EXPECTED columns:
- Equals: = or none (it is the default)
- Not equals: ! or != or <> . Not equals is the only operator that can be composed for now, so for instance "! < 3" or "! [ 1, 2, 3 ]" are valid expressions
- List of conditions: ; as separator character. All the conditions should be true. For instance "2 ; >3" is a valid expression
- Ranges: < or > or <= or >=. For instance "< 2" is a valid expression
- List of values: [ and ]. Comma is the separator character. Defined semantic is "at least one of them"

This PR change the general approach from an single operator approach to expression to be evaluated (to be ready to plug other expression languages like FEEL, MVEL or similar). This also has an impact on XML format so you need to create a new scenario to test this PR.

Small bugfixes included:
- Scenario is now saved under test resources instead of main
- No more error running scenario with empty description. Full fix for this issue (to have a better test result report) will be track under this JIRA https://issues.jboss.org/browse/DROOLS-3085

@Rikkola @gitgabrio @kkufova


danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

danielezonca:
@kkufova 
Updated the code and update PR description:
_List of conditions: ; as separator character. All the conditions should be true. For instance "2 ; >3" is a valid expression
Ranges: < or > or <= or >=. For instance "< 2" is a valid expression_

Now to specify multiple condition you should use ; instead of , but you can compose the conditions as you prefer like ">1; ! <3; [ 2, 3 ]; 5"

danielezonca:
@kkufova 
Yes I'm aware of this limitation and I have this jira for it
https://issues.jboss.org/browse/DROOLS-3085
I decided to postpone this fix because we are going to update result panel (so maybe that message will be not visible).
We plan also to have soon highlight the errors in the table that is a better alternative to current error reporting.

danielezonca:
@manstis 

Thanks

PR #957, opened 2018-10-09
Drools 3052 Implement read only mechanism for the cells in Scenario grid

gitgabrio:
Scope of this PR is to implement read only mechanism for the cells in Scenario grid.
More specifically:
1) Index column should never be editable
2) description column should always be editable
3) GIVEN/EXPECTED columns should not be editable if not mapped to a real property
Different placeholder messages should appear in the different condition (empty editable cell vs not editable one)
Cell editing should start with single click

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/AbstractCommand.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/AbstractCommand.java
@@ -62,6 +62,7 @@ public abstract class AbstractCommand implements Command {
      * <p>
      * isReadOnly: <code>true</code>;
      * </p>
+     * <p>
      * columnRenderer: new ScenarioGridColumnRenderer()
      * </p>
      * @param title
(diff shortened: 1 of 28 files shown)


gitgabrio:
@kkufova 
Should be fixed now

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/AbstractCommand.java
```diff
@@ -0,0 +1,91 @@
...
+                                                           ScenarioGridLayer gridLayer,
+                                                           String placeHolder) {
+        ScenarioHeaderTextBoxSingletonDOMElementFactory factoryHeader = FactoryProvider.getHeaderTextBoxFactory(scenarioGridPanel, gridLayer);
+        ScenarioSimulationBuilders.HeaderBuilder headerBuilder = getHeaderBuilder(title, columnId, columnGroup, factMappingType, factoryHeader);
+        return getScenarioGridColumn(headerBuilder, scenarioGridPanel, gridLayer, true, placeHolder);
+    }
+
+    protected ScenarioGridColumn getScenarioGridColumnLocal(ScenarioSimulationBuilders.HeaderBuilder headerBuilder) {
+        // indirection add for test
+        return getScenarioGridColumn(headerBuilder, scenarioGridPanel, scenarioGridLayer, false, ScenarioSimulationEditorConstants.INSTANCE.insertValue());
```
Rikkola: Why do the tests need this?

gitgabrio: @Rikkola 
static methods

gitgabrio:
@manstis 
Hullo! Could merge this, squashing commits? Thx!!

PR #961, opened 2018-10-12
[DROOLS-3124] Nested GridCell/GridCellValue

gitgabrio:

Clicking on a Grid cell toggle it to editing mode (insert a ScenarioGrdiCell); but clicking again on the same cell while editing will insert a new ScenarioGridCell inside the previous one.
Scope of this PR is to fix these mis-behavior.
To verify: repeatedly click on same cell should start its editing but no "nested" cell should be created (look at attachment here https://issues.jboss.org/browse/DROOLS-3124 to see how a "nested" cell appear)


gitgabrio:
@kkufova 
Fixed with 
[DROOLS-3124] Fixing editing enabling.

danielezonca:
I think this is a different minor bug that needs time to investigate and solve.
Let merge this fix for the main problem and track the remaining bug with this Jira
https://issues.jboss.org/browse/DROOLS-3150

gitgabrio:
@manstis 
Hullo! Could you please merge this and squash commits? Thanks!

PR #963, opened 2018-10-15
Drools 3095 Grid with multiple data/domain object instances

gitgabrio:
Scope of this PR is to allow creation of two-level headers inside grid.
More in detail, the first level group the type of columns (GIVEN/EXPECTED)
The second level group the properties by the instance they belongs to
The third level represent the actual property mapping.

See . here https://issues.jboss.org/browse/DROOLS-3095 for more details.


gitgabrio:
@kkufova @danielezonca 
Hi Klara, I will split your comment in three parts:
1) UX modification/improvement
2) UI modification/improvement
3) bugs.

For me, everything related to the "usage" is a UX decision, so I will follow what they will propose.
The look-n-feel (text color, background, whatever) it is a matter of UI: as soon as I have the right html to use, I'll put it in code.
So, for the moment being I will fix the bugs you discovered.

gitgabrio:

Bugs 4 5 6 and 7 should have been fixed.
Implemented workflow as per UX request (first select item and then click on "Add" button)
To test thoroughly also for regressions.

kkufova:
Note to myself: check [DROOLS-3054](https://issues.jboss.org/browse/DROOLS-3054) within this PR as well.

gitgabrio:
@kkufova 
Hi Klara. First, maybe there is some mismatching between version, or something else, not sure. 
Anyway:
Points 1-3, here's my screenshots:

screenshot from 2018-10-24 15-29-31


screenshot from 2018-10-24 15-30-20

I tried to recreate the structure of your project, but I can not reproduce the issue you mentioned.

Point 4, yes, for the moment being if the user already set an instance and a property,  the instance "type" can not be changed. It is just our proposal for the moment being.

Point 5 - actually two different issues: the first one is that when you click on a header a "filter" is applied to the right panel - I have to improve the click management because with this feature there are some new "corner" cases to manage; about the exception: a bug

Point 6: it is all inherited from the component we tried to reuse. Really, too much work to make it works as we need.

Point 7: as you may see from my screenshot, I see highlight and whatever *but*  I have two monitors, and I do not see the highlighting in one of them; and that reminded me that the same issues happened in a project I followed years ago.  Silly as it may sound, try to change the color setting of your screen

Point 8: It's not a bug, it's a feature! :smile: :smile: :smile:

danielezonca:
@kkufova 
I tried 1. and I have no visual bug so I think it is something on your machine 

@gitgabrio 
Here some bugs/changes:
1. when I create an instance, it should be available on right panel as existing instance even if I keep the default name: for instance if I create a `Person` I should find `Person [Person]` in the list of existing instance
2. if I click out of the table to remove the selection, the right panel keeps same status instead of reset to default one
3. change GIVEN 1, EXPECTED 1 etc to INSTANCE 1 and third level header to PROPERTY 1 to better match the new meaning

gitgabrio:
@danielezonca 
point 2-3
fixed with last commits
point 1: I would keep it as it is; I do not understand fully the scope (it will led to duplication of what is already in place in the "data object" part of the list; and the behavior would be just the same) and it would add complexity to code.

kkufova:
@gitgabrio, I must say that the PR looks great. I found only a few bugs:

1. It is not possible to enter any values into the table.
2. The placeholders are not correct (they are the same regardless the column type is set or not):
bug1
3. If you click on a property header (without a type + the instance header does not have a type either), everything from the right panel disappears. That's why it is **not possible** to select the property header type first (and automatically populate the instance header).
4. However, it is possible to do this:
bug2
**How to achieve it?** Click on the instance header and select a FIELD in the right panel. The property header gets populated, but the instance header does not.
5. It's also possible to achieve this:
bug3
First, you need to set the instance header to `Author`, then the property header to `isAlive`, and then change the instance header to `Book`.

gitgabrio:
@kkufova 
1: I *can* introduce values
2: Fixed 
3: I *can not* reproduce
4: I *can not* reproduce
5: Fixed

Please check for regression, too.

kkufova:
From our latest call with @gitgabrio:

1. It is possible to mismatch instance header and property header: set Librarian as an instance header and then `isAvailable` as a property header (to do it, you must set it after clicking on the instance header again). The instance header does not switch to Book and all tests will fail.
2. The Description column placeholder is not correct (it's "_Define valid type_").
3. Populate the table with types and add new column to the right. If you try to set its instance header, the warning popup opens ("_Change instance type?_").
4. Populate the table with types and add new column to the right. Click on PROPERTY header. The right panel is empty, which means that it is **not possible to set the property header before the instance header** and automatically populate the instance header.
5. Click on a second-level header (instance header). Select for example `isAvailable` from Book and click **Add**. The instance header is not populated with the correct type (Book). I think this is connected to the first point.
6. Once you set an instance type, it is not possible to introduce an alias (to type in your own name for the header).

danielezonca:
@gitgabrio 

7. if I select a property from an existing instance it should be filtered out from right panel evertything except that instance even if there are other partial match (for instance `Agenda` and `Agenda 1`). This means an exact match method
8. When user creates a new instance, filter out from right panel all existing facts in the same section (given/expect)
9. Disable instance update for defined instances
10. if I select a property already mapped, the right panel filter out all other instances but it doesn't expand the current one nor select the existing mapping
11. Update factIdentifier to be different for each different instances

gitgabrio:
@kkufova 
1: fixed making instance unchangeable
2: fixed
3: fixed
4: fixed (? could not reproduce)
5: fixed
6: fixed
7: fixed (but there is an issue if the user assign the same name to two instances)
8: fixed
9: fixed

gitgabrio:
@kkufova @danielezonca 
10: fixed

gitgabrio:
@kkufova @danielezonca  @Rikkola 

11: fixed

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/ListGroupItemViewImpl.java
```diff
@@ -92,6 +110,41 @@ public boolean isToExpand() {
         return toExpand;
     }
 
+    @Override
+    public void unselect() {
+        listGroupItem.removeClassName("selected");
+    }
+
+    @Override
+    public void enable() {
```
Rikkola: This and the disable() below do nothing and are commented out.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/ListGroupItemViewImpl.java
```diff
@@ -92,6 +110,41 @@ public boolean isToExpand() {
...
+//                });
+//        listGroupItemContainer.removeClassName("disabled");
+//        IntStream.range(0, listGroupItemContainer.getChildNodes().getLength())
+//                .forEach(index -> {
+//                    ((Element) listGroupItemContainer.getChild(index)).removeClassName("disabled");
+//                });
+    }
+
+    @Override
+    public void disable() {
```
Rikkola: Another commented out method.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -291,7 +307,7 @@ protected void loadContent() {
     void populateRightPanel() {
         // Execute only when RightPanelPresenter is actually available
         getRightPanelPresenter().ifPresent(presenter -> {
-            presenter.onDisableEditorTab();
+            // presenter.onDisableEditorTab();
```
Rikkola: if this is commented out, then does the dock get removed? If you open a DRL editor after this does the right side dock stay visible?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/handlers/ScenarioSimulationGridPanelClickHandler.java
```diff
@@ -129,23 +135,11 @@ public void onClick(ClickEvent event) {
         final int canvasY = getRelativeY(event);
         final boolean isShiftKeyDown = event.getNativeEvent().getShiftKey();
         final boolean isControlKeyDown = event.getNativeEvent().getCtrlKey();
-        // Workaround to differentiate click and double click on the same element
-        // we wait 300 ms to see  if another click happen in the meantime: if so, it is a double click,
-        // and we ignore it
```
Rikkola: Nice to see this went away.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/handlers/ScenarioSimulationGridPanelDoubleClickHandler.java
```diff
@@ -56,11 +56,12 @@ public void onNodeMouseDoubleClick(final NodeMouseDoubleClickEvent event) {
...
-            return handleBodyCellDoubleClick(event);
-        } else {
-            return true;
-        }
+        return true;
+//        if (!handleHeaderCellDoubleClick(event)) {
+//            return handleBodyCellDoubleClick(event);
+//        } else {
+//            return true;
+//        }
```
Rikkola: Commented code here. Not sure if this should be removed or not.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/ListGroupItemViewImpl.java
```diff
@@ -92,6 +110,41 @@ public boolean isToExpand() {
...
+//        factProperties.addClassName("disabled");
+//        IntStream.range(0, factProperties.getChildNodes().getLength())
+//                .forEach(index -> {
+//                    ((Element) factProperties.getChild(index)).addClassName("disabled");
+//                });
+//        listGroupItemContainer.addClassName("disabled");
+//        IntStream.range(0, listGroupItemContainer.getChildNodes().getLength())
+//                .forEach(index -> {
+//                    ((Element) listGroupItemContainer.getChild(index)).addClassName("disabled");
+//                });
```
Rikkola: Commented code still here and below.

gitgabrio:
@manstis 
Hi! Could you please merge and squash commits this! Thanks!

PR #969, opened 2018-10-22
[DROOLS-3167] Adjust Index/Description columns width.

gitgabrio:
Scope of this PR is to reduce the width of the "Index" column and move the gained space to the "Details" one.

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
@@ -183,6 +183,7 @@ public class ScenarioSimulationUtils {
                                                                                                     String placeHolder) {
         ScenarioSimulationBuilders.ScenarioGridColumnBuilder toReturn = ScenarioSimulationBuilders.ScenarioGridColumnBuilder.get(factoryCell, headerBuilder);
         toReturn.setPlaceHolder(placeHolder);
+        toReturn.setWidth(getColumnWidth(headerBuilder.getColumnId()));
         return toReturn;
     }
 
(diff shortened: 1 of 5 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
```diff
@@ -232,6 +233,17 @@ public static ScenarioGridColumn getScenarioGridColumn(ScenarioSimulationBuilder
         return headerBuilder;
     }
 
+    protected static double getColumnWidth(String columnId) {
+        switch (columnId) {
+            case "Index":
```
danielezonca: You can use ExpressionIdentifier.INDEX.getName()
WDYT?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
```diff
@@ -232,6 +233,17 @@ public static ScenarioGridColumn getScenarioGridColumn(ScenarioSimulationBuilder
...
+                return 70;
+            case "Description":
```
danielezonca: You can use ExpressionIdentifier.DESCRIPTION.getName()
WDYT?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
gitgabrio: I had to create an enum, but I think it was somehow missing anyway, so better that way. It is a little bit more verbose but less "magic words" around

gitgabrio:
@manstis 
Hi! Could you please merge this, squashing commits? Thx!

PR #971, opened 2018-10-22
[DROOLS-3170] Removing vertical scroll bar from RightPanel

gitgabrio:
Scope of this PR is to remove the vertical scroll bar from RightPanel.

Both scrollbars (horizontal and vertical) should appear only when needed (i.e the content is extending the visible margins) and automatically disappear when not needed.


gitgabrio:
jenkins retest this please

PR #972, opened 2018-10-23
[DROOLS-3133] Add support for EXPECTED conditions on new object

danielezonca:
With this change it is possible to have new instances in the Expected section.
It looks for an object that should match all the conditions defined on it, so it is not possible to have a partial fail: for instance if I create two conditions on Person.name and Person.age, both should be satisfied by the same object to have a success.

**How to test:**
- Create a drl that insert new object in the consequence (for instance an `Alert`) 
- Add a check to a scenario that verify the new object
- Run test

_NOTE: this PR introduce a dependency to drools-core that should be removed in the future, I created this Jira to track the activity https://issues.jboss.org/browse/DROOLS-3177_

@Rikkola @gitgabrio @kkufova


danielezonca:
Integrated into https://github.com/kiegroup/drools-wb/pull/975

PR #973, opened 2018-10-23
DROOLS-3142: [DMN Designer] Relation: Add additional Properties to drill-down. Updates required to drools-wb

manstis:
See https://issues.jboss.org/browse/DROOLS-3142

This PR contains a small refactor to `drools-wb` following changes in `uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids`

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableViewImplTest.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/GuidedDecisionTableViewImplTest.java
@@ -184,8 +184,9 @@ public class GuidedDecisionTableViewImplTest {
                                                                Collections.singletonList(0.0),
                                                                false,
                                                                false,
-                                                               HEADER_ROW_HEIGHT,
                                                                1,
+                                                               HEADER_ROW_HEIGHT,
+                                                               HEADER_ROW_HEIGHT,
                                                                0.0);
     }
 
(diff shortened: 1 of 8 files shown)


manstis:
Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/502
- https://github.com/kiegroup/kie-wb-common/pull/2196
- https://github.com/kiegroup/drools-wb/pull/973

gitgabrio:
@manstis 
I found out a compatibility issue with BaseGridWidgetMouseDoubleClickHandler
in your branch you deleted a protected method getHeaderRowsYOffset(). That broke my code, but maybe even someone elses' who could tell?
When dealing with frameworks classes we should strive to keep compatibility: if a method is not "private" inside  a not-final "framework" class, it means it *can* be used by extending classes. If so, we can not delete that method. In that case, I think it is better to change the method implementation but keep the "contract" intact.

manstis:
@gitgabrio Hi, this PR includes only the changes to keep the Simulation Testing Editor working following the changes in `uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids` it does not add full support for header selection as may be required for Simulation Testing. i.e. once the _generic_ support has been merged you'd need to write your own `CellSelectionStrategy` for your columns' header meta-data (that, for example, need to select the whole column). The _default_ behaviour is select only a single cell in the header.

As far as your comment about removing `getHeaderRowsYOffset()` from the super-class.. I agree and should not _break_ things (that's why there is this PR for Simulation Testing). There was unnecessary duplication of the method and hence it was refactored. **All** our code is considered _internal_ (other than `appformer` itself that was promoted in the community as a reusable framework - not the _extensions_). If somebody in the community uses _internal_ code it and it subsequently breaks it is unfortunate. I will try to be better in the future.

manstis:
@gitgabrio 

> The correct behavior is triggered by showing the rightpanel: do you have any idea on what could be the reason?

I suspect you have some code somewhere that is selecting cells for a column in the grid.

> Another (minor) issue I see, is that the header cells are inside a different box then the grid ones: is this fixable, somehow ?

Ah, well.. this is much more problematic. I was able to take advantage of a lot of the existing code in `appformer` to render the _header_ selections. However it does have the effect (as you have seen) that _header_ and _body_ selections are treated as two discrete blocks. This is good for when you have many, many rows and the header stays floating whilst the body scrolls. Less good (visually) when you have fewer rows.

You could add a column _highlight_ yourself (instead of rendering the cell/header selections themselves) much like is shown for when you drag a column to a new position. You'd need a custom `GridRenderer` to accomplish that.

PR #975, opened 2018-10-25
[DROOLS-2684][DROOLS-3133] Highlight errors on table + support conditions on new objects

danielezonca:
[DROOLS-2684]
Added error reporting directly on the table. It will be highlighted errors both in inputs (GIVEN) and outputs (EXPECTED). Input errors are a validation error (data not parsable), output errors can be wrong single values or the whole object in case of condition of new object (see below for more details)

Errors of a scenario (= of a row) are removed as soon as user edit a value in that scenario. All the errors are removed when user runs again the tests.

_NOTE: There is a known UI bug tracked in this Jira https://issues.jboss.org/browse/DROOLS-3191_

[DROOLS-3133]
With this change it is possible to have new instances in the Expected section.
It looks for an object that should match all the conditions defined on it, so it is not possible to have a partial fail: for instance if I create two conditions on Person.name and Person.age, both should be satisfied by the same object to have a success.

**How to test:**
- Create a drl that insert new object in the consequence (for instance an Alert)
- Add a check to a scenario that verify the new object
- Run test

_NOTE: this PR introduce a dependency to drools-core that should be removed in the future, I created this Jira to track the activity https://issues.jboss.org/browse/DROOLS-3177_

_Minor UI improvement: this PR introduce also better padding and columns width_

@Rikkola @gitgabrio @kkufova


mbiarnes:
Jenkins please retest this

mbiarnes:
Jenkins retest this please

danielezonca:
@kkufova 
1. Fixed
2. Fixed
3. Fixed
screenshot from 2018-10-26 17-24-35

I also updated the colors to match this proposal (the screenshot still contains the old one)
https://redhat.invisionapp.com/share/2MORFSO4VUX#/screens/327518515

danielezonca:
@manstis 

PR #977, opened 2018-10-26
Structure refactor

gitgabrio:
do-not-merge

@manstis @Rikkola @porcelli @ederign @danielezonca @etirelli @mdproctor 
Definitions: 

- type: api, backend, client
- asset: a specific screen (e.g. ScenarioSimulation, EnumEditor, etc), made of an api, a backed, and a client module

Scope of this PR is to reorganize module so to have modules grouped by "type" and not by asset.
In that way, it is possible to define the set of dependencies shared by the same "type" of module in one single pom, instead of having them replicated for every pom of the same "type".
With this in place, will be possible to limit duplicated dependencies declarations and to verify hidden crossed-dependencies.
Later, the objective is to identify dependencies needed only for specific module and declare them only on the correct pom, and to remove all the dependencies that are declared all over only because they have been copy-and-pasted in the past.
Moreover, this new structure is a move toward the one required for the new tbroyer plugin, since it reflects the architecture needed by it.

This PR has no impact on hot-reload.

Suggested links:

https://books.sonatype.com/mvnref-book/reference/pom-relationships-sect-pom-best-practice.html


Rikkola:
I don't think we actually copy&paste just to get things working faster. We are forced by the enforcer plugin to have every dependency that is needed declared in the pom. Now why this is, not sure, but there must be a reason.

Rikkola:
I might ping @ge0ffrey and @yurloc here. Since the new plugin was mentioned.

manstis:
I'll defer decision on this PR to @porcelli 

The original split we had was meant to support dynamic composition of the workbench; when different screens could be added/removed at runtime. I believe there was proof of concept work undertaken a while ago but unfinished. IDK whether the requirement still exists or has been superseded now _new_ screens are meant to be TypeScript/ReactJS..... (perhaps I'm too early thinking _new_ development with this new stack is imminent).

I don't believe the structure in this PR supports dynamic composition.

gitgabrio:
@Rikkola 
During this refactoring I spent a lot of time with mvn dependency tool; at the end of the day, some of the dependencies where required by some screens, but not by all, moreover, they where absolutely the same - that's why I thought it was a copy-and-paste (the same I did when started with Scenario Simulation).
My choice has been to keep the ones declared only in one screen at "screen" level, and the others (required by more then one screen, even if not by all) in the "common" pom.
About enforcer plugin, in theory I could understand its usage; in the actual situation, I find strange to explicitly declare all the kie/drools dependencies that can't be of different version (if I am not wrong), and anyway this "broke" somehow one of the key feature of maven itself.
I've put a link in my initial comment to read what I am trying to achieve. 
Last, to make one of his proposal works, @ederign disabled completely that plugin.

gitgabrio:
@manstis 
I've been told about this "dynamic" composition, but also that actually it is not used anywhere (if I did not misunderstood).
Beside that, I take for granted that for each screen, api/backend/client must go together, none of this can work without the others, so this "triplet" is actually a functional unit.
I recreated this "units" here: https://github.com/gitgabrio/drools-wb/tree/STRUCTURE-REFACTOR/drools-wb-screens
Those poms are actually "bom" defining the (sub)modules they are "made" of. So, you could import one of this to have the "unit" at disposal. The enforcer plugin somehow prevent to completely fullfill the "maven" approach, because I had to declare again all the transitive dependencies.  
Beside that, I think this "dynamic" (or whatever) composition currently can't be achieved because some (most ?) of the screens are dependent *at runtime* by others to correctly works; to said it differently, if you try to remove some screen from drools wb you will discover that other won't work (even if the webapp compile and start correctly) anymore (and unfortunately not only the easy-to-spot like guided-table witrh xls counterpart): they need to find "component" at runtime that are provided or requested by other screen, in a completely hidden way.
This was actually one of the main reason behind this PR and the proposal to have specific "showcases" for each screen (this is not the place to discuss possible issues and solutions regarding multiplication of showcases): to identify and explicit dependencies, to remove hidden ones, and to move to "common" place shared dependencies, so that every screen is actually independent from others.
Please read my comment to Tony and the link I add in my initial proposal to see what I am aiming at.

Rikkola:
Even if this move happens, the dependencies between screens are still an issue that might have to get fixed at one point.

gitgabrio:
@Rikkola 
I agree. I think putting things in "order" would ease subsequent tasks.

yurloc:
## 1. Duplicate dependency declarations
The duplication of dependencies is not a truly duplicate code that deserves to be eliminated. Currently, the dependency declarations accurately describe compile, test and runtime dependencies of each individual Maven module. The fact that multiple modules have overlapping dependencies is a coincidence. Any such module might need to be refactored in future and may eliminate one of its dependencies or require a new dependency, which should be reflected in its own POM. If we move "common" dependencies to a parent POM, the information about which dependency is needed by which module becomes harder to be found. I think that losing this information is a too high price for eliminating a few lines of code.

> Duplication is generally a bad thing in software. We don’t like duplicated
code. When code is truly duplicated, we are honor-bound as professionals to
reduce and eliminate it.
>
> But there are different kinds of duplication. There is true duplication, in
which every change to one instance necessitates the same change to every
duplicate of that instance. Then there is false or accidental duplication. If
two apparently duplicated sections of code evolve along different paths—if
they change at different rates, and for different reasons—then they are not
true duplicates. Return to them in a few years, and you’ll find that they are
very different from each other.

-- Robert C. Martin, Clean Architecture (Pearson Education, Inc., 2018), 154-155.

## 2. Code organization
This is the second part of this PR. It seems to me that it has no functional impact and it's only needed to allow grouping of dependency declarations.

> Moreover, this new structure is a move toward the one required for the new tbroyer plugin, since it reflects the architecture needed by it.

@gitgabrio Can you elaborate on this? Any pointers on what the tbroyer plugin code organization requirements are?

Sorry, it's not clear to me at this point if the structure change is really needed to accomplish something. Unless it's necessary, I'd prefer to keep the current structure as it has a higher cohesion -- all the code related to one feature (screen) can be found under a single root and isolated from other features' code, whereas after the structure change the three components of each screen would be dislocated under three different roots. So, for example, in order **to build a single screen** (to see if it compiles, to run all tests, to run code analysis etc.), I would need to **run three separate Maven builds** (in the correct order). The current organization is closer to *package-by-feature*, the new organization is closer to *package-by-layer* (see [Package by feature, not layer](http://www.javapractices.com/topic/TopicAction.do?Id=205)).

So -1 from me for the reasons I've mentioned (inaccurate dependency declarations, lower code cohesion) unless the change is necessary to make something else work, which I currently don't see (maybe I'm missing the need for this change but I can't remember that tbroyer plugin would require this).

porcelli:
agree with @yurloc; this is how we used to operate on old days of guvnor (5.x) and we had enough issues.

-1

gitgabrio:
@yurloc 
Hi, first of all, thanks for the time you give to answer. Please take in account I know that I may be wrong,  anyway I'll try to explain my thoughts. It is going to be a very long comment, though, and I'll try to organize it as best as I could.
##  1. Comments on your quotes
First, your  quotes, interesting as they are, are referring to actual code, not to module organization. And I think they are covering different aspects of the code-life (pom, bom, and other are concept strictly related to Maven, or other high level project management tool, while the code itself could be built without it) 
### a. Duplication
My very first point would be to create *single* poms to group all the dependencies *absolutely needed* (both at compile time and runtime) by every specific submodule, and use submodule' pom to declare only dependencies needed by it - to give them a life actually independent from each other (I think this is what is meant in your first quote).
### b. Packaging
This is related on how the classes are packaged in the jars, (I perfectly agree), and not on how the modules are layered inside a maven project.
##  2. Tbroyer plugin
In its archetypes/example (I followed webapp-multimodule) there is a client/shared/server architecture, and my point is to simply follow that design in a multimodule structure (I made a very simple project showing that here https://github.com/gitgabrio/gwtmultimodule).
##  3. Single vs multiple maven build
In the new structure there is a folder with "aggregators" (https://github.com/gitgabrio/drools-wb/tree/STRUCTURE-REFACTOR/drools-wb-screens)  to achieve what you say.

##  4. What, and why, am I aiming at?
In the past I've worked - and even created from scratch - with different gwt projects, at least one of them pretty complex, and when I begun working with drools-wb I found a lot of things "strange" to me. During the time @Rikkola @manstis and @porcelli explained me why they are there, so I do know the historical reason; nevertheless, I think that the actual situation contains some "issues" that could be fixed/improved.
The very first one is that - regardless of supposed modularity - actually the screens are bound together, you can not work (out of the box) on any of them outside drools-wb (that put them all together). I spent hours in trying to find the right "combination" for the showcase I used for scenario simulation. Some times ago I made an archetype (here https://github.com/gitgabrio/drools-wb-showcase-archetype) to realize a screen-specific showcase, but even if in the created module the dependencies (the ones I can be aware of) are there, it fails to start due to some missing dependency required at runtime (and this frustrates any attempt to use the pom-declared dependencies as the actual requirement for a screen). Having all the *absolutely-needed* shared dependencies grouped (by layer, following your quote) would ease the work to identify and manage them; for example, if a new dependency is required by all submodules of the same layer, or a modification of the same kind is required, it is possible to do that on a single pom (if I did not misunderstood, currently @ederign makes manually such modifications on every pom when needed - and I think this is pretty error-prone and time-consuming).
It just crossed my mind an email where we have been requested a list of the requirements for each module: probably having this kind of centralized knowledge  would help.
Another things, but here I could be very wrong: when the new js approach would be in place, I think (I could be wrong, of course) that the new "js" client modules would/could have different dependencies than their "java" counterpart; keeping the "by-feature" grouping would increase (IMHO) the duplication.

@porcelli 
About guvnor, I may be wrong, but I remember you told me that inside guvnor the application was "monolithic" (i.e. not modules) while what I'm proposing it is simply (taking @yurloc quote) to change group "by-feature" to group "by-layer" - the level of grouping would be absolutely the same.

# Conclusion
I am the very first one saying this modification has impacts and costs that now we can't afford (that's why I wrote "do-not-merge" as the very first line).
Anyway, in my experience putting things in good (or, at least, best) order in the long terms buy by itself.
I've read with interest your different opinions and I hope to have expressed clearly mine, regardless about when(if) ever this modification will come to light.

PR #970, opened 2018-10-22
[DROOLS-3119] Implement a Junit activator to be able to run Scenarios via maven

danielezonca:
This PR implement a Junit activator and the mechanism to inject it into user projects be able to run test scenarios via Maven.

**How it works:**
Every time user creates/saves a Test Scenarios asset the service add ScenarioJunitActivatorTest.java as test classes in the default package. It also adds some dependencies to project's POM as test runtime support.

**Dependencies:**
`org.drools:drools-compiler:7.14.0-SNAPSHOT:test`
`org.drools:drools-wb-scenario-simulation-editor-api:7.14.0-SNAPSHOT:test`
`org.drools:drools-wb-scenario-simulation-editor-backend:7.14.0-SNAPSHOT:test`

**How to test:**
- Perform `mvn install` of the project drools-wb/drools-wb-screens/drools-wb-scenario-simulation-editor
- Create a project
- Clone it to verify that no ScenarioJunitActivatorTest class nor POM dependencies added
- Create one or more test scenarios instances
- Refresh local copy
- Perform `mvn test` to run tests

_NOTE: to be able to build locally an user project you need first to change version of plugin kie-maven-plugin to 7.13.0.Final or another non-SNAPSHOT version (it seems there is no SNAPSHOT version of the plugin available on our Nexus)_

@Rikkola @gitgabrio @kkufova

PR #978, opened 2018-10-29
[DROOLS-3204] Fix null checks + NPE on empty scenarios

danielezonca:
Contains a list of fixes:
- NPE when run an empty scenario (both with internal runner or with junit one)
- Fix wrong `null` management on expect fields
- Use TextArea instead of TextBox for inline editing
- Updated "Description" and "Index" labels
- Updated placeholder colors

@Rikkola @kkufova @gitgabrio 

To be merged after https://github.com/kiegroup/drools-wb/pull/975 and https://github.com/kiegroup/drools-wb/pull/970


danielezonca:
@manstis 

PR #989, opened 2018-11-09
[DROOLS-3290] Bump scesim file version to 1.1, added automatic migration support (in memory for now)

danielezonca:
Define automatic migration mechanism. It is in memory for now to be compatible with both internal WB and external maven runner.

@kkufova @Rikkola @gitgabrio

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/util/InMemoryMigrationStrategy.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/util/InMemoryMigrationStrategy.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/util/InMemoryMigrationStrategy.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package org.drools.workbench.screens.scenariosimulation.backend.server.util;
+
+import java.util.function.Function;
+
+public class InMemoryMigrationStrategy implements MigrationStrategy {
+
+    @Override
+    public Function<String, String> from1_0to1_1() {
```
gitgabrio: I think this method name should be changed to "migrate" or something anyway "agnostic" of the actual migration level.

danielezonca:
@gitgabrio 
Yes my idea is to have a specific method for migration for each version (1.0 -> 1.1, 1.1 -> 1.2, etc). I agree that this is not the best if we will have a lot of migrations steps but this has the advantage for now that it is statically defined and it is impossible to forget to implement any step of the migration strategy.
I would like to keep this approach and move to a generic list of ordered migration steps only if needed (aka > 10 steps)

danielezonca:
@manstis 

Rikkola:
@gitgabrio @danielezonca It is unlikely that we need that many steps. Adding new items does not need migration, only if something is removed/changed. And I think step by step migration from one version to the next is the only way we can stay sane.

PR #990, opened 2018-11-12
JBPM-7869 : NoClassDefFoundError warnings when starting Workbench

Rikkola:
https://issues.jboss.org/browse/JBPM-7869

While the first fix worked for EAP at the same time it broke some tests that needed the exluded classes for testing. This moves anything that has tests and the tests to the -adapter module.

This the new location for the java files is probably where they should be anyway.

@manstis @jomarko If you two could please take a look. This just moves a group of file to the module next to it.


Rikkola:
Thanks @manstis that means this change did not bring the original issue back in.

Now we need QA to comment if this fixes the test setup I broke.

PR #996, opened 2018-11-14
DROOLS-2910: [DMN Designer] Update cell (double-)click interactions

manstis:
See https://issues.jboss.org/browse/DROOLS-2910

This PR _fixes_ "scesim" following changes in https://github.com/kiegroup/appformer/pull/528

I say _fixes_ as it appears "scesim" does not use the _default_ mechanism for "editing" cells.

I tested the Guided Decision Table Editor and "scesim". Both _appear_ to operate OK.


manstis:
Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/528
- https://github.com/kiegroup/kie-wb-common/pull/2273
- https://github.com/kiegroup/drools-wb/pull/996

gitgabrio:
@manstis 
Hi. I 've launched your modification. Actually it still works but the handler you add is ignored.
I've put log and breakpoint but ScenarioSimulationGridWidgetMouseEventHandler.handleHeaderCell is never invoked.
I think this would require from me a refactoring on all the click management, as I told you some days ago, but for the moment being I think it does not matter.
Beside that, I see you created ScenarioSimulationGridWidgetMouseEventHandlerTest. In that same package there is already an AbstractScenarioSimulationGridPanelClickHandlerTest used to define click/positions/mock, used inside actual  handlers "Test" implementation - I've wrote that to avoid duplicating declarations and definitions all over test classes: could you please reuse it ? Thanks!

manstis:
@gitgabrio Changes (to extend `AbstractScenarioSimulationGridPanelClickHandlerTest`) added in new commit.

karreiro:
Jenkins please retest this.

PR #997, opened 2018-11-14
[DROOLS-3257] Implemented support for char, byte and short, fixed NPE if data to test is null

danielezonca:
This PR add the possibility to use DO with fields of type char, byte or short.
It fixes also a NPE if the actual value is `null` and user is trying to test it with a condition like `! ok`

@Rikkola @gitgabrio @kkufova


danielezonca:
jenkins please retest this

danielezonca:
@kkufova 
I did a test and I think it works, let me explain:

This is the rule I created (without dialect specification, I think default is `java`)
```
package com.myspace.testproject;

import com.myspace.testproject.Person

rule SetField
when
    $p : Person()
then
    if($p.getShortField() == null) {
        $p.setShortField((short) 1);
        $p.setCharField('a');
        $p.setByteField("0".getBytes()[0]);
    }
end
```

And this is the scenario
screenshot from 2018-11-21 15-32-05

In the first scenario I just tried to set a value in the `GIVEN` and verify the same in the `EXPECT` to be sure that the general parsing works.
At the same time with `!ok` I created a condition to check on a `null` field (`Person.name`)
The second scenario verifies the values set by the rule.
Note that:
1. You need to cast to `short` otherwise the validation fails (`[KBase: defaultKieBase]: Rule Compilation error The method setShortField(Short) in the type Person is not applicable for the arguments (int)`)
2. Using default dialect I can also use `char` values without any problem
3. Almost the same for for `byte`: without `"0".getBytes()[0]` or similar method that generate a valid byte value you have a validation fail (`[KBase: defaultKieBase]: Rule Compilation error The method setByteField(Byte) in the type Person is not applicable for the arguments (int)`)

danielezonca:
@manstis 

PR #998, opened 2018-11-15
[DROOLS-3322] It is not possible to set a column type by typing in the data object's name manually	

gitgabrio:
@kkufova @danielezonca 
Scoope of this pR is to prevent user to manually set an instance header with the name of a Data Object or an already existing instance


gitgabrio:
@kkufova 
1 I would avoid this, 'cause the user could insert an "illegal" value (for ex there is already an existing "Book" instance; so, we should have to introduce another check)
2 Agree, seems unreasonable. I'll block the editing before an actual instance has been assigned

gitgabrio:
@kkufova 
Done modifications. Please check again

kkufova:
@gitgabrio, also, please do not forget to create another PR to the `7.14.x` branch.

gitgabrio:
@manstis 
Hi Could you please merge this, squash commits? Thx!

PR #999, opened 2018-11-15
[DROOLS-3320] Insert column from property header near the clicked one

gitgabrio:

Scope of this PR is to have column inserted from "property" header attached to the original clicked one.


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/commands/InsertColumnCommandTest.java
```diff
@@ -51,11 +51,28 @@ protected ScenarioGridColumn getScenarioGridColumnLocal(String instanceTitle, St
 
     @Test
     public void execute() {
+        insertColumnCommand.asProperty = true;
+        insertColumnCommand.isRight = false;
+        insertColumnCommand.execute();
+        verify(insertColumnCommand, times(1)).getScenarioGridColumnLocal(anyString(), anyString(), eq(COLUMN_ID), eq(COLUMN_GROUP), eq(factMappingType), eq(scenarioGridPanelMock), eq(scenarioGridLayerMock), eq(ScenarioSimulationEditorConstants.INSTANCE.defineValidType()));
+        verify(scenarioGridModelMock, times(1)).insertColumn(eq(3), eq(gridColumnMock));
+        reset(scenarioGridModelMock);
+        reset(insertColumnCommand);
```
Rikkola: The added exposure for from private to protected is not necessary here. If you chunk each of these blocks to separate named tests it will be easier to see how the widget should be used and what is tested.

gitgabrio: @Rikkola 
I can not understand what you mean. "private/protected" access is set in target classes, while eventually splitting the method in more (sub)methods does not involve the access modifier.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/InsertColumnCommand.java
```diff
@@ -40,7 +40,7 @@
     private String columnId;
     private int columnIndex;
     protected boolean isRight;
-    private boolean asProperty;
+    protected boolean asProperty;
```
Rikkola: Sorry @gitgabrio. I mean this change from private to protected. If you use split the test and use the constructor this does not have to change and the class works as it is expected.

gitgabrio: @Rikkola 
Done.

gitgabrio:
@manstis
Hi Could you please merge this, squash commits? Thx!

PR #1002, opened 2018-11-20
[DROOLS-3338] Avoid reloading page on "Enter" pressed

gitgabrio:
Scope of this PR is to avoid that on "Enter" pressed the page get reloaded


gitgabrio:
@kkufova 
FIxed keydown

gitgabrio:
Not fixed.

PR #1003, opened 2018-11-21
[DROOLS-3162] Implement full page width support for core-grids 

danielezonca:
This PR adds the support to have columns with auto width to fit entirely the visible area. All `auto` columns try to fit the entire page except if this means to be shorted than the defined minimum, in this case the grid will be larger than then the visible area 

**How it works:**
- All the columns are `fixed` by default (to be retro-compatible)
- In scesim all `GIVEN` and `EXPECT` columns have `auto` width

@Rikkola @manstis @kkufova @gitgabrio 

PR list:
https://github.com/kiegroup/appformer/pull/543
https://github.com/kiegroup/drools-wb/pull/1003
https://github.com/kiegroup/kie-wb-common/pull/2299


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationBuilders.java
```diff
@@ -87,13 +87,17 @@ public ScenarioGridColumnBuilder setPlaceHolder(String placeHolder) {
...
+            ScenarioGridColumn scenarioGridColumn = new ScenarioGridColumn(headerMetaDataList,
+                                                                           actualScenarioGridColumnRenderer,
+                                                                           width,
+                                                                           isMovable,
+                                                                           factoryCell,
+                                                                           placeHolder
             );
+            // by default scenario columns should be auto to have auto resize
+            scenarioGridColumn.setColumnWidthMode(GridColumn.ColumnWidthMode.auto);
+            scenarioGridColumn.setMinimumWidth(width);
```
Rikkola: If these are the defaults for ScenarioGridColumn, should we set them inside the ScenarioGridColumn constructor?

danielezonca:
@kkufova 
Yes sorry I forgot to say that it needs this https://github.com/kiegroup/appformer/pull/542 to be tested. I need to have it merged then rebase master and then it can be tested.
Sorry for that

kkufova:
@danielezonca, no problem, just let me know when the PR is ready for review.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationBuilders.java
danielezonca: Yes you are right, moved to the constructor

danielezonca:
@kkufova 
Please try again, it should be fixed by https://github.com/kiegroup/appformer/pull/543/commits/a24e51a3da81354e8442f969f4bd8dce31661dde

danielezonca:
@manstis 

PR #1007, opened 2018-11-28
Drools 3347  -Implement undo/redo

gitgabrio:
@kkufova @Rikkola  @danielezonca 

Scope of this PR is to actually implement the UNDO/REDO behavior.
In our context, UNOD/REDO means restoring the overall status at a given moment. In other words (especially for redo) it does not mean re-execution of a command. Whenever a new command is executed, the "redo" list is cleared, so it is not possible for ex to issue a "Insert new row" and then repeat it with "Redo".


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/Simulation.java
```diff
@@ -109,6 +109,24 @@ public void resetErrors() {
         scenarios.forEach(Scenario::resetErrors);
     }
 
+    public Simulation cloneSimulation() {
+        Simulation toReturn = new Simulation();
+        // TODO CHECK ALL THE FOLLOWING *SHOULD BE* REAL CLONES
```
Rikkola: Would be better to make tickets for the TODO's. These tend to be left into the code and never checked.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/test/java/org/drools/workbench/screens/scenariosimulation/model/SimulationTest.java
```diff
@@ -56,6 +56,13 @@ public void addScenarioTest() {
                       IllegalArgumentException.class);
     }
 
+//    @Test
+//    public void cloneSimulationTest() {
+//        final Simulation retrieved = simulation.clone();
+//        assertNotNull(retrieved);
+//        assertFalse(simulation == retrieved);
+//    }
+
```
Rikkola: This should be removed or added back in.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/ScenarioSimulationEventHandler.java
```diff
@@ -0,0 +1,429 @@
...
+    @Override
+    public void onEvent(ReloadRightPanelEvent event) {
+        context.getStatus().setDisable(event.isDisable());
+        context.getStatus().setOpenDock(event.isOpenDock());
+        commonExecution(context, new ReloadRightPanelCommand());
+    }
+
+    @Override
+    public void handle(ScenarioGridReloadEvent event) {
+        // TODO CHECK FOR POSITION
```
Rikkola: Again here are TODOs that can be forgotten.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractScenarioSimulationCommandTest.java
```diff
@@ -0,0 +1,123 @@
...
+    /*
+
+      @Override
+    public CommandResult<ScenarioSimulationViolation> undo(ScenarioSimulationContext context) throws UnsupportedOperationException {
+        if (!undoable || restorableStatus == null) {
+            String message = !undoable ? this.getClass().getSimpleName() + " is not undoable" : "restorableStatus status is null";
+            throw new UnsupportedOperationException(message);
+        }
+        return setCurrentContext(context);
+    }
```
Rikkola: Commented code that should be removed or returned.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractScenarioSimulationCommand.java
```diff
@@ -42,13 +44,65 @@
...
+
+    /**
+     * Auto-generated incremental identifier used  to uniquely identify each command
+     */
+    private final long id;
+
+    /**
+     * Flag that indicates if the command is <b>undoable</b>. Default is <code>false</code>
+     */
+    private final boolean undoable;
```
danielezonca: I think this logic can be moved to AbstractCommand. WDYT?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractScenarioSimulationCommand.java
```diff
@@ -42,13 +44,65 @@
...
+    protected ScenarioSimulationContext.Status restorableStatus = null;
+
+
+    /**
+     * Calling this constructor will set the command as <b>undoable</b>
+     *
+     *
+     * @param undoable
+     */
+    protected AbstractScenarioSimulationCommand(boolean undoable) {
```
danielezonca: I think it could be useful to have also the default constructor with undoable = `true` (or `false` if you think that this should be the default) so it is not needed to call `super(true/false)` in all the implementations

gitgabrio: I would prefer that actual command must esplicitly declare if they are undoable or not (and so the developer) without giving a default choice, to enforce the requirement.
Giving a "default"  would only open the way for possible mistake without giving a real advantage (the default constructor should be anyway protected - if made public, the possible source of mistake would propagate to the caller code).

gitgabrio: Wording different - the developer who wrote the command should make the decision about "undoability" of the command itself, and the code using that command should not be able to modify that behavior.

gitgabrio:
@gitgabrio  @kkufova 
In some condition undo/redo led to failing test.
scesim file of failing tests have wrong class definition of value (ex. com.Book instead of java.lang.Integer.
Also, when the tests fails, undo seems to completely reset the status to the beginning, instead of just reverting to the previous one

[Failing_Working_scesims.zip](https://github.com/kiegroup/drools-wb/files/2643432/Failing_Working_scesims.zip)

kkufova:
I'm waiting for the last fix; "invisible" operations (such as selecting a cell) are causing "useless" undo/redo.

gitgabrio:
jenkins please retest this

gitgabrio:
jenkins please retest this

PR #1009, opened 2018-11-28
[DROOLS-3091] Negative values should support space (-x vs. - x)

danielezonca:
@kkufova @Rikkola @gitgabrio


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/util/ScenarioBeanUtil.java
```diff
@@ -219,4 +219,8 @@ private static byte parseByte(String value) {
         }
         return value.getBytes()[0];
     }
+
+    private static String cleanStringForNumberParsing(String rawValue) {
+        return rawValue.replaceAll("(-)\\s*([0-9])", "$1$2");
```
Rikkola: I always read these as curse words from comics.

danielezonca:
@manstis 

PR #1010, opened 2018-11-28
[DROOLS-3378] The number of executed tests is not correct when there are tests failures

danielezonca:
@kkufova @gitgabrio @Rikkola


danielezonca:
@manstis 

PR #1017, opened 2018-12-04
DROOLS-2932: Make edit selected cell logic general

jomarko:
@manstis @gitgabrio This contains changes from previous ensembles and the last commit make the edit selected cell action general.

Ensemble with:
- https://github.com/kiegroup/appformer/pull/561
- https://github.com/kiegroup/kie-wb-common/pull/2327


jomarko:
@ederign done
@gitgabrio please let us know if you are going to have a look or it is fine to merge without your approval.

gitgabrio:
@jomarko 
I'm reviewing

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/metadata/ScenarioHeaderMetaData.java
```diff
@@ -75,7 +75,7 @@ public ScenarioHeaderMetaData(String columnId, String columnTitle, String column
 
     public void edit(final GridBodyCellEditContext context) {
         if (readOnly) {
-            throw new IllegalStateException("A read only header cannot be edited");
+            return;
```
gitgabrio: @jomarko  @manstis  @ederign 
The reason I throw the exception there is because that method should never be invoked if the readOnly flag is true, and that control should be made by caller code **before** invoking this one.
Wording differently, it is meant exactly to raise an exception to make the developer aware that there is an error in the flow. Replacing it with the "return" will just "hide" error of implementations. Please restore the exception and eventually manage it where needed.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/metadata/ScenarioHeaderMetaDataTest.java
```diff
@@ -44,11 +44,13 @@ public void constructorFail() {
         new ScenarioHeaderMetaData("", "", "", factoryMock, true, true);
     }
 
-    @Test(expected = IllegalStateException.class)
```
gitgabrio: @jomarko @manstis  @ederign 
Please see my previous comment and restore the annotation/implementation.

jomarko:
Hi @gitgabrio,

ok, will revert. I see this as something we should discuss more. This revert will cause Unexpected error dialog when user will try to edit read only  cell. Will schedule meeting soon.

Note for myself - affetcs https://issues.jboss.org/browse/DROOLS-3437.

jomarko:
Changes pushed.

gitgabrio:
@jomarko 
Ok, thanks! I can not fully understand exactly "This revert will cause Unexpected error dialog when user will try to edit read only cell." If I am not wrong, this will happen when starting editing from keyboard, not from click. So, it could be somehow related on the comment I put there: https://github.com/kiegroup/kie-wb-common/pull/2327/files.
Basically, I think "edit" keyboard operation (and more generally other keyboard operations) should implement a check of "executability" before get invoked - that way, editing a "readOnly" cell would be impossible also from the keyboard, and if that happen throwing the exception will be welcomed, because it would mean there is a "bug" somewhere in the code.

PR #1021, opened 2018-12-07
Fix infinite loop during header row index computation

jomarko:
Due to recent changes in https://github.com/kiegroup/appformer/pull/556 the ScenarioSimulationGridHeaderUtilitiesTest started to fail on infinite loop


gitgabrio:
@jomarko 
Many thanks for your quick action and your modifications. But the actual problem was elsewhere, so I created a new PR to include your modification and the actual fix:

https://github.com/kiegroup/drools-wb/pull/1022

Again, many thanks for working while you are sick!
Gabriele

jomarko:
@gitgabrio ok, thank you, I am closing this one.

PR #1022, opened 2018-12-07
[DROOLS-3417] Avoiding endless loop on ScenarioSimulationGridHeaderUtilitiesTest

gitgabrio:
@kkufova @jomarko @Rikkola 

That loop was caused by a change inside upper layer.
Before this change, by default ScenariogridModel had 1 header row, even when called by the default constructor. After the change, the default header row number is 0.
Inside ScenarioSimulationGridHeaderUtilities there is a method that uses that row number in a "while" loop for a subtraction, but since now that number is 0, that loop will never end.
The fix has been to check for "0" rows before that loop, eventually returning null.
Tests have been modified as per @jomarko suggestion. A new one has been added.

Please check ASAP to merge it.


jomarko:
jenkins please retest this

jomarko:

PR #1023, opened 2018-12-10
[DROOLS-3408] Fix wrong parsing of list of values in expression

danielezonca:
NOTE: nested list of values and nested list of conditions are not supported and I don't think to support it because it requires to change completely the parsing logic. This expression language is intended to be really simple and for advanced use cases we will support `FEEL` and/or `MVEL`. For instance `[ [ 1, 2 ,3 ], 2 ]` or `1; [ 1 ; 2 ]` doesn't work

@Rikkola @gitgabrio @kkufova


danielezonca:
jenkins retest this please

PR #1026, opened 2018-12-11
[DROOLS-3356] Add support to simple java types as Data Object in the right panel

gitgabrio:

Scope of this PR is to add Simple java types to the Right panel.
Such types are not retrieved from dmo but (being standard java) manually inserted inside the presenter.
Behavior differences for DMN:
- header can't be editable
- FactIdentifier must be equals to name (=factAlias)
- there can be only one instance of a given type at all, regardless if in "GIVEN"  or "EXPECT"
Please test for unexpected behavior, especially when running scenario test.


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+import org.drools.workbench.screens.scenariosimulation.client.models.FactModelTree;
+
+/**
+ * Abstract class to provide common methods to be used by actual implementations.
+ */
+public abstract class AbstractDataManagementStrategy implements DataManagementStrategy {
+
+
+
+    protected FactModelTree getSimpleClassFactModelTree(Class clazz) {
```
jomarko: This seems to me as util method that could be used by **delegation** instead of **inheritance**. Is the inheritance used because of the testeability?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/DataManagementStrategy.java
```diff
@@ -26,9 +32,15 @@
  */
 public interface DataManagementStrategy {
 
+    Map<String, Class> SIMPLE_CLASSES_MAP = Collections.unmodifiableMap(Stream.of(
+            new AbstractMap.SimpleEntry<>(Boolean.class.getSimpleName(), Boolean.class),
+            new AbstractMap.SimpleEntry<>(Double.class.getSimpleName(), Double.class),
+            new AbstractMap.SimpleEntry<>(Integer.class.getSimpleName(), Integer.class),
+            new AbstractMap.SimpleEntry<>(Number.class.getSimpleName(), Number.class),
+            new AbstractMap.SimpleEntry<>(String.class.getSimpleName(), String.class)).
```
jomarko: This list seems to me as not complete list of simple types, have you considered reuse(extend and reuse) existing class that collects simple data types?

https://github.com/kiegroup/kie-soup/blob/master/kie-soup-project-datamodel/kie-soup-project-datamodel-api/src/main/java/org/kie/soup/project/datamodel/oracle/DataType.java

gitgabrio: This list reflect what we are currently supporting. It is a sublist of what the user can import from the "Import" tab. Inside that tab, there are also List/ArrayList/Collection - that does not makes sense for us.

jomarko: Ok, list of Types on Imports tab comes from this line [1] and is shown to the user here [2]. I still think we should not re-declare list of something existing. However have no resources investigate more neither I am going to approve or decline, wanted just discuss this topic.

[1]
https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-widgets/kie-wb-config-resource-widget/src/main/java/org/kie/workbench/common/widgets/configresource/client/widget/bound/ImportsWidgetPresenter.java#L83

[2]
https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-widgets/kie-wb-config-resource-widget/src/main/java/org/kie/workbench/common/widgets/configresource/client/widget/bound/ImportsWidgetViewImpl.java#L227

kkufova:
NOTE: aliases should be possible for rule-specific scenarios, but not for dmn-specific scenarios.

danielezonca:
@kkufova 
The DMN one is not a bug. It is impossible to edit a value during DMN evaluation so it is useless to use the same instance both in GIVEN and EXPECT. For DMN you can set the instance only in once in the whole table

kkufova:
@danielezonca, you're right! It makes sense. Thank you for the explanation!

PR #1029, opened 2019-01-03
[DROOLS-3356][DROOLS-3362][DROOLS-3369] Add DMN data type retrieve and support to Test Scenario (Preview)

danielezonca:
This PR include https://github.com/kiegroup/drools-wb/pull/1026 to fix remaining bugs (that PR will be probably closed).
This PR contains DROOLS-3369 (DMO equivalent for DMN) and DROOLS-3362 (automatically scenario template creation for DMN).
Now is possible to create a scenario and automatically have the template created with all the inputs in the GIVEN section and the decisions in the EXPECT.
If user removes a column, he can add it again using the right panel like a RULE scesim.

NOTE 1: please test it with a new scesim file or any scesim file created in master
NOTE 2: please clean code cache of `drools-wb-webapp` (`gwt*` folders, `errai` folders, `gwt:clean`) before test it
NOTE 3: DMN scenario does not support `null` values yet
NOTE 4: use FEEL to test DMN scesim. For instance string need to be quoted

@Rikkola @gitgabrio @kkufova


danielezonca:
@manstis 
Can you please squash and merge this please?

PR #1030, opened 2019-01-03
Add test for function in DRL

tomasdavidorg:
This test is moved from selenium test suite.


tomasdavidorg:
@manstis @jomarko Could you please take a look?

PR #1035, opened 2019-01-09
[DROOLS-3472] Change empty cell behavior, from null to skip it

danielezonca:
Now empty cell means to skip it (both for GIVEN/EXPECT).
Rule scesim file you can use `null` as alternative to express a null condition.
DMN scesim now just skip cell if empty or use FEEL unary expression evaluation if not empty

@Rikkola @gitgabrio @kkufova


Rikkola:
@danielezonca  I turst your vision. Just bringing up all the issues I can think of. There are some complex situation that need to be solved and not sure how. For example how to make sure static functions were called with the right values and how to mock services/functions that query for data from outside of the kbase.

danielezonca:
@manstis 
Could you please squash and merge?

PR #1037, opened 2019-01-10
DROOLS-2623: Stabilize xls-dtables tests

jomarko:
Replacement of #870 

@manstis would you mind to have a look?

I find out the problem is caused by `DecisionTableXLSServiceImplCDITest`. There I set the system property `drools.dateformat`. I debugged hours today, but didn't find out, why the system property from that test is propagated also to other tests, even when I do clean up in `@AfterClass` method.

Due to the time I have spent already with the issue I prefer this alternative, much less appropriate, solution.

Please let me know what do you think.


manstis:
@jomarko It's plausible that `drools.dateformat` is propagated to other tests on Jenkins as tests are ran in parallel (and I suspect in the same JVM so the System Property would carry across)... this PR looks good to me.

jomarko:
@manstis thank you, I will wait jenkins finishes build of this PR, when green will ask you for merge and close the original one.

jomarko:
@manstis , please merge.

PR #1042, opened 2019-01-21
DROOLS-3175: Add invoke context menu keyboard operation

jomarko:
Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/607
- https://github.com/kiegroup/kie-wb-common/pull/2408


kiereleaseuser2:
Build finished. No test results found.

jomarko:
jenkins please retest this

jomarko:

jomarko:

jomarko:
@gitgabrio I think we both understand the issue and have same goal to reuse the existing logic from two different places.

However I do not think we can do it in way you are proposing with static methods. Here are my thoughts:
- The existing logic we want to reuse is implemented using **@Dependent**  context menu beans
- These context menu instances are accessible at two places. First one is **ScenarioSimulationGridPanelClickHandlerProducer*. Second one is **ScenarioSimulationGridPanelClickHandler**.
- These two classes currently can call **context menus logic** via static methods as they hold **context menu beans** which can be passed to any static method anywhere.
- Now we want to add third place that holds **context menu beans** inside so from this third place we could call any static method
- This third place is **ScenarioGrid**
- This grid is aware of **ScenarioSimulationGridPanelClickHandler** and I cold add methods in **ScenarioSimulationGridPanelClickHandler** to expose **context menu beans** to the **ScenarioGrid**, however this seems to me the inappropriate way of getting dependencies.


If you are proposing scenario grid should obtain references to **context menu beans** via its click handler, I can go this was, however it is non intuitive for me.

jomarko:
@kkufova please have a look.

jomarko:
@kkufova conflicts fixed

jomarko:
jenkins please retest this

jomarko:
@gitgabrio Hi, after last week with intensive development, I would like to close this as soon as possible to be able to work on follow-up tasks. Could you please provide your review here? I think you are the last person I miss the review from. Thank you!

jomarko:
@kkufova please, may I ask you again for review?

kkufova:
I checked the PR again and everything still works well.

I noticed that we have a few conflicts here, so please let me know whether an additional review is needed, but I don't think so.

kkufova:
Still works well! :+1:

jomarko:
@manstis please merge.

manstis:
@jomarko Don't we need @gitgabrio to approve too?

gitgabrio:
@kkufova 
Did you test any possible keyboard interaction? Also, with collection editor, what happen if you open it?
Are there any regression on that?

gitgabrio:
@manstis @jomarko 
Just to make it clear. Beside the overall design (you know perfectly well what I think of it) my concern is due to the fact that with the previous keyboard PR an unexpected behavior and a "functionally" duplicated code has been introduced, and I had to fix them on-the-fly while working at other PR. Now, I would like to avoid the same this time, so before merging it I want to be sure the same won't happen again.

kkufova:
@gitgabrio, there are no regressions. The collection editor is correctly opened. The keyboard navigation works well in both rule-based and DMN-based Test Scenarios.

jomarko:
@gitgabrio all parts of the ensemble rebased and jenkins builds finished. Please have a look when you have a moment.

PR #1045, opened 2019-01-28
Drools 3485

gitgabrio:
@kkufova @manstis @danielezonca  @jomarko 
Scope of this PR is to implement the "Collection Editor" popup.
Such popup should follow UX guidance published here https://issues.jboss.org/browse/DROOLS-3098.


kkufova:
I took a first quick look and I must say that it works quite well! However, I managed to find a few bugs and I have a few questions and notes.

1. (bug) First of all, I noticed that this PR introduces a weird cell selection regression; if you have a table with more rows (for example three) and you select a header cell, a table cell is automatically selected too. This is valid for all header cells (GIVEN/EXPECT, instances, and properties).

bug1

2. (question) I noticed that when creating a new items in a list, data type validation is not implemented (it is possible to enter a string into a boolean field). Will the data type validation be implemented within this PR? Will it be implemented at all?

3. (bug) Once you create a list and then remove all the items from the list, the corresponding cell does not go back to being completely empty (with a placeholder _Insert value_), but it contains `[]` symbols. This is correct; empty list is a valid expression, but this makes it impossible to make the cell completely empty again. It must be possible to be able to -- at any point -- enter an empty list or leave an empty cell, in case you want to skip the cell.

4. (note) I noticed that it is not possible to scroll automatically in the modal window, the only way to scroll is manually operate the scrollbar. This is not a huge problem, but it is very irritating if the list contains a lot of items and you want to create new ones. Have you noticed this behavior?

5. (bug) If you open a non-empty list from a cell, click **Cancel**, open an empty cell below and click **Add new item**, the whole list from the first cell is inserted. Check this screencast: http://file.emea.redhat.com/kkufova/screencast-30-01-19-1.webm

6. (bug) If you want to edit a string inside the modal window (for example a book name) and you use keyboard arrows, instead of moving in the text box with the book name, you are actually selecting different cells in the table under the modal window.

7. (bug) If you insert a map into a cell, then open the map modal window again, try to edit one of the items and then click **Cancel**, an unexpected error is thrown:

> Uncaught exception: Exception caught: (TypeError) : Cannot read property 'Od' of undefined Caused by: (TypeError) : Cannot read property 'Od' of undefined

8. (bug) It is not possible to remove an item from a modal map window; an unexpected error is thrown:

> Uncaught exception: Exception caught: (TypeError) : Cannot read property 'Od' of undefined Caused by: (TypeError) : Cannot read property 'Od' of undefined

And this is it from my side. I'm sure I haven't discovered all issues yet, but I think this is enough for the initial review. :-)

gitgabrio:
@kkufova @danielezonca 
> 1. (bug) First of all, I noticed that this PR introduces a weird cell selection regression; if you have a table with more rows (for example three) and you select a header cell, a table cell is automatically selected too. This is valid for all header cells (GIVEN/EXPECT, instances, and properties).
_FIXED_

> 2. (question) I noticed that when creating a new items in a list, data type validation is not implemented (it is possible to enter a string into a boolean field). Will the data type validation be implemented within this PR? Will it be implemented at all?
_Will be implemented, not in this PR._

> 3. (bug) Once you create a list and then remove all the items from the list, the corresponding cell does not go back to being completely empty (with a placeholder _Insert value_), but it contains `[]` symbols. This is correct; empty list is a valid expression, but this makes it impossible to make the cell completely empty again. It must be possible to be able to -- at any point -- enter an empty list or leave an empty cell, in case you want to skip the cell.
_Agree, but ti would require a UX interaction; for the moment being ignore this and mark as future improvement (new jira ticket). Btw, this is not a *bug* (from my point of view, a missing functionality, as this, is not a bug; a bug is an unexpected/wrong behavior; keep the differences between the twos help in organize/plan/schedule work/priorities otherwise everything is a "bug")_

> 4. (note) I noticed that it is not possible to scroll automatically in the modal window, the only way to scroll is manually operate the scrollbar. This is not a huge problem, but it is very irritating if the list contains a lot of items and you want to create new ones. Have you noticed this behavior?
_Yes, I know that; I'm not sure to follow your question: do you want another way of scrolling down (do not know which one)? *Or* do you think the "+Add new item" should be always visible in the bottom, so that even with long collections you do not have to scroll down to insert new one?_

> 5. (bug) If you open a non-empty list from a cell, click **Cancel**, open an empty cell below and click **Add new item**, the whole list from the first cell is inserted. Check this screencast: http://file.emea.redhat.com/kkufova/screencast-30-01-19-1.webm
_Can not reproduce: could you try again with fresh install/clean cache?_

> 6. (bug) If you want to edit a string inside the modal window (for example a book name) and you use keyboard arrows, instead of moving in the text box with the book name, you are actually selecting different cells in the table under the modal window.
_This is related to the keyboard handling. Agree is bad, but it is not a quick fix, so I want to take my time and look at this with ease, maybe together with @jomarko_ 

> 7. (bug) If you insert a map into a cell, then open the map modal window again, try to edit one of the items and then click **Cancel**, an unexpected error is thrown:
> 
> > Uncaught exception: Exception caught: (TypeError) : Cannot read property 'Od' of undefined Caused by: (TypeError) : Cannot read property 'Od' of undefined
_FIXED_

> 8. (bug) It is not possible to remove an item from a modal map window; an unexpected error is thrown:
> 
> > Uncaught exception: Exception caught: (TypeError) : Cannot read property 'Od' of undefined Caused by: (TypeError) : Cannot read property 'Od' of undefined
_FIXED_

kkufova:
1. (suggestion) I think the **Remove** button is a great idea, however, we should discuss the exact functionality. First of all, it took me a while to understand that it deletes the items from the list, but it does not close the window (I was trying it without any items, so it looked like the button does not work). I think once you click **Remove**, it would be best to close the whole modal window completely and go back to _Insert value_ state of the cell (regardless of how many items the list have: you click **Remove** and all the items from the list are removed and the window is closed). In this way, it wouldn't be necessary to click **Remove** + **Save**. Now, it is very hard to distinguish how to create an empty list and how to create an empty cell (empty list = **Save**, empty cell = **Remove** + **Save**). If the list does not contain any items, the **Remove** does not have any visible action; this is my main problem. WDYT?

2. (suggestion) Regarding the scrollbar issue; would it be possible to scroll automatically? When I want to scroll for example in my browser, I don't have to manually click on the scrollbar and move it; I can do it using the mouse's "ball" or using my touchpad. This is the main issue I was explaining in my last comment.
Regarding the **Add new item** link; I think it is alright how it is now.

3. (bug) If you add new item to the list and then click on the top of the item (either next to **books** or **Item**) in the modal window, the item from the list completely disappears. Check this screencast: http://file.emea.redhat.com/kkufova/screencast-31-01-19-1.webm.

4. (bug) In the screencast above, you can also notice that after clicking **Save**, the empty list is added into a wrong row; it is added into a row above the row where it should be. How to reproduce the issue: insert three rows into the table. Click into the second one and perform the magic of disappearing items. Click **Save**. The empty list is in the first row instead of the second one.

5. (bug) This may be related to point three (I'm not sure), but collapsing items in the modal window causes them to disappear completely. Check this screencast: http://file.emea.redhat.com/kkufova/screencast-31-01-19-2.webm.

6. (suggestion) If you are in the middle of creating an item in the list and you click **Save** (or any other button), the modal window closes and the changes are lost. I know this is a currently correct behavior, but I would prefer a warning message similar to: _You have unsaved changes. Do you wish to proceed anyway?_ WDYT?

7. (bug?) I'm not completely sure this is a "bug", but I think it should be fixed. Resizing columns in the table automatically opens the list/map modal window. Check this screencast: http://file.emea.redhat.com/kkufova/screencast-31-01-19-3.webm.

NOTE: I created a Jira for the issue described in point 6 in my last comment: https://issues.jboss.org/browse/DROOLS-3588

I think we're getting close to closing! :-)

danielezonca:
@gitgabrio 
There is a bug in `FactMapping.getGenericTypes`. This list should contains the canonical name for each generic but this is true only for java builtin types. For instance it is `java.lang.Integer` for an `Integer` but it is only `Person` for `com.Person`.

gitgabrio:
> There is a bug in `FactMapping.getGenericTypes`. This list should contains the canonical name for each generic but this is true only for java builtin types. For instance it is `java.lang.Integer` for an `Integer` but it is only `Person` for `com.Person`.

_FIXED_

gitgabrio:
jenkins please retest this

gitgabrio:
jenkins please retest this

kkufova:
I found only three issues and neither of them is a blocker. I would like to see the first one fixed, but I think the other two can wait. @gitgabrio, please let me know what you think; I would create Jiras for the unaddressed issues. I expect I will be able to approve the PR right after the fix of no. 1 and one final review.

1. (bug) Clicking on GIVEN/EXPECT selects a cell in the first row. I think clicking on these headers should not select anything.

2. (UI improvement) The size of the modal window is different based on the location from where it is invoked. See this screencast: http://file.emea.redhat.com/kkufova/screencast-05-02-19-1.webm.

3. (UI improvement) Collapsing in the modal window is not precise. This is definitely not a blocker, but something we should consider fixing at least in the future if not now. If you collapse all the items, the whole list should switch to "collapsed" too:
bug1
If you want to "uncollapse" the whole list, you need to press the button twice instead of once.

kkufova:
However, I noticed that the functionality does not work for DMN-based scenarios yet; the modal window is not opened from a cell that should contain a list. Is this expected? Does this PR only cover rule-based scenarios?

gitgabrio:
@kkufova 
For this PR I would not do other UI fix/improvement since I think it is better to merge it ASAP.

danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

gitgabrio:
> 1. (bug) Clicking on GIVEN/EXPECT selects a cell in the first row. I think clicking on these headers should not select anything.

_FIXED_

kkufova:
I'm waiting for a conflict resolution before the final review.

gitgabrio:
@kkufova  @danielezonca 
Fixed merge. Pls retest

kkufova:
The following scenario causes an unexpected error:

1. Introduce an instance alias.
2. Undo the action.
3. Try to invoke the list/map modal window.

http://file.emea.redhat.com/kkufova/screencast-06-02-19-3.webm

gitgabrio:
> The following scenario causes an unexpected error:
> 
> 1. Introduce an instance alias.
> 2. Undo the action.
> 3. Try to invoke the list/map modal window.
> 
> http://file.emea.redhat.com/kkufova/screencast-06-02-19-3.webm

_FIXED_

gitgabrio:
jenkins please retest this

gitgabrio:
@manstis 
Hi. Could you please merge this, squashing commits? Thanks!

PR #1046, opened 2019-01-29
Drools 3242 - [Scenario Simulation] Update cell (double-)click interactions

gitgabrio:
@kkufova  @danielezonca @manstis 

See  https://issues.jboss.org/browse/DROOLS-3242


gitgabrio:
@manstis 
HI. If you have time, could you take a look ? Otherwise, could you please squash commits and merge?

PR #1048, opened 2019-01-29
[DO NOT MERGE] Update the cheat sheet in Test Scenarios (Preview).

kkufova:
@danielezonca, please take a look. Just a quick fix to address the skip/null changes.


kkufova:
jenkins please retest this

kkufova:
@danielezonca, as Anna's PR is closed, I updated this one. Could you please check it one more time?

danielezonca:
I think Anna is going to create a new PR based on the approach discussed with Porcelli. She probably just closed the old one too early :)

kkufova:
@danielezonca, my understanding is that it may take a while to agree on the way how it can be done, so I thought we can have this in the meantime. :-)

PR #1049, opened 2019-01-29
[DROOLS-3357] Implement filtering right panel 

gitgabrio:

See https://issues.jboss.org/browse/DROOLS-3357


gitgabrio:
@kkufova 
Well, this is exactly the wanted behavior: hide a property already assigned to avoid inserting the same property twice. Eventually, the issue I see is when you rename the property (it should not be visible anyway). @danielezonca WDYT ?

kkufova:
@gitgabrio, I think we should preserve the original behavior, as it was working great in my opinion and the logic behind it was clear. When you select a property header cell, the property should be highlighted in the right panel. If you rename the property, the original property should still be highlighted in the right panel (no renaming here). The only difference should be that properties used in different header cells should not be in the right panel for one property cell.

gitgabrio:
@kkufova @danielezonca 
Based on Klara suggestion on slack, I'm going to check that
1) the property of the selected column will be shown (highlighted) on right panel
2) when selecting another (unassigned) column, the already assigned property(ies) will be hidden
3) hidden property should work *also* for renamed ones (if a "books" property has been renamed to "books2", for example, it should be anyway invisible during assignment of a property to another column of same instance)

gitgabrio:
@kkufova 
Shoud be fixed. Pls test again

danielezonca:
jenkins retest this please

kkufova:
@gitgabrio, the highlighting works fine, but if the right panel was not collapsed before creating a new column, it is not automatically collapsed or reloaded and it is still possible to select the same property twice. See this screencast: http://file.emea.redhat.com/kkufova/screencast-06-02-19-1.webm.

gitgabrio:
@kkufova 

> @gitgabrio, the highlighting works fine, but if the right panel was not collapsed before creating a new column, it is not automatically collapsed or reloaded and it is still possible to select the same property twice. See this screencast: http://file.emea.redhat.com/kkufova/screencast-06-02-19-1.webm.

_FIXED removing automatic selection of newly created column_

kkufova:
The following scenario still does not work: http://file.emea.redhat.com/kkufova/screencast-07-02-19-1.webm.

gitgabrio:
@kkufova 

> The following scenario still does not work: http://file.emea.redhat.com/kkufova/screencast-07-02-19-1.webm.

We realize that this fix is longer then expected, so we prefer to merge this as it is, and create a specific ticket

kkufova:
This is the newest screencast of the breaking scenario: http://file.emea.redhat.com/kkufova/screencast-08-02-19-1.webm. This should be fixed before merging.

gitgabrio:
@kkufova 

> This is the newest screencast of the breaking scenario: http://file.emea.redhat.com/kkufova/screencast-08-02-19-1.webm. This should be fixed before merging.

_FIXED_

gitgabrio:
jenkins please retest this

danielezonca:
jenkins retest this please

manstis:
This compiles and tests pass locally.

Jenkins build failed due to tidying up after the PR successfully compiled and tests passed.
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 38.807 s (Wall Clock)
[INFO] Finished at: 2019-02-08T08:13:20-05:00
[INFO] ------------------------------------------------------------------------
[drools] $ sh /home/jenkins/workspace/KIE/master/pullrequest/drools-wb-pullrequests/clean-up-script.sh
Unexpected error while executing the Upstream repositories builder for PR builds! Error while executing clean-up script, non-zero exit code!
java.lang.RuntimeException: Error while executing clean-up script, non-zero exit code!
```

PR #1050, opened 2019-01-29
[DROOLS-3555] Implement cheatsheet for DMN simulation

annabaker:
@gitgabrio @danielezonca @kkufova 

who to @ regarding automatic translation on the larger properties for rule/dmn cheatsheets?

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
@@ -27,6 +27,47 @@ hideRightPanel=Hide Right Panel
 testTools=Test Tools
 testEditor=Test Editor
 scenarioCheatSheet=Scenario Cheatsheet
+dmnCheatSheet=\
+  <p>To start off, right-click a cell in the table and use the contextual menus to modify the generated table as required.</p>\
+  <p>Before defining the individual test scenarios, verify the instance and field mappings for each column. If any \
+  changes are necessary, use the <b>Test Editor</b> tab to update the mapping.</p>\
+  <p>Now, define the test scenarios. Each cell of the table must contain a valid expression.</p>\
+  <p>Values in the GIVEN part of the table are evaluated as literal expressions, whereas values in the EXPECT section are \
+  evaluated as unary expressions. The correctness of these unary tests is verified using the results from a DMN evaluation.</p>\
+  <p>DMN-based scenarios use the FEEL expression language. Supported data types are:</p>\
+  <ul>\
+  <li><i>numbers</i> and <i>strings</i> (strings must be delimited by quotation marks), for example <tt>"John Doe"</tt> or <tt>""</tt></li>\
+  <li><i>boolean values</i> (<tt>true</tt>, <tt>false</tt>, and <tt>null</tt>)</li>\
+  <li><i>dates</i> and <i>time</i>, for example <tt>date("2019-05-13")</tt> or <tt>time("14:10:00+02:00")</tt></li>\
+  <li><i>functions</i></li>\
+  <li><i>contexts</i>, for example <tt>{x : 5, y : 3}</tt></li>\
+  <li><i>ranges</i> and <i>lists</i>, for example <tt>[1 .. 10]</tt> or <tt>[2, 3, 4, 5]</tt></li>\
+  </ul>
+ruleCheatSheet=\
+  <p>To start off, right-click a cell in the table and use the contextual menus to insert new columns and rows.</p>\
+  <p>Before defining the individual test scenarios, set the instance and field mappings for each column. Select a header \
+  cell and in the <b>Test Editor</b> tab, choose the required field. Click <b>Add</b> to set the mapping.</p>\
+  <p>Now, define the test scenarios. Each cell of the table must contain a valid expression.</p>\
+  <p>In rule-based scenarios, the following expressions are supported:</p>\
+  <ul>\
+  <li><tt>=</tt> (or no symbol) specifies equality of a value. This is the default operator of each column and the only \
+  operator that is supported in the GIVEN section of the table.</li>\
+  <li><tt>!</tt>, <tt>!=</tt>, <tt>&lt;&gt;</tt> specify inequality of a value. This operator can be combined with other operators.</li>\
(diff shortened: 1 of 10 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
```diff
@@ -27,6 +27,43 @@ hideRightPanel=Hide Right Panel
 testTools=Test Tools
 testEditor=Test Editor
 scenarioCheatSheet=Scenario Cheatsheet
+dmnCheatSheet=<p><h4>DMN-based scenarios</h4>\
+  <p>DMN-based scenarios use the FEEL expression language. Supported data types are:</p>\
```
danielezonca: Before list the supported data types please add a paragraph with
`The text in a GIVEN cell is evaluated as literal expression and in the EXPECT as unary expression. This is because it verifies if the unary test is valid using the result from DMN evaluation`
or something similar

Rikkola:
@danielezonca I think the HTML in the property file is risky, since translations might then break the UI.

To stay sane there might have to be some sort of middle ground that keeps the tags for bold words in for example.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
kkufova: I would rephrase it to the following:

> Values in the GIVEN part of the table are evaluated as literal expressions, whereas values in the EXPECT section are evaluated as unary expressions. The correctness of these unary tests is verified using the results from a DMN evaluation.

gitgabrio:
@Rikkola @danielezonca 
I also thought that put html inside properties would break automatic translation. On the same time, we can not simply split on multiple lines/paragraphs, because different languages have completely different logical constructions. @Rikkola is there not any other situation inside workbench with such requirement?

Rikkola:
@gitgabrio I tried to think of a situation, but no. We don't at the moment. We use Zanata for translations and I believe it is just key <-> value pairs that we can work with.

Only place closest to this is what I have in Verifier reporting, but there I just split based on paragraphs. It works for most of the cases and haven't heard of any issues with the languages we support.

annabaker:
> The text should be as follows:
> 
>     * rule-based scenarios: http://file.emea.redhat.com/kkufova/1050-rule-based.txt
> 
>     * DMN-based scenarios: http://file.emea.redhat.com/kkufova/1050-dmn-based.txt
> 
> 
> I incorporated the required changes proposed in #1048 and by @danielezonca. I also changed the introductory text for DMN-based scenarios. This should be the final version.

I've modified the .properties file to reflect the final versions

annabaker:
@danielezonca @gitgabrio @kkufova 

As discussed, leaving the HTML in the properties file for now and will modify the approach maybe using HTML templates in another sprint.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
```diff
@@ -27,6 +27,49 @@ hideRightPanel=Hide Right Panel
 testTools=Test Tools
 testEditor=Test Editor
 scenarioCheatSheet=Scenario Cheatsheet
+dmnCheatSheet=<p>\
```
kkufova: Do we need this tag? I'm not sure, but I think it causes an unnecessary vertical space.

bug1

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
```diff
@@ -27,6 +27,49 @@ hideRightPanel=Hide Right Panel
...
+  <ul>\
+  <li><i>numbers</i> and <i>strings</i> (strings must be delimited by quotation marks), for example <tt>"John Doe"</tt> or <tt>""</tt></li>\
+  <li><i>boolean values</i> (<tt>true</tt>, <tt>false</tt>, and <tt>null</tt>)</li>\
+  <li><i>dates</i> and <i>time</i>, for example <tt>date("2019-05-13")</tt> or <tt>time("14:10:00+02:00")</tt></li>\
+  <li><i>functions</i></li>\
+  <li><i>contexts</i>, for example <tt>{x : 5, y : 3}</tt></li>\
+  <li><i>ranges</i> and <i>lists</i>, for example <tt>[1 .. 10]</tt> or <tt>[2, 3, 4, 5]</tt></li>\
+  </ul>\
+  </p>
+ruleCheatSheet=<p>\
```
kkufova: Do we need this tag?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
porcelli: this is definitely not recommended way. the i18n mechanism is not meant to have HTML on it.

it may break and eventually will break.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
annabaker: I've removed it :)

annabaker: Also modified "cheatsheet" to be more consistent (cheatSheet / *CheatSheet)

PR #1062, opened 2019-02-07
[DROOLS-3600] Add i18n support to Test Scenario cheatsheet

annabaker:
@kkufova @danielezonca @gitgabrio


kkufova:
To be honest, I'm not really sure this is the way to go, especially in case of such long texts.

@danielezonca, @gitgabrio, @Rikkola, @manstis, what do you think?

gitgabrio:
@kkufova 
Completely agree with you - and I think to remember to already have made a slight comment on that somewhere :smile: 
Anyway, it is what has been suggested to @annabaker , that's why she did so.

Rikkola:
The best way would be to have full templates for each language, but we can not support that in the i18n tooling we have.
The second best way would be to have each line/paragraph in the i18n properties file and still leave the text formatting in as HTML tags. This is because each language has the highlighted words in the different order. There is still a risk a bad i18n file breaks the UI, but the risk is reduced.

kkufova:
The cheat sheet for DMN is not precise. On the left: new version, on the right: the original version.

bug1

This is exactly the reason why I don't like this approach; it is very hard to discover issues like these, there's no chance I would find it in the code. I had to manually compare the original and the new text in Business Central.

danielezonca:
@kkufova 
From my point of view our localization mechanism is designed for short text with no formatting. That's the reason why in this use case it doesn't fit really well but I don't see any alternative.
The error you reported is related to a wrong concatenation in the java code that need to be fixed like many other possible errors in string manipulation :)

annabaker:
@kkufova how are you comparing those versions?  when i view the cheatsheet on the UI, i'm not seeing those errors.
screenshot from 2019-02-25 19-08-42

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelViewImpl.java
```diff
@@ -219,11 +291,55 @@ protected void setContainersDisabledStatus(boolean disabled) {
...
+                + " " + "<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet8() + "</i>" + " " + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet9() + " "
+                + "<tt>\"John Doe\"</tt>" + " " + ScenarioSimulationEditorConstants.INSTANCE.or() + " " + "<tt>\"\"</tt>");
+        dmnCheatSheetList2.setInnerHTML("<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet10() + "</i> (<tt>true</tt>, <tt>false</tt>, "
+                + ScenarioSimulationEditorConstants.INSTANCE.and() + " " + "<tt>null</tt>)");
+        dmnCheatSheetList3.setInnerHTML("<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet11() + "</i>" + " "
+                                                + ScenarioSimulationEditorConstants.INSTANCE.and() + " " + "<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet12() + "</i>"
+                                                + ScenarioSimulationEditorConstants.INSTANCE.forExample() + " " + "<tt>date(\"2019-05-13\")</tt>" + ScenarioSimulationEditorConstants.INSTANCE.or() + " "
```
danielezonca: @annabaker 
missing space here
`"<tt>date(\"2019-05-13\")</tt>" + ScenarioSimulationEditorConstants`

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelViewImpl.java
```diff
@@ -219,11 +291,55 @@ protected void setContainersDisabledStatus(boolean disabled) {
...
+                                                + "<tt>time(\"14:10:00+02:00\")</tt>");
+        dmnCheatSheetList4.setInnerHTML("<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet13() + "</i>");
+        dmnCheatSheetList5.setInnerHTML("<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet14() + "</i>" + ScenarioSimulationEditorConstants.INSTANCE.forExample() +
+                                        "                    + \" \" + \"<tt>{x : 5, y : 3}</tt>");
+        dmnCheatSheetList6.setInnerHTML("<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet15() + "</i>"
+                                                + " " + ScenarioSimulationEditorConstants.INSTANCE.and() + " " + "<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet16() + "</i>"
+                                                + ScenarioSimulationEditorConstants.INSTANCE.forExample() + "<tt>[1 .. 10]</tt> or <tt>[2, 3, 4, 5]</tt>");
```
danielezonca: @annabaker 
Missing space
`ample() + "<tt>[1`

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/RightPanelViewImpl.java
```diff
@@ -219,11 +291,55 @@ protected void setContainersDisabledStatus(boolean disabled) {
...
+                + ScenarioSimulationEditorConstants.INSTANCE.and() + " " + "<tt>null</tt>)");
+        dmnCheatSheetList3.setInnerHTML("<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet11() + "</i>" + " "
+                                                + ScenarioSimulationEditorConstants.INSTANCE.and() + " " + "<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet12() + "</i>"
+                                                + ScenarioSimulationEditorConstants.INSTANCE.forExample() + " " + "<tt>date(\"2019-05-13\")</tt>" + ScenarioSimulationEditorConstants.INSTANCE.or() + " "
+                                                + "<tt>time(\"14:10:00+02:00\")</tt>");
+        dmnCheatSheetList4.setInnerHTML("<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet13() + "</i>");
+        dmnCheatSheetList5.setInnerHTML("<i>" + ScenarioSimulationEditorConstants.INSTANCE.dmnCheatSheet14() + "</i>" + ScenarioSimulationEditorConstants.INSTANCE.forExample() +
+                                        "                    + \" \" + \"<tt>{x : 5, y : 3}</tt>");
```
danielezonca: @annabaker 
Wrong escaping
`"                    + \" \" + \"`

kkufova:
@annabaker, this is what I see:

bug1

annabaker:
@kkufova could you tell me if it looks okay now?  i'm not sure why the UI was loading properly for me in my screenshot before.

annabaker:
@manstis could you please merge?  thank you

PR #1065, opened 2019-02-08
[DROOLS-3624] In a scenario it should be possible to create only one …

annabaker:
…item at a time in a collection

@kkufova @danielezonca @gitgabrio


kkufova:
@gitgabrio, @annabaker, it is still possible to edit multiple items + simultaneously be creating a completely new item. As discussed with @gitgabrio online, this should not be possible either.

See the following screencast for more information.
http://file.emea.redhat.com/kkufova/screencast-11-02-19-1.webm

annabaker:
@kkufova @gitgabrio 

i've fixed this -- please take a look :)

kkufova:
@annabaker, I discovered two issues:

1. Firstly, if you delete an item in the middle of editing it, subsequently it is not possible to edit any other item and add a new one: http://file.emea.redhat.com/kkufova/screencast-13-02-19-1.webm. The same problem also occurs in the map modal window.

2. Secondly, when you try to add a new item and then cancel the operation, the possibility to edit any of the existing items is restricted: http://file.emea.redhat.com/kkufova/screencast-13-02-19-2.webm.

gitgabrio:
@kkufova 
I fixed that behavior. Anyway, I find much more consistent to *also* disable the "delete" (trash) button when starting editing/adding new item.

@annabaker
Great job! I like how you did your modifications! Anyway, I would like to push my modifications to your branch, but it seems I have no permission: could you give me them? Thanks!

gitgabrio:
Fine for me

@manstis 

PR #1067, opened 2019-02-11
[DROOLS-3630] Fix NPE on right panel complex nested objects

danielezonca:
@kkufova @Rikkola @gitgabrio


kkufova:
@danielezonca, I have three comments:

1. First, do we want to allow these nested objects to be selected as column types? Currently, it is possible; my understanding was that this option should be restricted.

bug1

2. Secondly, the issue [DROOLS-3621](https://issues.jboss.org/browse/DROOLS-3621) is still present.

bug2

3. Finally, I discovered another unexpected error. See this screencast: http://file.emea.redhat.com/kkufova/screencast-12-02-19-1.webm.

Please let me know what you think.

danielezonca:
@kkufova 
1. No, I prefer to keep this PR for blockers only
2. Yes, this fix doesn't cover that bug
3. Fixed (please test also DMN scenario because this fix involve code that is shared)

danielezonca:
@manstis 

PR #1069, opened 2019-02-13
Drools 3621 Right-panel highlighting in Test Scenarios does not work for all-level nested properties

gitgabrio:

See https://issues.jboss.org/browse/DROOLS-3621


gitgabrio:
@manstis 

PR #1070, opened 2019-02-14
[DROOLS-3654] Replacing build-helper-maven-plugin with gwt-helper-maven plugin

gitgabrio:
@ederign @porcelli @manstis @Rikkola @danielezonca @etirelli 

The main scope of this PR is to allow editing/debugging/hot-reload client-code (GWT) of modules external of drools-wb while working with drools-webapp. This should be configurable on a per-developer basis, modifying personal .m2/settings.xml, without touching the versioned pom.xml
The current solution is based on the build-helper-maven-plugin, but it has two limitations
1) it is needed to statically define every external source inside the plugin configuration
2) it is not possible to override such configuration with the user' .m2/settings.xml.

The gwt-help-maven-plugin (https://github.com/gitgabrio/gwt-helper-maven-plugin) has been designed to overcome this limitations, following suggestions from @ederign:

1) it uses a pattern matching to (recursively) find external sources
2) it uses a list of directories to define the path to read for external sources
3) it may use such pattern/directories defined as properties in the pom, and since properties are overridable with m2./settings.xml profile, it is possible for each user to define its own profile and its own pattern, to match the setting s/he has/needs on its machine.

For the moment being I've developed that plugin in my spare time, and being an "experimental" feature I avoided any reference to RedHat. If it is accepted, I would like to remove all personal reference and use standard RedHat ones. Moreover, it is currently deployed to Sonatype' Nexus, and I am not sure this follows our "policies".
I would also be very happy if someone else would be interested in contributing to it.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-client/src/main/java/org/drools/workbench/screens/guided/template/client/editor/TemplateDataTableWidget.java
+++ b/drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-client/src/main/java/org/drools/workbench/screens/guided/template/client/editor/TemplateDataTableWidget.java
@@ -148,10 +148,12 @@ public class TemplateDataTableWidget extends Composite
                             height);
     }
 
+    @Override
     public void onDeleteRow(DeleteRowEvent event) {
         model.removeRow(event.getIndex());
     }
 
+    @Override
     public void onCopyRows(CopyRowsEvent event) {
         copiedRows.clear();
         for (Integer iRow : event.getRowIndexes()) {
@@ -160,6 +162,7 @@ public class TemplateDataTableWidget extends Composite
         }
     }
 
+    @Override
     public void onPasteRows(PasteRowsEvent event) {
         if (copiedRows == null || copiedRows.size() == 0) {
             return;
@@ -176,17 +179,20 @@ public class TemplateDataTableWidget extends Composite
         }
     }
 
+    @Override
     public void onInsertRow(InsertRowEvent event) {
         List<String> data = cellValueFactory.makeRowData();
-        model.addRow(Integer.toString(event.getIndex()),
+        model.addRow(event.getIndex(),
                      data.toArray(new String[data.size()]));
     }
 
+    @Override
     public void onAppendRow(AppendRowEvent event) {
         List<String> data = cellValueFactory.makeRowData();
         model.addRow(data.toArray(new String[data.size()]));
     }
 
+    @Override
     public void onUpdateModel(UpdateModelEvent event) {
 
         //Copy data into the underlying model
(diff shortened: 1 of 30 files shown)


gitgabrio:
@manstis 
This is exactly one of my concerns. If possible, I would like to move it under RH umbrella *before* merging. On the other hand, before moving that plugin to RH I think it should at least be tested a little bit more thoroughly. I did some tests, and @danielezonca used it once some days ago, but I think it should be tested _on real cases_ - I mean by you guys who work daily with multiple/crossed project setup. @etirelli WDYT?

etirelli:
@gitgabrio there is no reason not to move your code under kiegroup. We have repositories here for research and testing of new components, so my suggestion: go ahead and add your plugin here, and invite the team to test it. @manstis 's suggestion of having a profile for it so that people can switch between them during the trial phase is a good one IMO.

gitgabrio:
@ederign @porcelli @manstis @Rikkola @danielezonca @etirelli
Thanks for your appreciation. Now, what does "move under pkiegroup" exactly means?
My idea/doubts are:
1) I have to change the package name: could someone of you tell me what is the correct one?
2) I think I will have to add the "Red Hat" license header to all files of the plugin (the same I put inside drools-wb code): correct?
3) do I have to add a specific "License" file
4) which repository I have to use to move the sources?
5)  where/how do I deploy the newly built artifact?

Asking here because I do not know if there is a documentation for all that somewhere, nor do I know how/who to ask that.
Thanks.

manstis:
@gitgabrio IMO, in response to your questions:-

> I have to change the package name: could someone of you tell me what is the correct one?

This would depend on the module; but `org.kie.maven.gwthelper.plugin` seems reasonable.

> I think I will have to add the "Red Hat" license header to all files of the plugin (the same I put inside drools-wb code): correct?

Correct

> do I have to add a specific "License" file ?

Not as far as I know.

> which repository I have to use to move the sources to?

I think https://github.com/kiegroup/droolsjbpm-integration would be best (and fits where other Maven plugins and _container_ helpers are stored).

> where/how do I deploy the newly built artifact?

Presuming you also update the `pom.xml` to inherit from `kiegroup` parent your plugin would fall into our existing build/deploy/publish mechanism. You'd probably need to declare your plugin in one of our `bom`'s [here](https://github.com/kiegroup/droolsjbpm-build-bootstrap) (IDK which is most suitable perhaps, if nobody else comments, @mbiarnes can assist).

gitgabrio:
@manstis 
Thx! I'll be back as soon as I do all the above!

gitgabrio:
@ederign @porcelli @manstis @danielezonca 

I created two PRs to
1) include plugin inside droolsjbpm-integration (https://github.com/kiegroup/droolsjbpm-integration/pull/1741)
2) add the pluginn version definition to parent pom (https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/932)

For some reason the enforcer check fails on java validation during plugin build, so until is not fixed this can't be merged. Could someone tell me how to fix the issue I reported here https://github.com/kiegroup/droolsjbpm-integration/pull/1741 ? Thanks!

gitgabrio:
@Rikkola 
My mistake.

mbiarnes:
Please merge these together:

https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/932
https://github.com/kiegroup/droolsjbpm-integration/pull/1741
https://github.com/kiegroup/drools-wb/pull/1070

mbiarnes:
Jenkins please retest this

mbiarnes:
Jenkins please retest this

mbiarnes:
Jenkins please retest this

gitgabrio:
@kiegroup/gatekeepers

Could you please merge this, squashing commits? Thanks!

manstis:
@gitgabrio We need all PRs to be approved and green before considering merge.

manstis:
@mariofusco This should never have been merged; it lacks ACKs and includes files that have nothing to do with the proposed changes... Can you please revert?

PR #1078, opened 2019-02-21
Drools 3619  More property headers can have the same alias

gitgabrio:

https://issues.jboss.org/browse/DROOLS-3619


kkufova:
@gitgabrio, it does not work.

bug1

gitgabrio:
@kkufova 
Fixed.

kkufova:
What we decided:

- it should not be possible to use the same alias for two properties within the same instance;
- it should be possible to use the same alias for two properties of two different instances;
- if a property is renamed in GIVEN, the same property must be renamed in EXPECT and vice versa;
- it should not be possible to use an alias for a property in GIVEN (`Book.name`) and the same alias for a different property in EXPECT (`Book.copies`)
-- this means that aliases are *unique* within one instance, regardless of position (GIVEN or EXPECT).

kkufova:
@gitgabrio, the last requirement is not satisfied:

> it should not be possible to use an alias for a property in GIVEN (`Book.name`) and the same alias for a different property in EXPECT (`Book.copies`)

gitgabrio:
1) GIVEN - Book - name -> rename to "huhu"
2) EXPECT - Book - numberOfCopies -> rename to "huhu"

it should not be possible

gitgabrio:
@kkufova 
Klara please retest this ( :smile: )

kkufova:
@manstis, could you please merge this?

PR #1079, opened 2019-02-21
[DROOLS-3588] Avoid changing cell selection with arrow keys when the collection popup is shown

gitgabrio:
@kkufova  @Rikkola @danielezonca 

https://issues.jboss.org/browse/DROOLS-3588


gitgabrio:
@manstis

PR #1080, opened 2019-02-26
[RHPAM-1923] Fix ScenarioJunitActivator creation

danielezonca:
@kkufova @Rikkola @gitgabrio 

How to reproduce:
Import one of the project from https://github.com/kiegroup/kie-wb-playground.git and create a scesim file

Note 1: this PR create a specific package for the ScenarioJunitActivator (`test`) so after the first scesim file, user will have this package in the package list.
Note 2: this PR also remove the ScenarioJunitActivator file from the old position if exists. The old position is the package equivalent to the groupId so usually `com.myspace`

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
@@ -122,6 +122,7 @@ public class ScenarioSimulationServiceImpl
     private Properties props = new Properties();
 
     private static final String KIE_VERSION = "kie.version";
+    private static final String junitActivatorPackageName = "test";
 
     {
         String propertyFileName = "kie.properties";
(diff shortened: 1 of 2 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
```diff
@@ -122,6 +122,7 @@
     private Properties props = new Properties();
 
     private static final String KIE_VERSION = "kie.version";
+    private static final String junitActivatorPackageName = "test";
```
Do not use "test" as package name, since it is confusing and it is a word too easily used elsewhere.
Whenever there is the need to create an internal-use only keyword, is always better to use something really specific.

danielezonca: What about `testscenario`?

danielezonca: Done

kkufova:
@danielezonca, @jhrcek, @adrielparedes 

Importing sample projects using the Git link `https://github.com/kiegroup/kie-wb-playground` is not valid and recommended. Importing the sample projects directly in the UI works well and there's no issue.

However, there's probably a problem with packages in the `Mortgage_Process` and `Evaluation_Process` repositories. When trying to create a Test Scenario asset here, the following error appears, but the asset is created anyway:

> Unable to complete your request. The following exception occurred: Impossible to retrieve package information from path: default://master@MySpace/example-Mortgage_Process/src/test/resources/com/myspace/mortgage_app.

@jhrcek, @adrielparedes, can this be related to the recent changes done to `kie-wb-playground`?

Anyway, I cannot reproduce the bug. As @danielezonca said, it should also be enough to create a project with a custom groupId, but I don't see any issue here either. When I completely change the groupId manually after the project is already created, the same error message as above appears, but I think this is a different issue than the one originally reported.

> Unable to complete your request. The following exception occurred: Impossible to retrieve package information from path: default://master@MySpace/TestProject2/src/test/resources/klara/is/great.

@danielezonca, WDYT?

jhrcek:
> @jhrcek, @adrielparedes, can this be related to the recent changes done to `kie-wb-playground`?

Those changes in kie-wb-playground I was talking about aare actually not recent. The change which I believe made kie-wb-playground into "not-valid-business-central-project-repo" was [this](https://github.com/kiegroup/kie-wb-playground/pull/64). After this PR maven build is required to build valid workbench projects (through adding mandatory kie dependencies via xslt transformation) out of that repo. The build result is a zip with workbench projects which is included in business-central.war . So my advice would be not to try importing the projects from that repo directly, but use the built-in "Try samples" feature in business-central instead.

danielezonca:
@manstis 

manstis:
@danielezonca Ah.. this is the first of the "... this too" (on the 7.18.x PR!)

PR #1085, opened 2019-03-01
DROOLS-3724 : Defining non default Kie Base and Kie Session results in a hang in the Test Scenario Runner

Rikkola:
https://issues.jboss.org/browse/DROOLS-3724


manstis:
@Rikkola Would you mind explaining what the issue was and _why_ this PR fixes it?

Looking at the changes you appear to have simply changed the parameter passed to the Audit View.

PR #1086, opened 2019-03-01
[7.18.x] DROOLS-3724 : Defining non default Kie Base and Kie Session results in a hang in the Test Scenario Runner

Rikkola:
https://issues.jboss.org/browse/DROOLS-3724

@manstis @kkufova  Could you two please take a look.


manstis:
@Rikkola Would you mind explaining what the issue was and _why_ this PR fixes it?

Looking at the changes you appear to have simply changed the parameter passed to the Audit View.

Rikkola:
The rule name array is null and that causes the NPE. So the fix is to check if the array is null and use an empty list in that case.

Rikkola:
The issue also only happens if you only have Stateless ksession set up and when you don't specify the session before you run it. 
@kkufova best way to reproduce the fixed issue:
1. Make a new project
2. Make a new kbase and ksession of the type Stateless
3. Make a new legacy test scenario and run it.

Rikkola:
@kiegroup/gatekeepers if you could please merge this.

Rikkola:
Yep it was fine

PR #1087, opened 2019-03-06
[DROOLS-2680] Run a single row scenario test

danielezonca:
@kkufova @gitgabrio @Rikkola @yesamer

- Right click on row
- Choose `Run single scenario`

https://issues.jboss.org/browse/DROOLS-2680

It contains also a minor fix on `null` evaluation with `;` expressions


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
```diff
@@ -90,6 +90,7 @@ insertRowAbove=Insert row above
 insertRowBelow=Insert row below
 deleteRow=Delete row
 duplicateRow=Duplicate row
+runSingleScenario=Run single scenario
```
kkufova: Only one thing: can we rename it to just _Run scenario_? It's simpler and I think sufficient.

danielezonca:
@manstis 

PR #1091, opened 2019-03-11
Drools 3608

gitgabrio:
See https://issues.jboss.org/browse/DROOLS-3608

Please check for regression. For the moment being there is not any modification client-side, so editing of nested objects is not possible.


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/pom.xml
```diff
@@ -157,9 +157,27 @@
       <artifactId>kie-dmn-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-dmn-core</artifactId>
+      <classifier>tests</classifier>
```
danielezonca: Why are you adding this as classifier? This project already declare `kie-dmn-core` as test dependency

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/pom.xml
```diff
@@ -157,9 +157,27 @@
...
+    </dependency>
 
   </dependencies>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-dmn-core</artifactId>
+        <version>${version.org.kie}</version>
```
danielezonca: Same comment as above, this should not be needed

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
```diff
@@ -46,26 +47,21 @@
...
     @Override
-    public FactModelTuple retrieveType(Path path, String dmnPath) {
+    public FactModelTuple retrieveFactModelTuple(Path path, String dmnPath) throws WrongDMNTypeException {
         DMNModel dmnModel = getDMNModel(path, dmnPath);
         SortedMap<String, FactModelTree> visibleFacts = new TreeMap<>();
         SortedMap<String, FactModelTree> hiddenFacts = new TreeMap<>();
-
         ErrorHolder errorHolder = new ErrorHolder();
-
+        Map<String, List<String>> genericTypeInfoMap = new HashMap<>();
```
danielezonca: Are you sure this map should be the same for all the `FactModelTree`s?
If I have understood correctly, this means that if you have two data objects with a list with the same field name (i.e. `Person.names` and `Book.names`) they will collapse in a single entry of the map

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
```diff
@@ -85,103 +81,190 @@ public DMNRuntime getDMNRuntime(Path path) {
...
     }
 
-    protected void checkTypeSupport(DMNType type, ErrorHolder errorHolder, String path) {
+    /**
+     * Check the given <code>DMNType</code> to eventually detect and add errors to given <code>ErrorHolder</code>
+     * @param type
+     * @param errorHolder
+     * @param fullPropertyPath
+     */
+    protected void checkType(DMNType type, ErrorHolder errorHolder, String fullPropertyPath) {
```
danielezonca: Personally I prefer `checkTypeSupport` because this method verify if there is any field/object with a not supported structure. I think `checkType` is a bit too generic

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/DMNDataManagementStrategy.java
```diff
@@ -57,9 +59,13 @@ public void populateRightPanel(final RightPanelView.Presenter rightPanelPresente
...
         }
         else {
-            dmnTypeService.call(getSuccessCallback(rightPanelPresenter, scenarioGridModel),
-                                getErrorCallback(rightPanelPresenter))
-                    .retrieveType(currentPath, dmnFilePath);
+            try {
+                dmnTypeService.call(getSuccessCallback(rightPanelPresenter, scenarioGridModel),
+                                    getErrorCallback(rightPanelPresenter))
+                        .retrieveFactModelTuple(currentPath, dmnFilePath);
+            } catch (Exception e) {
```
danielezonca: This is a remote method call, I don't think the exception is actually fired. I expect it will be wrapped and just invoked `getErrorCallback` in case of error

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/test/java/org/drools/workbench/screens/scenariosimulation/backend/server/AbstractDMNTest.java
```diff
@@ -16,81 +16,322 @@
...
-        when(inputDataNodeSimpleMock.getName()).thenReturn(SIMPLE_TYPE_NAME);
+        decisionNodes = new HashSet<>();
+        DecisionNode decisionNodeSimpleNoCollection = getDecisionNode(simpleTypeNoCollection, SIMPLE_DECISION_DATA_NAME);
+        decisionNodes.add(decisionNodeSimpleNoCollection);
+        DecisionNode decisionNodeCompositeNoCollection = getDecisionNode(compositeTypeNoCollection, COMPOSITE_DECISION_DATA_NAME);
+        decisionNodes.add(decisionNodeCompositeNoCollection);
 
-        inputDataNodes.add(inputDataNodeSimpleMock);
-        when(dmnModelMock.getInputs()).thenReturn(inputDataNodes);
+//        dmnModelLocal = new DMNModel() {
```
danielezonca: Bunch of commented lines of code

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/test/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImplTest.java
```diff
@@ -46,125 +56,185 @@ public void init() {
...
+        person.addField(EXPANDABLE_PROPERTY_PHONENUMBERS, phoneNumberCompositeCollection);
+        DMNTypeServiceImpl.ErrorHolder errorHolder = new DMNTypeServiceImpl.ErrorHolder();
+        dmnTypeServiceImpl.checkType(person, errorHolder, "fieldName");
         assertEquals(0, errorHolder.getTopLevelCollection().size());
         assertEquals(1, errorHolder.getMultipleNestedObject().size());
         assertEquals(0, errorHolder.getMultipleNestedCollection().size());
         assertEquals("fieldName.phoneNumbers.complexNumbers", errorHolder.getMultipleNestedObject().get(0));
     }
+
+    private void commonFactModelTreeValidation(FactModelTuple factModelTuple, DMNNode dmnNode, SortedMap<String, FactModelTree> hiddenFacts) {
```
danielezonca: This entire method is a leap of faith for me :D 
Just kidding, it seems fine to me but it is still really hard to follow

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/pom.xml
gitgabrio: @danielezonca 
Because the class used is declared inside the "test" directory, so it is not inherited by the "normal" dependency

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/pom.xml
gitgabrio: @danielezonca 
This is needed because that artifact with the "test" classifier is not declared/managed in parent bom; so, since every dependency must be "managed", I've put it there.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
gitgabrio: @danielezonca 
I'll check - but I think there should be a "common" map - not sure about your case

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
gitgabrio: @danielezonca 
It _was_ checkTypeSupport due to the lack of support for collection, etc. But on one side we are adding this support, on the other side I was thinking to keep it and use as a more generic check/validation.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/test/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImplTest.java
gitgabrio: @danielezonca 
_REFACTORED_

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/DMNDataManagementStrategy.java
gitgabrio: @danielezonca 
That exception is thrown by backend method 
_ FactModelTuple retrieveFactModelTuple(Path path, String dmnPath) throws Exception;_
so it must be catched.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
gitgabrio: @danielezonca 
_FIXED_

gitgabrio:
jenkins please retest this

gitgabrio:
jenkins please retest this

kkufova:
@gitgabrio, I found the following bugs:

1. First, import [this DMN file](http://file.emea.redhat.com/kkufova/dmn-list.dmn). The file does not contain a top-level collection, it is working fine on master. The collection editor popup should display city and country, however, instead it displays only `value`:

bug1

2. [This is a basic example](http://file.emea.redhat.com/kkufova/dmn-list2.dmn) of a DMN file with a top-level collection. The **Unsupported DMN asset** modal window is still opened, even though it works fine and it is possible to create valid test scenarios for the file. The modal is still there for all DMN files with top-level collections.

3. I am not completely sure that [this DMN file](http://file.emea.redhat.com/kkufova/dmn-map.dmn) should work, but currently it causes an unexpected error when trying to invoke the collection popup from `database` when it is moved to the GIVEN section:
>  | Uncaught exception: Index: 0, Size: 0

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/test/java/org/drools/workbench/screens/scenariosimulation/backend/server/AbstractDMNTest.java
```diff
@@ -16,81 +16,219 @@
...
-    protected static final String COMPLEX_DECISION_TYPE_NAME = "COMPLEX_DECISION_TYPE_NAME";
-    protected static final String BASE_TYPE = "BASE_TYPE";
-    protected static final String COMPLEX_TYPE = "COMPLEX_TYPE";
-
-    protected DMNType simpleTypeMock;
-    protected DMNType complexTypeMock;
-    protected DMNType nestedComplexTypeMock;
-    protected Map<String, DMNType> complexFields;
-    protected Map<String, DMNType> nestedComplexFields;
+//    @Mock
```
Rikkola: Minor note. This can be removed.

gitgabrio:
> 1. First, import [this DMN file](http://file.emea.redhat.com/kkufova/dmn-list.dmn). The file does not contain a top-level collection, it is working fine on master. The collection editor popup should display city and country, however, instead it displays only `value`:

_FIXED_



> 2. [This is a basic example](http://file.emea.redhat.com/kkufova/dmn-list2.dmn) of a DMN file with a top-level collection. The **Unsupported DMN asset** modal window is still opened, even though it works fine and it is possible to create valid test scenarios for the file. The modal is still there for all DMN files with top-level collections.

_The collection editor for top level list is currently broken client-side (it works only for simple type collection), that's why is unsupported_



> 3\. I am not completely sure that [this DMN file](http://file.emea.redhat.com/kkufova/dmn-map.dmn) should work, but currently it causes an unexpected error when trying to invoke the collection popup from `database` when it is moved to the GIVEN section:
> 
> > | Uncaught exception: Index: 0, Size: 0

_I'm not sure it should be possible at all to move objects from expected to given (or the other way around) at all; the semantics of DMN is different then RULES: in rules in the RightPanel we see DataType (think of it like java class) and each column created out of them is actually an instance (that's why you can create more then one, wherever you want) while with DMN they are actually "instance" (whatever type) of very specific kind - InputData for the "GIVEN" group and "DecisionData" for the "EXPECTED"); so, moving one object from EXPECTED to GIVEN would mean change its meaning from "decision" to "input", but this should not happen IMO, 'cause it would be a change in the underlying dmn model (that we should not do). @danielezonca WDYT?_

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/DMNDataManagementStrategy.java
danielezonca: @gitgabrio 
Yes but it is not "true". This method is on client side and it call remotely the backend method. I'm quite sure even if you throw that exception, it will be not propagated to the client as exception. I expect it will invoked the error `getErrorCallback(rightPanelPresenter)`.
So at the end it is misleading that `try { .. } catch` because you are not catching it

danielezonca:
@gitgabrio 
It is possible to move a column from `EXPECT` to `GIVEN` because if you want to test a subdecision (so test only partially the model) you should provide all the inputs to that node, even if it is a decision node.
It is useless the opposite (from `GIVEN` to `EXPECT`) but at the moment it is not forbidden

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/DMNDataManagementStrategy.java
gitgabrio: @danielezonca 
_FIXED_

gitgabrio:
@kkufova 
I am not completely sure that this DMN file should work, but currently it causes an unexpected error when trying to invoke the collection popup from database when it is moved to the GIVEN section:

| Uncaught exception: Index: 0, Size: 0

_This is a client-side issue, related to unsupported top-level collection, and will be addressed in specific ticket (https://issues.jboss.org/browse/DROOLS-3759)_

kkufova:
@gitgabrio, I tried the most basic example which seems like it's not working; try [this DMN file](http://file.emea.redhat.com/kkufova/dmn-test.dmn) and [this test scenario](http://file.emea.redhat.com/kkufova/sc-test.scesim). Both defined test scenarios are failing.

gitgabrio:
@kkufova 
I imported both the DMN file and the scesim one without error: what is the issue you are seeing?

kkufova:
@gitgabrio, the tests are failing but they should be passing instead.

gitgabrio:
@kkufova 
I've verified with @danielezonca, and that failure is expected. Top level collection are not yet fully supported (i.e. they do not works, yet). If previously you did not see "failure" launching the run it is only by "chance".

kkufova:
@gitgabrio, @danielezonca, what are you saying? What exactly does this PR introduces then? I don't accept the answer that "top-level collections don't work yet", as they work with some DMNs (and they don't work outside of this PR). I am not sure I'm comfortable with this partial solution, we already reported two issues because of this PR (two I'm aware of) and this looks like a candidate for another one. The DMN I uploaded is the most basic DMN with a list that I can think of. If this does not work, what should work?

gitgabrio:
@kkufova @danielezonca 
I'm going to create an container ticket for the top-level feature and set the existing ones as subtask, to make it clear the scope of every single piece.
This mean that I will close this PR and I will create a new one that will include everything.

kkufova:
@gitgabrio, sounds good to me, thanks!

PR #1093, opened 2019-03-14
[DROOLS-3729] No more possible to introduce aliases in DMN-based test scenario

yesamer:
@kkufova @gitgabrio @danielezonca see https://issues.jboss.org/browse/DROOLS-3729

PR #1094, opened 2019-03-14
DROOLS-3752: Update header cells edit mode invocation

jomarko:
There were two issues:
- `AbstractScenarioSimulationCommand.commonExecution` previously re-enforced grid panel focusing for all commands. This focus was causing flushing `ScenarioHeaderTextAreaDOMElement` and this edit box was immediatelly discarded. Now `AbstractScenarioSimulationCommand.commonExecution` re-enforces grid panel focus, just for commands that change grid structure, for example append row, column, etc.
- `ScenarioSimulationGridHeaderUtilities` contained two methods for edit header precondition (`isEditableHeader` and `isHeaderEditable`) while the second one wasn't used at all but still contained valid logic. Now both are merged and used in `CommonEditHandler.manageHeaderLeftClick`

@kkufova @gitgabrio could you please take a look when you have a moment. Thank you in advance.

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractScenarioSimulationCommand.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractScenarioSimulationCommand.java
@@ -181,7 +181,9 @@ public abstract class AbstractScenarioSimulationCommand extends AbstractCommand<
     protected CommandResult<ScenarioSimulationViolation> commonExecution(ScenarioSimulationContext context) {
         context.getScenarioGridPanel().onResize();
         context.getScenarioGridPanel().select();
-        context.getScenarioGridPanel().setFocus(true);
+        if (undoable) {
+            context.getScenarioGridPanel().setFocus(true);
+        }
         return CommandResultBuilder.SUCCESS;
     }
 }
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
@@ -100,7 +100,7 @@ public class ScenarioSimulationGridHeaderUtilities {
                                            final Integer uiHeaderRowIndex) {
         final GridColumn.HeaderMetaData headerMetaData = column.getHeaderMetaData().get(uiHeaderRowIndex);
         if (!(headerMetaData instanceof ScenarioHeaderMetaData)) {
-            return false;
+            throw new IllegalStateException("Header metadata has to be an instance of ScenarioHeaderMetaData");
         }
 
         final ScenarioHeaderMetaData scenarioHeaderMetaData = (ScenarioHeaderMetaData) headerMetaData;
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
@@ -90,10 +90,29 @@ public class ScenarioSimulationGridHeaderUtilities {
         return ci.getColumn();
     }
 
-    public static boolean isEditableHeader(final GridColumn<?> column,
+    /**
+     * Checks whether the edit mode can be invoked on header cell from given column on given row.
+     * @param column
+     * @param uiHeaderRowIndex
+     * @return true if conditions are met, false otherwise
+     */
+    public static boolean isEditableHeader(final ScenarioGridColumn column,
                                            final Integer uiHeaderRowIndex) {
-        GridColumn.HeaderMetaData headerMetaData = column.getHeaderMetaData().get(uiHeaderRowIndex);
-        return headerMetaData instanceof ScenarioHeaderMetaData && !((ScenarioHeaderMetaData) headerMetaData).isReadOnly();
+        final GridColumn.HeaderMetaData headerMetaData = column.getHeaderMetaData().get(uiHeaderRowIndex);
+        if (!(headerMetaData instanceof ScenarioHeaderMetaData)) {
+            return false;
+        }
+
+        final ScenarioHeaderMetaData scenarioHeaderMetaData = (ScenarioHeaderMetaData) headerMetaData;
+        if (scenarioHeaderMetaData.isEditingMode() || scenarioHeaderMetaData.isReadOnly()) {
+            return false;
+        }
+
+        if (!column.isInstanceAssigned() || !column.isEditableHeaders()) {
+            return false;
+        }
+
+        return scenarioHeaderMetaData.isInstanceHeader() || (scenarioHeaderMetaData.isPropertyHeader() && column.isPropertyAssigned());
     }
 
     public static EnableRightPanelEvent getEnableRightPanelEvent(final ScenarioGrid scenarioGrid,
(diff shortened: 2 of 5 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
```diff
@@ -90,10 +90,29 @@ public static ScenarioHeaderMetaData getColumnScenarioHeaderMetaData(final Scena
...
+     * @param column
+     * @param uiHeaderRowIndex
+     * @return true if conditions are met, false otherwise
+     */
+    public static boolean isEditableHeader(final ScenarioGridColumn column,
                                            final Integer uiHeaderRowIndex) {
-        GridColumn.HeaderMetaData headerMetaData = column.getHeaderMetaData().get(uiHeaderRowIndex);
-        return headerMetaData instanceof ScenarioHeaderMetaData && !((ScenarioHeaderMetaData) headerMetaData).isReadOnly();
+        final GridColumn.HeaderMetaData headerMetaData = column.getHeaderMetaData().get(uiHeaderRowIndex);
+        if (!(headerMetaData instanceof ScenarioHeaderMetaData)) {
```
gitgabrio: @jomarko 
In this case, an exception must be thrown

jomarko: @gitgabrio thank you, I pushed required changes

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractScenarioSimulationCommand.java
```diff
@@ -181,7 +181,9 @@ protected ScenarioGridColumn getScenarioGridColumnLocal(ScenarioSimulationBuilde
     protected CommandResult<ScenarioSimulationViolation> commonExecution(ScenarioSimulationContext context) {
         context.getScenarioGridPanel().onResize();
         context.getScenarioGridPanel().select();
-        context.getScenarioGridPanel().setFocus(true);
+        if (undoable) {
```
gitgabrio: @jomarko 
Why this should be conditional ?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
```diff
@@ -136,11 +155,4 @@ public static String getPropertyName(final Simulation simulation, final int colu
                 .map(ExpressionElement::getStep)
                 .collect(Collectors.toList()));
     }
-
-    public static boolean isHeaderEditable(BaseGridRendererHelper rendererHelper, ScenarioHeaderMetaData clickedScenarioHeaderMetadata, ScenarioGridColumn scenarioGridColumn) {
```
gitgabrio: @jomarko 
I agree that maybe this two methods are redundant and could be merged, but I can not see where the _if/check_ that there was in the deleted one have gone

kkufova:
@yesamer, you were right, the issue with aliases in DMN-based scenarios is fixed within this PR as well.

yesamer:
> @yesamer, you were right, the issue with aliases in DMN-based scenarios is fixed within this PR as well. I think #1093 can be closed.

@kkufova  Yes, I just closed it.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractScenarioSimulationCommand.java
jomarko: @gitgabrio I tried to explain in PR description. The purpose of the `context.getScenarioGridPanel().setFocus(true);` at this place to keep canvas focused, in other words keep canvas reactive on keyboard shortcuts  after column was added, row deleted etc. All these actions are **undoable**, they have some UI effect. The issue is that `AbstractScenarioSimulationCommand` is extended also by commands without UI effect on canvas, for example `EnableRightPanelCmmand`. It means that user invokes context menu for some header cell, however at the same time right panel is focused (by `EnableRightPanelCommand`), the `setFocus(true)` is called, what causes destroying resources bound with canvas, the context menu is gone even before user spot it, or select something. Is it more clear now please?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
jomarko: @gitgabrio The checks are moved into `ScenarioSimulationGridHeaderUtilities.isEditableHeader`. I omitted just the `rendererHelper == null` check.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractScenarioSimulationCommand.java
gitgabrio: @jomarko
Scope of the undoable flag is to know of a given behavior (undo) is enabled on the specific instance. Now, there is not any kind of relationship between undoability and the kind of operation the command is meant ti execute. But you are changing the ui status only based on that flag. To try to explain differenti, a command could be undoable without ever interacting with ui, so why It should change the status of the panel focus. If the currently implemented commands are interacting directly with ui is only by chance, absolutely not by design.

jomarko: @gitgabrio I am sorry, I do not understand your answer, anyway, I will try to change the approach completely as I can not explain more or better this solution.

jomarko:
@gitgabrio Hi, I pushed a new commit where I do not use `undoable` property more. Please have a look. Thank you.

gitgabrio:
@jomarko 
I will try to explain better my question about "undoable" flag. I do understand perfectly that we need to keep the focus on panel to have the keyboard handling working. What I try to explain is that there is not - absolutely - any relationship between the "undoable" flag and the kind of operation the command is doing. An "undoable" command could have nothing to do at all with the ui, and on the other side a "not-undoable" command could change the status of the ui. Nowhere this kind of relationship has been planned nor implemented. It is only "by chance" that currently the undoable commands are working on the ui. Having said that, my point was that you should not consider the "undoable" flag to set the panel focus.
Now, you have two chances: 
1) set always the focus on the panel regardless of the command invoked (the one I was suggesting with "why it is conditional" question)
2) set the focus on the panel on a per-command basis.
I would prefer option 1), just to not have the same line of code repeated in a lot of places (the consequence of your last commit - that is the option 2 - is that whenever we add a new command or change something in the implementation, we have to keep in mind also this status to keep updated all over the code, and this is pretty error-prone).
Last consideration, thanks for the effort you put to fix that, but next time feel free to ask explanation if it is not perfectly clear what I am suggesting.

gitgabrio:
@jomarko 
I think I was not clear enough, my fault. Please revert your last commit, and put that "context.getScenarioGridPanel().setFocus(true);" in one single place, inside AbstractScenarioSimulationCommand.commonExecute but without the if(undoable) check.
The cause of the issue you correctly point out is related to the overall design (like the one about the InsertColumnCommand inheritance), so I want to put in place a clean solution, that require modification to the hierarchy of some commands; I think @yesamer will be able to fix this in that ticket.

jomarko:
@gitgabrio ok, I reverted the changes, to keep focusing grid regardless of command type (undoable / non undoable)

jomarko:
@gitgabrio @yesamer please see the [video](https://drive.google.com/file/d/1-Wwpjni8PacV5n6u6QZiM5BbxxWaQnBP/view?usp=sharing) I fixed that header cells can be edited just if bound to instance/property, however they can be edited just by keyboard, double click is not working, please let me know if I should fix it or you will have a look in a separate ticket.

gitgabrio:
@jomarko 
Many thanks for your help! So, the bug is still there without that if(undoable) check.
This works by chance, anyway we can not keep this PR opened, so restore the if(undoable) and @yeser will fix the overall design later.
Thanks.

jomarko:
@gitgabrio thank you, I understood. I force pushed, here is [video](https://drive.google.com/open?id=1zguHdNXkJZeugpQPGJu01ip4UIyURCp2) to confirm the state.

jomarko:

PR #1099, opened 2019-03-19
[DROOLS-3768] Undo of added property header throws error + test scenarios table disappears

yesamer:
Basically, this issue occurs with the following steps:
- Insert a new column;
- Select the new created column in the grid;
- Undo the operation;
In the code, after the UNDO command, the previous state (without the new column) is then restored. 
But, in `ScenarioGridModel.getInstanceScenarioGridColumns(ScenarioGridColumn selectedColumn) `, it tries to find the selected column in its internal model. Considering that the selected column is the inserted one and this was previously removed after the UNDO command, this leads to the error.
I added a check in the code to avoid this situation. (+ some nullpointer checks)

https://issues.jboss.org/browse/DROOLS-3768

@danielezonca @gitgabrio @kkufova please review. Thanks


kiereleaseuser3:

kkufova:
ok to test

kkufova:
@manstis, I don't have the rights to start the build. Is there a way to automatically start the builds of Yeser's PRs? When I create a PR, I don't have this problem.

manstis:
Jenkins please retest this.

manstis:
@kkufova I think @yesamer will need to be added to the `kie-group` or something in Jenkins.

Let me see if I can do it.. otherwise we might need @mbiarnes help.

kkufova:
@manstis, I think you need to comment `ok to test` first. :woman_shrugging:

manstis:
@kkufova OK I don't appear to have the permissions to add @yesamer.

@mbiarnes Do you have permissions to add @yesamer to the Jenksins group that has builds automatically started!?

manstis:
ok to test.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategy.java
```diff
@@ -98,7 +98,7 @@ protected static FactModelTree getSimpleClassFactModelTree(Class clazz) {
      */
     protected List<String> getPropertiesToHide(ScenarioGridColumn selectedColumn, ScenarioGridModel scenarioGridModel) {
         List<String> toReturn = new ArrayList<>();
-        if (!selectedColumn.isPropertyAssigned()) {
+        if (selectedColumn != null && !selectedColumn.isPropertyAssigned()) {
```
gitgabrio: @yesamer 
I see two issues there
1) that getPropertiesToHide is meant to need the selected column, so it must throw an exception if selectedColumn is null (because it means there is a bug in the calling code) - if you find a way to enforce that please do it
2) if undo/redo needs to call that method, it means that some needed information are missing from status, so you have to add them there; said differently, when calling undo/redo, status must have all the informations it needs to set the status

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategyTest.java
```diff
@@ -70,6 +72,23 @@ public void getPropertiesToHideListPropertyAssigned() {
         commonGetPropertiesToHideList(true);
     }
 
+    @Test
+    public void getPropertiesToHideListWithSelectedColumnNotInGridModel() {
+        doReturn(true).when(gridColumnMock).isPropertyAssigned();
+        List<String> retrieved = abstractDataManagementStrategySpy.getPropertiesToHide(scenarioGridColumnMock, scenarioGridModelMock);
+        assertTrue(retrieved.isEmpty());
+        reset(scenarioGridModelMock);
```
danielezonca: No need to reset a mock at the end of the test

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategyTest.java
```diff
@@ -70,6 +72,23 @@ public void getPropertiesToHideListPropertyAssigned() {
...
+    }
+
+    @Test
+    public void getPropertiesToHideListWithSelectedColumnIsNull() {
+        doReturn(true).when(gridColumnMock).isPropertyAssigned();
+        List<String> retrieved = abstractDataManagementStrategySpy.getPropertiesToHide(null, scenarioGridModelMock);
+        assertTrue(retrieved.isEmpty());
+        verify(scenarioGridModelMock, never()).getSimulation();
+        reset(scenarioGridModelMock);
```
danielezonca: No need to reset a mock at the end of the test

gitgabrio: @yesamer 
About the above statement, remember that soon before execution of a "@Test" annotated method, the "@Before" annotated one is invoked. Reset is only needed when making/verifying multiple "tests" inside the same "@Test" method

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategyTest.java
yesamer: Test removed.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategyTest.java
yesamer: Test removed

yesamer:
@kkufova  I applied the fix in the root cause, can you please retest it? Thanks :)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategy.java
yesamer: @gitgabrio  Reverted and fixed the root cause

yesamer:
jenkins please retest

kkufova:
@danielezonca, @gitgabrio, could you please take a look?

yesamer:
@kkufova Sorry, still not finished to fix this PR, are you sure it works correctly? Because I still have the error.

gitgabrio:
@kkufova 
There is still something to be done.

gitgabrio:
@kkufova 
The code has to be  refactored because there are some issues in there, so it is possible to have unexpected/inconsistent behavior.

kkufova:
@yesamer, @gitgabrio, is this ready for review?

yesamer:
@kkufova @gitgabrio Fine for me. But I suggest to wait for @gitgabrio review before starting test.

kkufova:
@danielezonca, could you please check this again?

danielezonca:
@manstis 

PR #1100, opened 2019-03-19
[DROOLS-3767] Reset "isEditing" status flag inside gridcell

yesamer:
https://issues.jboss.org/browse/DROOLS-3767

To avoid the described (and wrong) behaviour, a try - catch block was added in `ScenarioGridModel.edit()` method. Then, is ANY execption is catched, the "isEditing" status is reset as required.

About tests: in my opinion the current tests are enough: I don't know how to test the case where an exception is thrown during the process, feel free to give me some advice.

@danielezonca @gitgabrio  @kkufova please review.

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/handlers/CommonEditHandler.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/handlers/CommonEditHandler.java
@@ -120,11 +120,7 @@ public class CommonEditHandler {
         if (((ScenarioGridCell) cell).isEditingMode()) {
             return true;
         }
-        ((ScenarioGridCell) cell).setEditingMode((!scenarioGridColumn.isReadOnly()));
-        if (scenarioGridColumn.isReadOnly()) {
-            return false;
-        }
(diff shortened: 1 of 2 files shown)


kie-ci3:

manstis:
Personally, IMO, a general _catch all_ error handler hides the problems and is probably more dangerous (long term) than fixing whatever issues cause DROOLS-3767 to be reported! That's my 2c... and not one to stand in the way of this PR.

gitgabrio:
@manstis 
The original reason for that fix is that  BaseGridWidget.startEditingCell returns a boolean, but in the method flow  BaseCellSelectionManager.doEdit (that currently does not return nothing) is called, and BaseCellSelectionManager.edit always returns true, so there is no way to know if beginning editing was successful or not (for whatever reason).

Currently there is a bug with CollectionEditor when opening top-level list (with dmn-related scenario).
I'm already fixing that bug, but I realized that starting editing a cell *could* potentially throw some exception, maybe randomly due to some underlying transient issue (not so uncommon with client-side applications).
Having said that, I agree that whenever/wherever possible we must tackle and resolve the exception eventually thrown (as I 'm doing for the one in the CollectionEditor that cause this issue).

kkufova:
@gitgabrio, @manstis, @yesamer, are we waiting for any additional changes here?

gitgabrio:
@yesamer 
Fine for me

gitgabrio:
@manstis

manstis:
@gitgabrio Could you please approve before I merge.

gitgabrio:
@manstis
I thought I did :smile:

PR #1106, opened 2019-03-28
[DROOLS-3341] Duplicate DO columns

yesamer:
https://issues.jboss.org/browse/DROOLS-3341 

Dear @kkufova, I kindly ask you to start to test this new functionality 

@gitgabrio @danielezonca @annabaker  please review it.


kie-ci5:

kkufova:
@yesamer, very well done! I only noticed one thing I would like to improve: after you duplicate a column, the new instance does not **automatically** appear in the right panel; it appears there only after you click somewhere on the table header. Can this be fixed?

---
Also, I think that "duplicate column" is misleading. I would prefer renaming the option to "duplicate instance". @gitgabrio, @danielezonca, WDYT? I already proposed this change to Tao in the corresponding Jira.

Another thing; I noticed that removing columns is very uncomfortable and slow; it is currently not possible to remove the whole instance with multiple property columns at once, but you need to delete them one by one. I'm suggesting a different approach; I think property headers should be removed one by one (without a change), but after removing an instance header, it should automatically remove all the corresponding property headers too. @danielezonca, @gitgabrio, WDYT? Should I open a ticket for this?

yesamer:
Dear @kkufova  thanks for the review, I share the compliments with @annabaker which did a great part of the functionality. I can evaluate the fix you proposed, but I'm not sure to complete by the end of the sprint (tomorrow). I'll let you know.

danielezonca:
ok to test

kkufova:
@yesamer, thank you! Don't worry about a deadline, I think it would be okay to finish next week too.

yesamer:
jenkins please retest

danielezonca:
@kkufova 

Yes I agree with your about the possibility to remove the whole instance with a single command. We can create a ticket for that but I suggest to have a confirmation popup too because it could be error prone to remove the whole instance just right clicking a couple of pixels above a property header. Do you agree?

gitgabrio:
> Another thing; I noticed that removing columns is very uncomfortable and slow; it is currently not possible to remove the whole instance with multiple property columns at once, but you need to delete them one by one. I'm suggesting a different approach; I think property headers should be removed one by one (without a change), but after removing an instance header, it should automatically remove all the corresponding property headers too. @danielezonca, @gitgabrio, WDYT? Should I open a ticket for this?

@kkufova 
I thought it was already so. If not, please create a new ticket. @yesamer There is already a "confirmation" pupop used. Please check for consistency, I mean that "Confirmation popup" should be always shown for similar situations. Anyway, IMO let' close this PR with the "duplicate column" rename and the "automatic show in right panel" fixes, and tackle the rest on different ticket.

yesamer:
@kkufova I added the 2 changes you suggested. Can you please test it?
@danielezonca  and @gitgabrio  can you please review it?

yesamer:
jenkins please retest

kkufova:
@yesamer, I discovered that an unexpected error is thrown if you only set an instance, then duplicate the instance, and then try to set the property mapping. If you click somewhere on the canvas, the property suddenly appears. The same issue happens if you try to change a property mapping inside a duplicated instance. However, I think this is only a problem if the property is a collection (a list or a map). Both issues can be seen [here](https://issues.jboss.org/secure/attachment/12448995/screencast-29-03-19-1.webm). Please let me know if you don't understand this description.

I created an additional ticket for one minor issue, but I think this can be solved independently after this PR is merged: [DROOLS-3807](https://issues.jboss.org/browse/DROOLS-3807).

yesamer:
@kkufova @danielezonca @gitgabrio I analyzed the bug reported by Klara, in my opinion it's not related to Duplicate Instance functionality, because it occurs with not duplicated columns too. Then, I created a new ticket for that bug: https://issues.jboss.org/browse/DROOLS-3818
If is fine for you, please review this PR in order to merge it. In the meanwhile, I'll fix the bug in another PR.
Thanks.

kkufova:
@yesamer, me and @danielezonca agreed that it should not be possible to duplicate instances in DMN-based scenarios. Could you please disable the option completely for DMN-based scenarios? I think it should not be available in the context menu at all.

Secondly, I discovered a reproducer that causes an error. Check [this](https://issues.jboss.org/secure/attachment/12449044/screencast-01-04-19-1.webm) screencast.

yesamer:
@kkufova I applied the requested changes. Can you please test it?

kkufova:
@yesamer, I feel like this PR is getting out of hand. :-) I found another issues.

1. It should not be possible to have one single alias for two different properties. However, this is currently possible in duplicated instances:

bug1

2. However, the biggest problem in my opinion is the relationship between GIVEN & EXPECT columns. If you create a duplicate in GIVEN (for example Author 2), it is not possible to use this instance in EXPECT, which is a big drawback:
- you cannot create a duplicate Author 2 in EXPECT (not possible by design);
- and you cannot set type Author 2 to an unassigned column (an error is displayed).

Please let me know if you need more information about this.
@gitgabrio ^

yesamer:
@kkufova I'll check for point 1. Regarding point 2) I don't know how that part works to be honest, so an explanation of the correct behaviour or a documentation could be useful for me. Thanks.

kkufova:
@yesamer, please ignore point 2 for now. I tried again and I cannot reproduce the issue; I'm afraid it may have been caused either by [DROOLS-3768](https://issues.jboss.org/browse/DROOLS-3768), or [DROOLS-3818](https://issues.jboss.org/browse/DROOLS-3818). What I was trying to say is that the new instance created by a duplication in GIVEN section (for example Author 2) should be available in EXPECT section too (it should be possible to set Author 2 as a column type in EXPECT). When I tried it for the first time, I got an error, but I cannot reproduce the issue anymore.

gitgabrio:
@yesamer @kkufova 
I'll clone that branch locally to see what could be the issue.
@yesamer please for the moment being do not modify it

yesamer:
@gitgabrio Work completed! Can ypu please review it?
@kkufova Can you please retest it? As I said, I refactored completely the previous work.
Impacts of this refactoring could be in Assign Instance and Assign property functionality: the user case are: 
- TO assign an INSTANCE only to a selected column;
- TO assign an PROPERTY only to a selected column:
- TO assign an INSTANCE and a PROPERTY to a selected column:
Thank you!

danielezonca:
@kkufova 
Do you need to do another review or not? It is not clear to me just looking at the history of the comments

kkufova:
@danielezonca, yes, I have to check again.

danielezonca:
@manstis 

PR #1108, opened 2019-03-29
Revert DROOLS-3730

danielezonca:
Revert DROOLS-3730 because it causes https://issues.jboss.org/browse/DROOLS-3804

@kkufova


danielezonca:
@manstis 

Can you please merge this?

PR #1109, opened 2019-03-29
Drools 3792 Update right docks

gitgabrio:
Scope of this PR is to refactor overall right panels and to show the settings one.
Please note there has been a renaming inside client module, so most of the file changes are due to that.
Relevant modifications (for client module) are the ones of the righpanel package, ScenarioSimulationEditorPresenter and ScenarioSimulationDocksHandler

@Rikkola 
Please take a look if some of that code could be moved upward to be reused.


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/util/InMemoryMigrationStrategy.java
```diff
@@ -52,4 +52,22 @@
...
+    public Function<String, String> from1_3to1_4() {
+        return rawXml -> {
+            if (rawXml.contains("<type>")) {
+                return rawXml.replaceAll("<ScenarioSimulationModel version=\"1.3\">", "<ScenarioSimulationModel version=\"1.4\">")
+                        .replaceAll("<simulationDescriptor>", "<simulationDescriptor>\n  <fileName></fileName>")
+                        .replaceAll("<type>RULE</type>", "<kieSession>default</kieSession>\n<kieBase>default</kieBase>\n<ruleFlowGroup>default</ruleFlowGroup>\n<skipFromBuild>false</skipFromBuild>\n<type>RULE</type>")
+                        /*
+<dmnNamespace></dmnNamespace>
+<dmnName></dmnName>
+*/
```
Rikkola: Commented code

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/Docks.md
```diff
@@ -0,0 +1,28 @@
+Details on Docks management inside ScenarioSimulationEditor
+===========================================================
```
Rikkola: Using md-file makes sense, just need to make sure the i18n team knows about it.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/CheatSheetView.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
Rikkola: 2019

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/CheatSheetViewImpl.html
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE html>
+<!--
+  ~ Copyright 9 Red Hat, Inc. and/or its affiliates.
```
Rikkola: +2010

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsView.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
Rikkola: 2019

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SubDockView.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
Rikkola: Not sure how important these really are, but still pointing them out just in case.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/TestToolsViewImpl.java
```diff
@@ -0,0 +1,203 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
Rikkola: 2019

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/AbstractSettingsTest.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
```
Rikkola: 2019

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/CheatSheetPresenterTest.java
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
```
Rikkola: This year is 2019

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/Docks.md
Rikkola: Ah. Reading this out now, it is explanation about how the docks work, not show to users :)
But perhaps a good idea for presenting the docks help to users too.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/TestToolsViewImpl.java
gitgabrio: @Rikkola 
Actually this is a renaming - it was RightPanelViewImpl previously

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/CheatSheetPresenterTest.java
gitgabrio: @Rikkola 
?

gitgabrio:
jenkins please retest this

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/CheatSheetPresenterTest.java
Rikkola: It is correct, no need to change ;)

kkufova:
@gitgabrio, I have a couple of notes and questions:

0. First of all, I noticed that the design and the implementation is very different; there are different fields (for example, in actual implementation, the **Package** and **Choose DMN asset** fields are completely missing). Why?

1. (bug) I noticed that no data objects are displayed in the right panel of a rule-based scenario (even though they are available in the project). This seems like a regression.

2. (UX) I noticed that each dock (with the exception of **Test Report**) has a weird top vertical space. Can this be fixed? This applies for **Settings**, **Test Tools**, and **Scenario Cheatsheet**.
bug1

3. (UX) The list of UX issues in the **Settings** tab:
- the right and left indentation is completely missing; please add the same indentation as for example in the **Test Tools** tab;
- why is there the vertical scrollbar?
- the spaces between items are inconsistent; each space has a different space (noticeable for example between the **RuleFlowGroup** and **DMOSession** items);

4. (bug) It is not possible to change the name + type (rule vs. DMN) of the asset in the **Settings** tab. Is this correct?

5. (bug) I don't think we should allow the editing of the **KieSession**, **KieBase**, **RuleFlowGroup**, and **DMOSession**. If you edit one of them and save it, the following unexpected error is thrown:
```
 | Uncaught exception: Parameter named 'path' should be not null! Caused by: Parameter named 'path' should be not null!
```

6. (bug) The **Settings** tabs of a DMN-based and rule-based test scenarios interfere with each other:
- open a rule-based test scenario;
- open a DMN-based test scenario;
- open a rule-based test scenario again: the **Settings** tab is different.

gitgabrio:
> @gitgabrio, I have a couple of notes and questions:
> 
> 1. First of all, I noticed that the design and the implementation is very different; there are different fields (for example, in actual implementation, the **Package** and **Choose DMN asset** fields are completely missing). Why?
_The list of fields (editable or not) to be shown is here https://issues.jboss.org/browse/DROOLS-3572_
> 2. (bug) I noticed that no data objects are displayed in the right panel of a rule-based scenario (even though they are available in the project). This seems like a regression.
_FIXED_
> 3. (UX) I noticed that each dock (with the exception of **Test Report**) has a weird top vertical space. Can this be fixed? This applies for **Settings**, **Test Tools**, and **Scenario Cheatsheet**.
>    bug1
> 4. (UX) The list of UX issues in the **Settings** tab:
_FIXED_
> * the right and left indentation is completely missing; please add the same indentation as for example in the **Test Tools** tab;
_I have to as @srambach  to take a look and provide css_
> * why is there the vertical scrollbar?
_I do not see it_
> * the spaces between items are inconsistent; each space has a different space (noticeable for example between the **RuleFlowGroup** and **DMOSession** items);
_FIXED_
> 
> 1. (bug) It is not possible to change the name + type (rule vs. DMN) of the asset in the **Settings** tab. Is this correct?
_Read answer at 1. Anyway, do not make any sense changing the type of the scenario simulation_
> 2. (bug) I don't think we should allow the editing of the **KieSession**, **KieBase**, **RuleFlowGroup**, and **DMOSession**. If you edit one of them and save it, the following unexpected error is thrown:
> 
> ```
>  | Uncaught exception: Parameter named 'path' should be not null! Caused by: Parameter named 'path' should be not null!
_I do not see this exception: when should this happen ?_
> 1. (bug) The **Settings** tabs of a DMN-based and rule-based test scenarios interfere with each other:
> 
> * open a rule-based test scenario;
> * open a DMN-based test scenario;
> * open a rule-based test scenario again: the **Settings** tab is different.
_FIXED_

kkufova:
A couple of notes:

1. If you change the settings and then close the asset (without saving), the performed changes are saved anyway. Do we want this? As there's no **Cancel** button, I think changes should not be saved automatically when closing an asset, but it would be better to open a warning modal window instead (_You have unsaved changes..._).

2. Anyway, do we need the **Save** button in the dock? I think it's redundant. It does not bring any added value in my opinion.

3. The error after editing the **KieSession**, **KieBase**, **RuleFlowGroup**, or **DMOSession** is still there. You need to:
- change one of these fields;
- save the changes;
- close the asset;
- reopen the asset and
- try to change + save (in the dock) one of the fields again. The following unexpected error is thrown:
```
| Uncaught exception: Parameter named 'path' should be not null! Caused by: Parameter named 'path' should be not null!
```
However, I still believe that editing these fields should be restricted until the functionality is implemented. If you can edit them and it does nothing, it sends the wrong message to the user. **These fields should not be editable at the moment.**
I feel the same way about the skip checkbox. If it does not work, it should not be editable.

4. The **name** field should be fixed. Either it should be possible to rename the asset here, or the name should not be in the text box, but just a value (the same as the type). What is more, it does not reflect the actual state of the asset; if you rename the asset directly in the main top panel, it is not automatically renamed in the **Settings**.

gitgabrio:
> A couple of notes:
> 
> 1. If you change the settings and then close the asset (without saving), the performed changes are saved anyway. Do we want this? As there's no **Cancel** button, I think changes should not be saved automatically when closing an asset, but it would be better to open a warning modal window instead (_You have unsaved changes..._).

_It should not happen, if it happen I have to fix_
> 2. Anyway, do we need the **Save** button in the dock? I think it's redundant. It does not bring any added value in my opinion.

_Because of the above: editing those fields should not *automatically* save new values, unless you click the Save button_
> 3. The error after editing the **KieSession**, **KieBase**, **RuleFlowGroup**, or **DMOSession** is still there. You need to:
> 
> * change one of these fields;
> * save the changes;
> * close the asset;
> * reopen the asset and
> * try to change + save (in the dock) one of the fields again. The following unexpected error is thrown:
> 
> ```
> | Uncaught exception: Parameter named 'path' should be not null! Caused by: Parameter named 'path' should be not null!
> ```

_I'll try that_
> However, I still believe that editing these fields should be restricted until the functionality is implemented. If you can edit them and it does nothing, it sends the wrong message to the user. **These fields should not be editable at the moment.**

_Values from those fields are stored inside the scesim file_
> I feel the same way about the skip checkbox. If it does not work, it should not be editable.
> 

_Same as above_
> 1. The **name** field should be fixed. Either it should be possible to rename the asset here, or the name should not be in the text box, but just a value (the same as the type). What is more, it does not reflect the actual state of the asset; if you rename the asset directly in the main top panel, it is not automatically renamed in the **Settings**.

_The name should not be editable. I agree that it should be kept in sync after renaming from the "Rename" button: I'll take a look at that_

gitgabrio:
@gitgabrio (reminder)
Add "Only the default values are currently supported. Changing them will not have any impact." above "Save" button

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -360,25 +411,50 @@ protected void open(final String downloadURL) {
...
         if (dataManagementStrategy != null) {
-            getRightPanelPresenter().ifPresent(presenter -> {
-                setRightPanel(presenter);
-            });
+            final PlaceRequest currentRightDockPlaceRequest = getCurrentRightDockPlaceRequest(identifier);
+            switch (identifier) {
+                case SettingsPresenter.IDENTIFIER:
+                    getSettingsPresenter(currentRightDockPlaceRequest).ifPresent(presenter -> {
+                        setSettings(presenter);
+                        presenter.setCurrentPath(path);
```
Rikkola: Looking at this with new eyes. 
Now that you have a hold of the dock presenter here, the setup becomes a lot easier if instead of the path you pass in the entire Settings/TestTools/CheatSheet panel. Making these panels fields for the `ScenarioSimulationEditorPresenter`. 
Each dock for this editor then becomes just a placeholder that has a div. This way if you rename the file there is no need for the sync with the Path.

gitgabrio: @Rikkola 
Somehow I've resolved using AtomicLong (new id for each instance of presenter).
Anyway, the solution you proposed seems much better -I'll use it as soon as required code is in place

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -451,26 +563,40 @@ private String getFileDownloadURL(final Supplier<Path> pathSupplier) {
...
 
-    private Optional<RightPanelView> getRightPanelView() {
-        final DefaultPlaceRequest placeRequest = new DefaultPlaceRequest(RightPanelPresenter.IDENTIFIER);
-        if (PlaceStatus.OPEN.equals(placeManager.getStatus(placeRequest))) {
-            final AbstractWorkbenchActivity rightPanelActivity = (AbstractWorkbenchActivity) placeManager.getActivity(placeRequest);
-            return Optional.of((RightPanelView) rightPanelActivity.getWidget());
+    private Optional<TestToolsView> getTestToolsView(PlaceRequest placeRequest) {
+        final Activity activity = placeManager.getActivity(placeRequest);
+        if (activity != null) {
+//        if (PlaceStatus.OPEN.equals(placeManager.getStatus(placeRequest))) {
```
danielezonca: Is `placeManager.getActivity(placeRequest)` equivalent to `PlaceStatus.OPEN.equals(placeManager.getStatus(placeRequest))`?
(btw commented code to remove)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -451,26 +563,40 @@ private String getFileDownloadURL(final Supplier<Path> pathSupplier) {
...
             return Optional.empty();
         }
     }
 
-    private Optional<RightPanelView.Presenter> getRightPanelPresenter() {
-        return getRightPanelView().isPresent() ? Optional.of(getRightPanelView().get().getPresenter()) : Optional.empty();
+    private Optional<CheatSheetView> getCheatSheetView(PlaceRequest placeRequest) {
+        final Activity activity = placeManager.getActivity(placeRequest);
+        if (activity != null) {
+//        if (PlaceStatus.OPEN.equals(placeManager.getStatus(placeRequest))) {
```
danielezonca: Same as above

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -451,26 +563,40 @@ private String getFileDownloadURL(final Supplier<Path> pathSupplier) {
...
+            return Optional.empty();
+        }
     }
 
-    private Command getPopulateRightPanelCommand() {
-        return this::populateRightPanel;
+    private Optional<SettingsView> getSettingsView(PlaceRequest placeRequest) {
+        final Activity activity = placeManager.getActivity(placeRequest);
+        if (activity != null) {
+//        if (PlaceStatus.OPEN.equals(placeManager.getStatus(placeRequest))) {
```
danielezonca: Same as above

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/TestToolsViewImpl.java
danielezonca: @gitgabrio 
`git` is considering this as new file and `CheatSheetViewImpl` as a renaming of `RightPanelViewImpl`. Btw fine for me too keep it as is

srambach:
Regarding the spacing around the dock panels - 
Test Report has a couple more layers of divs around it with Bootstrap grid classes on them. Those are making the space around that one. However, the grid there is not really correctly done and could be removed so that it's consistent with the others. Then at the level where there is a .tab-pane, I'd suggest adding .kie-tab-pane--dmn-panel or something else unique to those, and put a 1em padding around. Currently this is being done with inline styles, so those inline styles should be removed.

As for the scroll bar, I only see it appropriately when the content is too tall.

srambach:
I do also see an issue in the Test Tools where there are phantom lines that are, I think, <hr>s between empty lists. Can those be removed until there's a list to separate?
image

gitgabrio:
@srambach 
I've removed the empty hr.
Beside that, could you please give me the modified html/css working inside http://10.0.77.96:8080/business-central/kie-wb.jsp ? Thx.

srambach:
You can add  `.kie-tab-pane--dmn-panel { padding: 1em; }` to wherever you are getting your css. I can put this into the CSS file I've been providing, but I'm not seeing it in the sources that you are including anyway (I could be missing it). 
Most of the work to be done is that there is inconsistency in the HTML of the 4 tab panels.
If you could get them consistent to this css class structure, it would help greatly. 
`<div class="uf-perspective-component">
  <div class="uf-scroll-panel">
    <div class="tab-pane kie-tab-pane--dmn-panel ">
      contents here
    </div>
  </div>
</div>`
If you can do that, I think there are some inline styles adding padding that can be removed. For example, `#kieScenarioCheatsheetTabContent` and `#kieTestEditorTabContent` both have an inline padding of 1em that should be removed.

For the test report tab, there are extra layers of Bootstrap grid that should be removed (see bold below)
`<div class="uf-perspective-component" >
  <div class="uf-scroll-panel" >
    **<div class="container-fluid">
        <div class="row">
            
            <div class="col-xs-9 tab-content kie-tab-pane--dmn-panel  kie-tab-content--filters" style="width: 100%;">**
                
                <div class="tab-pane kie-tab-pane kie__dock active" id="kieDockPanel1">
                  Content
                </div>
            **</div>
        </div>
    </div>**
  </div>
</div>`

I hope that helps; if not I'm happy to talk. I can try marking lines in the HTML files in this PR, but I'm not sure that covers everything.

gitgabrio:
@srambach 
Thanks! I'll try (to let you know, I'll use kie-tab-pane--scescim-panel to avoid any possible confusion).
I'll call you if I can't get it work :smile:

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsViewImpl.html
```diff
@@ -0,0 +1,96 @@
...
+            </div>
+        </div>
+        <br/>
+        <div class="checkbox">
+            <label for="kieScenarioSettingsSkipFromBuild">
+                <input type="checkbox" data-field="skipFromBuild" id="kieScenarioSettingsSkipFromBuild">Skip this simulation during the build
+            </label>
+        </div>
+        <br/>
+        <p data-field="toBeRemoved" style="font-style: italic">Only the default values are currently supported. Changing them will not have any impact on running scenario.</p>
```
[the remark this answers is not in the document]
gitgabrio: @kkufova 
I put that as it is because saving those values will actually change the file scesim - so the asset. It does not effect hot the scesim is "*run*, but it is a different aspect.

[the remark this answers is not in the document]
kkufova: @gitgabrio, OK, in that case, please change it to:

> Only the default values are currently supported. Changing them will not have any impact on test execution.

PR #1112, opened 2019-04-02
DROOLS-3438: Cancel cell edit mode by TAB

jomarko:
Ensemble together with:
- kiegroup/appformer#678

This PR prevents the browser from handling TAB key. Without this PR, next html page element was focused.

@kkufova may I ask you for review?
@gitgabrio @yesamer may I ask one of you for review too?


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/factories/AbstractTextBoxSingletonDOMElementFactory.java
```diff
@@ -37,10 +39,19 @@ public AbstractTextBoxSingletonDOMElementFactory(final GridLienzoPanel gridPanel
...
     @Override
     public TextArea createWidget() {
-        return new TextArea() {{
-            addKeyDownHandler(DomEvent::stopPropagation);
-            addMouseDownHandler(DomEvent::stopPropagation);
-        }};
+        final TextArea textArea = GWT.create(TextArea.class);
+        textArea.addKeyDownHandler(keyDownEvent -> {
+            final int keyCode = keyDownEvent.getNativeKeyCode();
+            if (keyCode == KeyCodes.KEY_TAB) {
```
gitgabrio: @jomarko 
A couple of question:
1) from the ticket description it seems that somehow both TAB and ESC should be managed (in different way) but I do not see that here: are the ESC managed elsewhere or in different way?
2) is this kind of interaction specific ot ScenarioSImulation? Is it not the same for other grid editor? If the behavior should be the same on different component, then it should be better to create a builder/getter somewhere else (some abstract/parent class) to avoid duplicate this snippet

jomarko: @gitgabrio thank you for review, you had good point, the same logic was for guided decision tables.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/domelements/ScenarioCellTextAreaDOMElement.java
```diff
@@ -115,7 +115,6 @@ public void setTabIndex(final int index) {
     public void flush(final String value) {
         String actualValue = value != null && value.isEmpty() ? null : value;
         if (scenarioGridCell != null) {
-            scenarioGridCell.setEditingMode(false);
```
gitgabrio: @jomarko 
Are you sure about removing that line? I remember times ago I put it there for some reason, not sure if it is still needed; anyway, the point at that time was that if ScenarioGridCell.editingMode is "true", the editing does not start. When you press "enter" or anyway end editing, the "flush" method gets invoked, and after that the cell must be editable again.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/domelements/ScenarioHeaderTextAreaDOMElement.java
```diff
@@ -45,7 +45,6 @@ public void setScenarioHeaderMetaData(ScenarioHeaderMetaData scenarioHeaderMetaD
     @Override
     public void flush(final String value) {
         if (scenarioHeaderMetaData != null) {
-            scenarioHeaderMetaData.setEditingMode(false);
```
gitgabrio: @jomarko 
Same comment as above

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/domelements/ScenarioCellTextAreaDOMElement.java
jomarko: Yes I am sure, see https://github.com/kiegroup/appformer/pull/678/files, detach is always called when flush.

jomarko:
@gitgabrio yes, it should be rebased with master.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/domelements/ScenarioCellTextAreaDOMElement.java
gitgabrio: OK

gitgabrio: @jomarko 
I've read your modification to ScenarioCellTextAreaDOMElement and BaseSingletonDOMElementFactory, what I see is
1) it is "true" that BaseSingletonDOMElementFactory.flush calls detach that in turns set editing mode false
2) it is not true that ScenarioCellTextAreaDOMElement.flush calls detach; so whatever code calls directly ScenarioCellTextAreaDOMElement.flush (being public is perfectly valid) does not reset the editing mode
3) moving setEditingMode(false) to the detach method add an unneeded "indirection" and contrived logic, binding together an event (detach) and a behavior (set editingmode) that are not strictly related
4) the editingmode is first of all a *concept*, that it is related to the fact that the user has ended editing; and "ending editing" is represented by the "flush" method, not by "detach"
5) it is just by "chance" that detach and flush are somehow related in our current situation.

jomarko:
@gitgabrio changes as discussed on gchat are now in, I also tried manually and seems working to me, could you please check?

jomarko:
Rebased with master

jomarko:
@manstis I think we can merge

PR #1113, opened 2019-04-03
[DROOLS-3097] Implement export/import mechanism for Test Scenario

danielezonca:
https://issues.jboss.org/browse/DROOLS-3097

It adds export capability to Test Scenario

NOTE: this PR contains also backend import service but not the UI for that (this will be implemented with https://issues.jboss.org/browse/DROOLS-3839)

@kkufova @gitgabrio @Rikkola   

PR list:
- https://github.com/kiegroup/drools-wb/pull/1113
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/945


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/FactMapping.java
```diff
@@ -67,7 +67,7 @@ public FactMapping() {
     }
 
     public FactMapping(FactIdentifier factIdentifier, ExpressionIdentifier expressionIdentifier) {
-        this(expressionIdentifier.getName(), factIdentifier, expressionIdentifier);
+        this(factIdentifier.getName(), factIdentifier, expressionIdentifier);
```
gitgabrio: This change could potentially have a big impact: did you test at runtime everything client-side is still working?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/SimulationDescriptor.java
```diff
@@ -143,7 +143,7 @@ public FactMapping addFactMapping(String factAlias, FactIdentifier factIdentifie
     }
 
     public FactMapping addFactMapping(int index, FactIdentifier factIdentifier, ExpressionIdentifier expressionIdentifier) {
-        return addFactMapping(index, expressionIdentifier.getName(), factIdentifier, expressionIdentifier);
+        return addFactMapping(index, factIdentifier.getName(), factIdentifier, expressionIdentifier);
```
gitgabrio: Same comment as above

kkufova:
@danielezonca, I found the following issues:

1. It is only possible to export a file once; clicking **Export** stops working after one use. You need to log out and log in again to Business Central to be able to export a file for the second time (and then it again stops working).

2. Also, I would prefer the **Export** button to be a drop-down. I know that we only support CSV format now, but this may change and it's not clear at the first sight what exactly is going to happen after the "export". The drop-down would make it clear that the user can export the `.scesim` file as a CSV. Something similar to this (from the DMN designer):



3. Regarding the actual exported CSV file, I noticed a few things:
- first of all, I'm not sure the strings (from a DMN-based scenario) are correctly exported. IMHO, it should only be _e.g._ `"SPEED"`, as opposed to:
```
1,,10,120,100,"""SPEED""",500,3,false
```
- second of all, why is `"#"` in quotation marks, but `Scenario description` isn't? Is it because a `#` can be a delimiter?
- next, I noticed how the headers are represented:
```
OTHER,OTHER,GIVEN,...
"#",Scenario description,Driver...
,,points...
```
Wouldn't it make more sense to use the following instead? To me, it may be considered redundant, but it's clear what it means and it's easier to read:
```
#,Scenario description,GIVEN,...
#,Scenario description,Driver...
#,Scenario description,points...
```
- finally, I think it is inconsistent how the expressions from a rule-based scenario are exported. If I understand correctly, if an expressions contains a comma (or `;`), quotation marks are automatically added. Correct? Wouldn't it make sense to add the quotation marks to all expressions?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/FactMapping.java
danielezonca: This constructor has been used only for testing purpose that's why nobody found the bug before (and there is no impact on UI too). We can refactor the test and remove this code but for now at least is correct :)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/SimulationDescriptor.java
danielezonca: Same answer :)

danielezonca:
@kkufova 
I will have a look to 1 and 2 but I don't think point 3 is a bug: I'm using Apache `commons-csv` library to generate the file and the format is related to the CSV specs ( https://tools.ietf.org/html/rfc4180 ). Actually I'm using the `EXCEL` format ( https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#EXCEL ) that use the same specs. If you want I can change to `DEFAULT`.
The summary of this is: I don't know why some values/symbols are escape with `"` but it is CSV spec :)

danielezonca:
@kkufova 
Merged Import feature too (https://github.com/kiegroup/drools-wb/pull/1117) and fixed bugs.

Created https://issues.jboss.org/browse/DROOLS-3887 to replace current Export/Import button

kkufova:
@danielezonca, the export does not really work; only the headers are exported:

```
OTHER,OTHER,GIVEN,GIVEN,GIVEN,GIVEN,EXPECT,EXPECT,EXPECT
"
```

```
OTHER,OTHER,GIVEN,EXPECT
"
```

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -392,6 +409,34 @@ protected void open(final String downloadURL) {
...
+    protected void onExportToCsv() {
+        importExportService.call(getExportCallBack(),
+                                 new DefaultErrorCallback())
+                .exportSimulation(CSV, context.getStatus().getSimulation());
+    }
+
+    protected RemoteCallback<Object> getExportCallBack() {
+        return rawResult -> {
+            final AnchorElement exportAnchorElement = view.getExportAnchorElement((String) rawResult, path.getFileName() + CSV.getExtension());
+            final NativeEvent CLICK_EVENT = Document.get().createClickEvent(0, 0, 0, 0, 0, false, false, false, false);
```
Rikkola: Just wondering. What is this used for?

kkufova:
Notes for myself:

- check both Chrome and Firefox;
- check in LibreOffice Calc.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
danielezonca: No more used :D
Btw I used this _trick_ to force client side file creation and download. Now I found a better way used by Stunner too and I updated the code

Rikkola: Ok. Thought this was something like that.

danielezonca:
@manstis 

Can you please squash and merge this and https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/945 ?

PR #1115, opened 2019-04-03
DROOLS-3103 : Support run test at project level

Rikkola:
https://issues.jboss.org/browse/DROOLS-3103

@kkufova @danielezonca @gitgabrio If you could please take a look.

Bundle:
https://github.com/kiegroup/appformer/pull/667
https://github.com/kiegroup/kie-wb-common/pull/2568
https://github.com/kiegroup/drools-wb/pull/1115

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/TestRunResult.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/TestRunResult.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,39 +14,29 @@
  * limitations under the License.
  */
 
-package org.drools.workbench.screens.testscenario.model;
+package org.drools.workbench.screens.scenariosimulation.model;
 
-import java.util.Set;
+import java.util.Map;
 
-import org.drools.workbench.models.testscenarios.shared.Scenario;
 import org.guvnor.common.services.shared.test.TestResultMessage;
 import org.jboss.errai.common.client.api.annotations.MapsTo;
 import org.jboss.errai.common.client.api.annotations.Portable;
 
 @Portable
-public class TestScenarioResult {
+public class TestRunResult {
 
-    private Scenario scenario;
-    private Set<String> log;
-    private TestResultMessage testResultMessage;
+    private final Map<Integer, Scenario> map;
 
-    public TestScenarioResult() {
-    }
+    private final TestResultMessage testResultMessage;
 
-    public TestScenarioResult(@MapsTo("scenario")Scenario scenario,
-                              @MapsTo("log")   Set<String> log,
-                              @MapsTo("testResultMessage")  TestResultMessage testResultMessage) {
-        this.scenario = scenario;
-        this.log = log;
+    public TestRunResult(@MapsTo("map") Map<Integer, Scenario> map,
+                         @MapsTo("testResultMessage") TestResultMessage testResultMessage) {
+        this.map = map;
         this.testResultMessage = testResultMessage;
     }
 
-    public Scenario getScenario() {
-        return scenario;
-    }
-
-    public Set<String> getLog() {
-        return log;
+    public Map<Integer, Scenario> getMap() {
+        return map;
     }
 
     public TestResultMessage getTestResultMessage() {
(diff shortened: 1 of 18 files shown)


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/ScenarioEditorPresenter.java
```diff
@@ -231,19 +228,6 @@ private void renderFixtures() {
...
-        testService.call(new RemoteCallback<Void>() {
-                             @Override
-                             public void callback(Void v) {
-                                 view.hideBusyIndicator();
-                             }
-                         },
-                         getTestRunFailedCallback()).runAllTests(user.getIdentifier(),
-                                                                 versionRecordManager.getCurrentPath());
-    }
-
```
Rikkola: I'm removing the "run all" from the legacy editor.

Rikkola:
The test failures look like I broke something.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioRunnerServiceImpl.java
```diff
@@ -46,50 +48,67 @@
...
-                new TestResultMessage(
-                        identifier,
-                        1,
-                        1,
-                        new ArrayList<>()));
+        for (Map.Entry<Path, ScenarioSimulationModel> entry : scenarioLoader.loadScenarios(path).entrySet()) {
+
+            final ScenarioSimulationModel scenarioSimulationModel = entry.getValue();
+
+            runTest(identifier,
```
gitgabrio: @Rikkola 
It seems that customTestResultEvent.fire(...); gets invoked once for each entry (inside the runTest method) and then another time at the end of the loop: is this actually wanted? How/where are managed all that different "events" ?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/handlers/ScenarioSimulationDocksHandler.java
```diff
@@ -36,13 +37,19 @@
...
+    @Inject
+    private PlaceManager placeManager;
 
     private UberfireDock toolsDock;
     private UberfireDock reportDock;
 
     @Override
     public Collection<UberfireDock> provideDocks(final String perspectiveIdentifier) {
+
+        placeManager.closePlace("org.kie.guvnor.TestResults");
```
gitgabrio: @Rikkola 
Why is this needed ? I have the feeling that injecting and using PlaceManager here somehow hides something that should be managed elsewhere - just an impression. Beside that, even if technically it is possible to inject stuff (and  all that works only because they are sort of "singleton") this creates a sort a complex coupling that we should strive to avoid, IMO.

Rikkola: This is since there might be a TestResult screen open. For example when running all the test the results are opened into a panel. Then when looking at the editor it tries to open the same screen in two locations. One in regular panel and another in docks.

To get rid of this I guess I would have to use similar approach to what you had where the editor and the dock instance are tied together.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioRunnerServiceImpl.java
Rikkola: Inside runTest method it fires the event to the CollectorEvent that does not publish the event outside of this class. Then here a new event is created based on the event and published.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/handlers/ScenarioSimulationDocksHandler.java
Rikkola: Actually does not make sense to call this here. Not sure why it was here. Removing and testing.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/src/main/java/org/drools/workbench/screens/scenariosimulation/model/TestRunResult.java
```diff
@@ -0,0 +1,45 @@
...
+package org.drools.workbench.screens.scenariosimulation.model;
+
+import java.util.Map;
+
+import org.guvnor.common.services.shared.test.TestResultMessage;
+import org.jboss.errai.common.client.api.annotations.MapsTo;
+import org.jboss.errai.common.client.api.annotations.Portable;
+
+@Portable
+public class TestRunResult {
```
danielezonca: I created a similar class during the implementation of the DMN test coverage report. So I will probably need to update the code as soon as this PR has been merged. Considering that has been created since a while I think it is better to finalize this first

PR #1118, opened 2019-04-11
RHDM-911: Prevent text from overlapping

jomarko:


jomarko:

jomarko:
@kkufova @Rikkola HI, could you please have a look? There is already finished full downs. build so should be easy to verify.

PR #1119, opened 2019-04-12
Drools 3873

gitgabrio:

See https://issues.jboss.org/browse/DROOLS-3873


kkufova:
@gitgabrio, it is not possible to rename any instance or property.

gitgabrio:
@manstis

PR #1123, opened 2019-04-19
Drools 3922 Import of external classes in Test Scenario doesn't work until you have a class in your package

gitgabrio:

see https://issues.jboss.org/browse/DROOLS-3922

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategy.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/strategies/AbstractDataManagementStrategy.java
@@ -57,7 +57,9 @@ public abstract class AbstractDataManagementStrategy implements DataManagementSt
         String fullName = clazz.getCanonicalName();
         simpleProperties.put("value", fullName);
         String packageName = fullName.substring(0, fullName.lastIndexOf("."));
-        return new FactModelTree(key, packageName, simpleProperties, new HashMap<>());
+        FactModelTree toReturn = new FactModelTree(key, packageName, simpleProperties, new HashMap<>());
+        toReturn.setSimple(true);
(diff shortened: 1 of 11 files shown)


kkufova:
@gitgabrio, I haven't discovered any serious bugs. However, could you please explain why it looks like there are two different lists of objects? In my opinion, there should only be one:

bug1

Also, I reported this issue: [DROOLS-3931](https://issues.jboss.org/browse/DROOLS-3931).

gitgabrio:
> @gitgabrio, I haven't discovered any serious bugs. However, could you please explain why it looks like there are two different lists of objects? In my opinion, there should only be one:
> 
> bug1
> 
> Also, I reported this issue: [DROOLS-3931](https://issues.jboss.org/browse/DROOLS-3931).

_This should reflect the types of the items and how they are parsed by the DMN engine. Basically, only *Borrower* is a complex type (so, it is in the upper list) while the others are *simple* types (number/string - so they are below). This is the same as in *Rule* scesim, where you have *bean* types in the upper list (Book, Author, etc.) and simple types below (Boolean, Number, String). Opening the DMN file in the DMN editor seems to confirm that. Please correct me if I'm wrong_

kkufova:
@gitgabrio, in that case, I have a proposal: could we maybe include a headers called for example **Simple Types** and **Complex Types** to make it clear what the division means?

gitgabrio:
@kkufova 
If every one agreed upon, no problem for me. Please keep in mind that beside *Simple Types* and *Complex Types* we also have the *aliases*: just let me know exact wording and css style. @danielezonca WDYT ?

kkufova:
@gitgabrio, you're right! In that case, I would use **Simple Types**, **Complex Types**, and **Custom Instances**. Do you agree? @danielezonca, WDYT?

When it comes to CSS style, I would use the same style as there is for **Select Data Object**.

gitgabrio:
@kkufova 
It is fine for me to do in the current PR - the headers are already in place, I just have to put words in them

gitgabrio:
@kkufova 
Add labels to group separator. Labels should be visible only if the group is not empty. This should work for bot Complex/Simple types and Complex/Simple instances

gitgabrio:
> Please add some unit tests to cover the fixes

_DONE_

gitgabrio:
@manstis

PR #1124, opened 2019-04-19
Drools 3862 If an instance alias contains ".", it is not possible to set a property mapping

gitgabrio:

see https://issues.jboss.org/browse/DROOLS-3862

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/events/EnableTestToolsEvent.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/events/EnableTestToolsEvent.java
@@ -15,6 +15,8 @@
  */
 package org.drools.workbench.screens.scenariosimulation.client.events;
 
+import java.util.List;
+
 import com.google.gwt.event.shared.GwtEvent;
 import org.drools.workbench.screens.scenariosimulation.client.handlers.EnableTestToolsEventHandler;
 
@@ -37,9 +39,9 @@ public class EnableTestToolsEvent extends GwtEvent<EnableTestToolsEventHandler> 
     private final boolean notEqualsSearch;
 
     /**
-     * The string to <b>eventually</b> use to select the property in the test tools  panel
+     * The <code>List</code> to <b>eventually</b> use to select the property in the test tools  panel
      */
-    private final String propertyName;
+    private final List<String> propertyNameElements;
 
     /**
      * Fire this event to show all the first-level data models <b>enabled</b> (i.e. <b>double-clickable</b> to map to an <i>instance</i> header/column)
@@ -48,7 +50,7 @@ public class EnableTestToolsEvent extends GwtEvent<EnableTestToolsEventHandler> 
     public EnableTestToolsEvent() {
         filterTerm = null;
         notEqualsSearch = false;
-        propertyName = null;
+        propertyNameElements = null;
     }
 
     /**
@@ -60,7 +62,7 @@ public class EnableTestToolsEvent extends GwtEvent<EnableTestToolsEventHandler> 
     public EnableTestToolsEvent(String filterTerm) {
         this.filterTerm = filterTerm;
         notEqualsSearch = false;
-        propertyName = null;
+        propertyNameElements = null;
     }
 
     /**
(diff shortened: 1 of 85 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/events/EnableTestToolsEvent.java
```diff
@@ -37,9 +39,9 @@
     private final boolean notEqualsSearch;
 
     /**
-     * The string to <b>eventually</b> use to select the property in the test tools  panel
+     * The <code>List</code> to <b>eventually</b> use to select the property in the test tools  panel
      */
-    private final String propertyName;
+    private final List<String> propertyNameElements;
```
Rikkola: Just an idea, but would it make sense to wrap `List<String>` to a `PropertyNameElements` class, so that it is clear what the list is about and to perhaps move some actions done against the list into the `PropertyNameElements` class?

Rikkola: Same goes for `List<String> headerCellElements`, not sure if the two can share a class.

kkufova:
@gitgabrio, two things:

1. I noticed that it is **not** possible to use a period at the end of an alias now (it automatically disappears, _e.g._ `Klara K.` changes to `Klara K`). Is this on purpose?

2. Also, once you introduce an alias with a period (such as `Auth.or` or `Boo.k`), you **cannot** change (or set) the types of property columns (nothing happens).

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/events/EnableTestToolsEvent.java
gitgabrio: @Rikkola 
Actually, in general I like the idea of having a more semantically-defined class. But for this specific situation my thoughts were
1) there is no fixed number of elements - so I should have anyway a list behind the scene
2) creating a specific class would require anyway some "effort" (code to write/maintain, small as it may be), that for this specific situation probably won't be paid off

Anyway, if for some reason there would be the needs/option to have something more specific instead of a List, I'll do

gitgabrio:
> @gitgabrio, two things:
> 
> 1. I noticed that it is **not** possible to use a period at the end of an alias now (it automatically disappears, _e.g._ `Klara K.` changes to `Klara K`). Is this on purpose?

_NOPE_

> 2. Also, once you introduce an alias with a period (such as `Auth.or` or `Boo.k`), you **cannot** change (or set) the types of property columns (nothing happens).

_LOOKING AT IT...._

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/events/EnableTestToolsEvent.java
Rikkola: Sure. I think here your the author so you decide since you are the one that has the better understanding on it.

gitgabrio:
@kkufova 
As agreed, I completely blocked use of _dot_ inside aliases (both _instance_ and _property_).
Now a more specific message is shown when alias is invalid.

kkufova:
@gitgabrio,

1. It's possible to use `.` at the end of a property alias (such as `name.`). This should be restricted too.
2. Please change the error messages to:
> An instance alias cannot contain periods!

> A property alias cannot contain periods!

gitgabrio:
@kkufova 

> 1. It's possible to use `.` at the end of a property alias (such as `name.`). This should be restricted too.
> 2. Please change the error messages to:
> 
> > An instance alias cannot contain periods!
> 
> > A property alias cannot contain periods!

_FIXED_

danielezonca:
Please note that we don't have real control of what user specifies in a DMN model so we cannot really avoid that someone create a DMN node with `.` and then use it into the scenario.
Have you tested this scenario?

kkufova:
@danielezonca, it should be okay, as you cannot create any aliases in DMN-based test scenarios. But I will check how type names with periods behave.

gitgabrio:
@kkufova @danielezonca 
Keep in mind that the check on the "dot" is done for manually inserted aliases not matching to any existing properties. For ex, for nested properties (author.book.name) it is valid if they are add from the right panel, and if by chance an object (inside dmn) contains/ends with a dot, it should be allowed too

kkufova:
@manstis, could you please merge?

PR #1127, opened 2019-04-24
[Drools-3890] Duplicated instance does not automatically disappear from right panel after removing it

yesamer:
@kkufova @gitgabrio @danielezonca Please test it and review.

Consider that is not related to Duplicated Column. It reproduces in this case too:
- Insert a column 
- Rename the instance (eg Author to Author_test)
- Delete it


yesamer:
@manstis Hi Michael! Can you please merge it? Thanks.

PR #1134, opened 2019-05-07
DROOLS-3432 : While using the same template key with difference objects/attributes causes an exception when switching to the Data tab in a Guided Rule Template

Rikkola:
https://issues.jboss.org/browse/DROOLS-3432
If the same variable is used for two fields the UI defaults to String cell editor.

PRs:
https://github.com/kiegroup/drools/pull/2332
https://github.com/kiegroup/drools-wb/pull/1134

### DIFF
--- a/drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-client/src/main/java/org/drools/workbench/screens/guided/template/client/editor/TemplateDataCellFactory.java
+++ b/drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-client/src/main/java/org/drools/workbench/screens/guided/template/client/editor/TemplateDataCellFactory.java
@@ -20,6 +20,7 @@ import java.math.BigInteger;
 import java.util.Date;
 
 import com.google.gwt.event.shared.EventBus;
+import org.drools.workbench.models.guided.template.shared.TemplateModel;
 import org.kie.soup.project.datamodel.oracle.DataType;
 import org.kie.workbench.common.widgets.client.datamodel.AsyncPackageDataModelOracle;
 import org.kie.workbench.common.widgets.decoratedgrid.client.widget.AbstractCellFactory;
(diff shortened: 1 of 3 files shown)


drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-client/src/main/java/org/drools/workbench/screens/guided/template/client/editor/TemplateDataCellFactory.java
```diff
@@ -63,6 +64,10 @@ public TemplateDataCellFactory(AsyncPackageDataModelOracle oracle,
      */
     public DecoratedGridCellValueAdaptor<? extends Comparable<?>> getCell(TemplateDataColumn column) {
 
+        if(column.getDataType().equals(TemplateModel.DEFAULT_TYPE)){
```
manstis: Unfortunately you need to do something similar for the Guided Decision Table's use of "BRL Fragments".

I had a quick look around for _where_ a similar change is needed, but unfortunately my brain let me down.

I'd put a break point in `BRLConditionVariableColumnConverter` and see where it leads.

Rikkola: No worries, good thing you spotted this. I'll look for it.

Rikkola: BRLConditionColumnPlugin.getDefinedVariables(RuleModel ruleModel) is where I need to add some changes. Adding this here as a reminder for myself ;)

jomarko:
No comment to code, manual check pending.

Rikkola:
@jomarko @manstis These PRs could use a rereview thank you.

Rikkola:
@manstis this needs to go in too :)

PR #1135, opened 2019-05-09
DROOLS-3918 : The error reporting should contain information about th…

Rikkola:
https://issues.jboss.org/browse/DROOLS-3918

https://github.com/kiegroup/appformer/pull/695
https://github.com/kiegroup/kie-wb-common/pull/2630
https://github.com/kiegroup/drools-wb/pull/1135

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioRunnerServiceImpl.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioRunnerServiceImpl.java
@@ -80,7 +80,7 @@ public class ScenarioRunnerServiceImpl extends AbstractKieContainerService
 
         final List<Failure> failureDetails = new ArrayList<>();
 
-        final Result result = runWithJunit(scenarioRunner, failures, failureDetails);
+        final Result result = runWithJunit(path, scenarioRunner, failures, failureDetails);
 
         return new SimulationRunResult(scenarios,
                                        scenarioRunner.getLastRunResultMetadata()
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/main/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerService.java
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/main/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerService.java
@@ -17,7 +17,7 @@
 package org.drools.workbench.screens.testscenario.backend.server;
 
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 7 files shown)


Rikkola:
@jomarko @gitgabrio @danielezonca Here is one for reviews.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioRunnerServiceImpl.java
```diff
@@ -80,7 +80,7 @@ public SimulationRunResult runTest(final String identifier,
 
         final List<Failure> failureDetails = new ArrayList<>();
 
-        final Result result = runWithJunit(scenarioRunner, failures, failureDetails);
+        final Result result = runWithJunit(path, scenarioRunner, failures, failureDetails);
```
gitgabrio: @Rikkola 
Related to my other comments on the PR, I have the impression (correct me if I am wrong) that *Path* is used only to forward its internal String representation. Just this simple usage creates an unbreakable bound between at least three repos (directly) and who can tell how many other (indirectly).

Rikkola: It needs to be the actual Path, since that is then used in the UI. The user clicks the Path and the asset opens.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/main/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerService.java
```diff
@@ -110,22 +111,30 @@ public TestScenarioResult run(final String identifier,
...
+
+            for (Map.Entry<Path, Scenario> entry : scenarios.entrySet()) {
+                ScenarioRunner4JUnit scenarioRunner = new ScenarioRunner4JUnit(
+                        entry.getValue(),
+                        kSessions,
+                        getMaxRuleFirings());
+                result.add(run(identifier,
+                               entry.getKey(),
+                               scenarioRunner));
+            }
```
jomarko: Minor comment, just for consistency, could we use `final` for all vars?
```java
final ArrayList<TestResultMessage> result = new ArrayList<>();
            final Map<Path, Scenario> scenarios = scenarioLoader.loadScenarios(path);

            final Map<String, KieSession> kSessions = getKSessions(path,
                                                                   scenarios.values());

            for (Map.Entry<Path, Scenario> entry : scenarios.entrySet()) {
                final ScenarioRunner4JUnit scenarioRunner = new ScenarioRunner4JUnit(
                        entry.getValue(),
                        kSessions,
                        getMaxRuleFirings());
                result.add(run(identifier,
                               entry.getKey(),
                               scenarioRunner));
            }
```

Rikkola: We should, updating.

Rikkola: Updated.

Rikkola:
@kiegroup/gatekeepers This bundle can be merged thank you.

PR #1136, opened 2019-05-15
[DROOLS-3842] Error after undo of invalid instance mapping

yesamer:
Now it's no more possible to add an instance to a column with a instance already assigned (in this case, ADD button is disabled)

https://issues.jboss.org/browse/DROOLS-3842

@jomarko @gitgabrio @danielezonca Can you please review?


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/ScenarioSimulationEventHandler.java
```diff
@@ -382,7 +382,7 @@ public void onEvent(SetInstanceHeaderEvent event) {
         if (context.getModel().getSelectedColumn() == null) {
```
jomarko: This check is now duplicated in `public boolean isSameInstanceType(String className)` which is invoked immediately below.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/TestToolsPresenter.java
```diff
@@ -283,6 +284,12 @@ public void onPerfectMatchSearchedEvent(String search, boolean notEqualsSearch)
                 .filter(entry -> filterTerm(entry.getKey(), search, notEqualsSearch))
                 .forEach(filteredEntry -> addSimpleJavaInstanceListGroupItemView(filteredEntry.getKey(), filteredEntry.getValue()));
         updateSeparators();
+        /* If notEqualsSearch is TRUE, then the instance is not assigned for the selected column. Therefore, it isn't
+           necessary to search through the maps to check it.
+         */
+        if (!notEqualsSearch) {
+            checkInstanceIsAssigned(search);
```
jomarko: Could we update the method name to say we are as checking so doing an action inside? Or return just boolean from the current and do the action according the result at this place? What do you think? Do we mix two things together or not? I may be wrong.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModelTest.java
```diff
@@ -510,16 +520,25 @@ public void validatePropertyHeaderUpdate() throws Exception {
         commonValidatePropertyUpdate(1, true, true, true, false);
     }
 
+    public void isSameInstanceType(String columnClassName, String value, boolean expectedResult) {
```
jomarko: can be private I think

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/ScenarioSimulationEventHandler.java
yesamer: Done!

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/models/ScenarioGridModelTest.java
yesamer: Done!

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/TestToolsPresenter.java
yesamer: I my opinion, it's better to change the name only. In this context, I see the search and the assignation as an atomic action, there is no need to perform a search without assign the result to the related view. For the name what do you suggest?

jomarko:
@yesamer thank you for quick code updates. Looks good to me. 

I did also manual check, I think there is still way how the property column can be bound, even if it shouldn't be possible see my [comment](https://issues.jboss.org/browse/DROOLS-3842?focusedCommentId=13734181&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13734181). May be I understand something wrongly, please let me know.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/TestToolsPresenter.java
gitgabrio: @yesamer 
If this is the topic, I agree with @jomarko 
"updateInstanceIsAssignedStatus" -> IMO better reflect the scope of the method - "check" by itself does not mean "change something"

yesamer:
> @yesamer
> Please test it thoroughly - it looks fine but the code you touched could have a lot of downside effects I can not foresee. Beside that, just rename one method pls.
@gitgabrio 
Done, Thank you. I'll perform more additional tests before merge it.

yesamer:
@jomarko @gitgabrio During my retest I found an issue with the search box: if you try to put any string an exception is thrown. Now it's fixed. Can you please retest it? Thanks!

yesamer:
> Retested, approving, but https://issues.jboss.org/browse/DROOLS-4039 still present.

4039 will be managed in another PR.

yesamer:
@mantis hi, can you please squash and merge it? Thanks!

PR #1137, opened 2019-05-15
[DROOLS-3840] Extend test scenario runner to support additional parameters

danielezonca:
This PR enable additional parameters to rule test scenarios. It is possible now to specify KieSession name and RuleFlowGroup/AgendaGroup
+
It contains a small fix on Test Coverage report lifecycle

https://issues.jboss.org/browse/DROOLS-3840

@gitgabrio @Rikkola @jomarko 

PRs list:
- https://github.com/kiegroup/drools-wb/pull/1137
- https://github.com/kiegroup/drools/pull/2347
- https://github.com/kiegroup/droolsjbpm-knowledge/pull/383


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/test/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImplTest.java
```diff
@@ -404,4 +421,20 @@ public void editPomIfNecessaryTest() {
...
+        assertEquals(Type.DMN, model.getSimulation().getSimulationDescriptor().getType());
+        verify(dmnTypeServiceMock, times(1)).initializeNameAndNamespace(any(), any(), anyString());
+
+        when(dmnTypeServiceMock.initializeNameAndNamespace(any(), any(), anyString())).thenThrow(new ImpossibleToFindDMNException(""));
+
+        try {
+            service.load(path);
+        } catch (Exception e) {
+            fail();
+        }
```
jomarko: Maybe I miss something, why this test invoked the method `load` twice? Shouldn't we rather assert the loaded `model` content?

jomarko:
### Manual review notes
- The **Settings** dock panel **Save** button is confusing for me as it does the same save as the toolbar **Save** button + **Settings**, but the toolbar **Save** button do not save **Settings**. I think both should behave same or we should keep just one. From my point of view is also no ideal that if changes from **Settings** panel don't affect the scenario run, until they are changed. I think it is inconsistent for example with DMN and Stunner. There the dock panel display some properties specific for opened values. However there are values keept while the model is opened. In scenario simulation is enough to switch docks, and the unsaved **Settings** are lost. Please let me know if you see my point. See also the video attached to jira.
- In jira Klara mentions filename and scenario type. I think her intention was to get there file name of tested asset(for dmn)? I think this because the filename of scenario is already in toolbar?
- I think we should rename **Skip this simulation during the build** to **Skip this simulation during the test** because the Build button doesn't run scenarios. Or am I missing something?
- Question, is RuleFlowGroup and AgendaGroup the same as we have one input field for both? Then not sure why we added two methods into interfaces - rest of this PR ensemble.
- Question, KieSession, in the project I created new default kie base and default kie session. Then put this kie session in test scenario, however the test failed. do I need to update also rules?
- DMN, I think all properties need to be editable, because if the path to dmn is changed, then also 99% probable also model and namespace name will change
- Test coverage report is fixed, I mean switching dmn/rule, thank you.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/test/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImplTest.java
jomarko: @danielezonca any comments to this please?

danielezonca: Sorry I missed you comment :)

I didn't create a new method but it covers a different code path. It just verifies that the exception is not propagated

jomarko: Thank you

danielezonca:
jenkins retest this please

jomarko:
All points discussed on GChat and reported one more complex issue for investigation https://issues.jboss.org/browse/DROOLS-4046

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
```diff
@@ -196,7 +204,19 @@ public ScenarioSimulationModel load(final Path path) {
         try {
             final String content = ioService.readAllString(Paths.convert(path));
 
-            return ScenarioSimulationXMLPersistence.getInstance().unmarshal(content);
+            ScenarioSimulationModel scenarioSimulationModel = unmarshalInternal(content);
+            Simulation simulation = scenarioSimulationModel.getSimulation();
```
So, this looks like it is possible and valid to have a **ScenarioSimulationModel** _without_ a **Simulation**: is it true?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
```diff
@@ -82,6 +83,14 @@ public FactModelTuple retrieveFactModelTuple(Path path, String dmnPath) {
         return factModelTuple;
     }
 
+    @Override
+    public Simulation initializeNameAndNamespace(Simulation simulation, Path path, String dmnPath) {
```
gitgabrio: @danielezonca 
I would expect an "initialize" method to return a _void_; moreover, since it returns the same instance given as parameter, returning it is confusing. 
A lot of similar method (i.e. methods that accept an instance of a given type and returns the same type) actually returns new/different instances; that's the reason of above comment.

I think there are two options to makes it clearer:
1) return Simulation without accepting one (i.e. create a new one inside the method) - if possible;
2) return void - so it is clear that the method change the status of the provided Simulation.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
```diff
@@ -97,30 +104,29 @@ public void reset() {
     protected void setRuleSettings(SimulationDescriptor simulationDescriptor) {
         view.getDmnSettings().getStyle().setDisplay(Style.Display.NONE);
         view.getRuleSettings().getStyle().setDisplay(Style.Display.INLINE);
-        view.getDmoSession().setValue(simulationDescriptor.getDmoSession());
-        view.getKieBase().setValue(simulationDescriptor.getKieBase());
-        view.getKieSession().setValue(simulationDescriptor.getKieSession());
-        view.getRuleFlowGroup().setValue(simulationDescriptor.getRuleFlowGroup());
+        view.getKieSession().setValue(Optional.ofNullable(simulationDescriptor.getDmoSession()).orElse(""));
```
gitgabrio: @danielezonca 
Please keep the model name and the view' one in synchro: if the property to se/get in the model is "dmoSession", the element in the view should have the same name.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
```diff
@@ -97,30 +104,29 @@ public void reset() {
...
+        view.getDmnName().setInnerText(Optional.ofNullable(simulationDescriptor.getDmnName()).orElse(""));
+        view.getDmnNamespace().setInnerText(Optional.ofNullable(simulationDescriptor.getDmnNamespace()).orElse(""));
     }
 
     protected void saveRuleSettings() {
-        simulationDescriptor.setDmoSession(view.getDmoSession().getValue());
-        simulationDescriptor.setKieBase(view.getKieBase().getValue());
-        simulationDescriptor.setKieSession(view.getKieSession().getValue());
-        simulationDescriptor.setRuleFlowGroup(view.getRuleFlowGroup().getValue());
+        simulationDescriptor.setDmoSession(getCleanValue(() -> view.getKieSession().getValue()));
```
gitgabrio: @danielezonca 
See above comment

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
```diff
@@ -97,30 +104,29 @@ public void reset() {
...
     }
 
     protected void saveDMNSettings() {
-        simulationDescriptor.setDmnFilePath(view.getDmnFilePath().getInnerText());
-        simulationDescriptor.setDmnName(view.getDmnName().getInnerText());
-        simulationDescriptor.setDmnNamespace(view.getDmnNamespace().getInnerText());
+        simulationDescriptor.setDmnFilePath(getCleanValue(() -> view.getDmnFilePath().getValue()));
+    }
+
+    private String getCleanValue(Supplier<String> supplier) {
```
gitgabrio: @danielezonca 
?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenterTest.java
```diff
@@ -79,11 +75,8 @@ public void setup() {
         when(settingsViewMock.getSkipFromBuild()).thenReturn(skipFromBuildMock);
         when(settingsViewMock.getSaveButton()).thenReturn(saveButtonMock);
 
-        when(simulationDescriptorMock.getFileName()).thenReturn(FILE_NAME);
-        when(simulationDescriptorMock.getKieSession()).thenReturn(KIE_SESSION);
-        when(simulationDescriptorMock.getKieBase()).thenReturn(KIE_BASE);
         when(simulationDescriptorMock.getRuleFlowGroup()).thenReturn(RULE_FLOW_GROUP);
-        when(simulationDescriptorMock.getDmoSession()).thenReturn(DMO_SESSION);
+        when(simulationDescriptorMock.getDmoSession()).thenReturn(KIE_SESSION);
```
gitgabrio: @danielezonca 
Same comment: please keep names aligned

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
danielezonca: @gitgabrio 
I think you are referring to the next line 
`if(simulation != null...`
I don't think it should be possible but I added `simulation != null` in if condition just to be sure to not introduce regressions. The old code had no access to `Simulation` so the new method has the same behavior in case of `simulation == null`

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
danielezonca: This method converts empty strings to `null`. Do you prefer something like `getNullIfEmptyString`?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
danielezonca: Done as option 2

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
danielezonca: I cannot do it completely because this parameter **is** `KieSession` and not `dmoSession`. I can update the view but the label to show the user has to be `KieSession` so we can only decide where to break the consistency.
I have updated the view as requested but the proper way to fix it is to update the model too and migrate the files but it is too much effort to do now.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
danielezonca: Same as above :)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenterTest.java
danielezonca: See comment above

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
gitgabrio: @danielezonca 
OK, my doubt is: *if* we should not have simulation == null, and we are already checking that, would not be better to explicitly throw an Exception, to make it clear (reading the code) that simulation must not be null?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
gitgabrio: @danielezonca 
I'm just not sure if it is worth to add a new method for that, not on the name; anyway keep it as it is, fine for me

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
danielezonca: @gitgabrio 
I understand your doubt and I agree with the approach to highlight unexpected situation instead of silent them.
But my point is different: _if_ this can happen and it needs to be managed, it is already managed outside/somewhere else because I'm updating an existing method. This method loads the file and should never fail in a not managed way otherwise this prevents the user the possibility to fix something manually (i.e. dmn file path is wrong).
The meaning of this code inside the if is _optional:_ try to load additional information otherwise go on.

PR #1139, opened 2019-05-15
[DROOLS-4034] Add an Instance to a Scenario Column leads to an exception

yesamer:
https://issues.jboss.org/browse/DROOLS-4034

Open a new TestScenario editor.
Try to assign an instance to a column. Instance only, without a property (eg. Author).
An exception is thown.

Issue introduced with fix of ticket DROOLS-3890, please retest this ticket too.

@jomarko  @gitgabrio @danielezonca can you please review it?
Thanks.


jomarko:
@yesamer I have no comments to code. Also manual review passed. However it is strange that jira is about binding column and the touched code is about delete column event. Is that correct? Don't we link different jira and PR?

yesamer:
> @yesamer I have no comments to code. Also manual review passed. However it is strange that jira is about binding column and the touched code is about delete column event. Is that correct? Don't we link different jira and PR?

@jomarko I confirm you, the related JIRA ticket is correct. The point is that, when you try to add a fact in a column, the code deletes the selected column and creates a new column with selected column.

yesamer:
@manstis 

PR #1144, opened 2019-05-20
DROOLS-4049: PMML files are not listed in Library only in Project Explorer

manstis:
See https://issues.jboss.org/browse/DROOLS-4049

This PR adds a basic PMML (XML) editor to the workbench. It does not include a `NewResourceHandler` and hence does not appear in the list of "New Assets" to create a new file. This is deliberate as the editor is really basic and PMML is only being added for the demo. If you upload a PMML file it will appear in the Library (and filter by "Model"). It opens in a XML editor. The previous DMN PR was a quick hack to get PMMLs listed in the Library however it had issues and hence has been closed.

Part of an ensemble:
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/960
- https://github.com/kiegroup/drools-wb/pull/1144

A PR will be submitted for Business Central once we're happy with this approach.


drools-wb-screens/drools-wb-pmml-text-editor/drools-wb-pmml-text-editor-api/src/main/java/org/drools/workbench/screens/pmmltext/type/PMMLResourceTypeDefinition.java
```diff
@@ -0,0 +1,83 @@
...
+public class PMMLResourceTypeDefinition
+        implements ResourceTypeDefinition {
+
+    private Category category;
+
+    public PMMLResourceTypeDefinition() {
+    }
+
+    @Inject
+    public PMMLResourceTypeDefinition(final Model category) {
```
danielezonca: Why `Model`? I think `Decision` fit better.

drools-wb-screens/drools-wb-pmml-text-editor/drools-wb-pmml-text-editor-client/src/main/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,21 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee" bean-discovery-mode="all">
+  <scan>
+    <exclude name="org.drools.workbench.screens.pmmltext.client.**"/>
```
danielezonca: What does it mean? Is it an exclusion? How `PMMLResourceType` can work considering that is annotated as `@ApplicationScoped` but the package will be excluded?

drools-wb-screens/drools-wb-pmml-text-editor/drools-wb-pmml-text-editor-api/src/main/java/org/drools/workbench/screens/pmmltext/type/PMMLResourceTypeDefinition.java
manstis: This was based on [this](https://issues.jboss.org/browse/DROOLS-4049?focusedCommentId=13735451&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13735451) and subsequent comments.

drools-wb-screens/drools-wb-pmml-text-editor/drools-wb-pmml-text-editor-client/src/main/resources/META-INF/beans.xml
manstis: Well....... (other than a copy and paste of the same type of thing from other `-client` modules) I believe it prevents WELD (on the server) from scanning classes that are purely client-side and handled by Errai CDI.

manstis: Used mainly in `SuperDevMode` when both server and client classes are on the (IDE) classpath.

drools-wb-screens/drools-wb-pmml-text-editor/drools-wb-pmml-text-editor-api/src/main/java/org/drools/workbench/screens/pmmltext/type/PMMLResourceTypeDefinition.java
manstis: @etirelli @danielezonca I don't mind what category you'd prefer PMML to be in.. but please both agree! "Model" is traditionally Java classes, "Decision" is DMN, DRL etc. "Other" is anything else that we can't categorise more explicitly.

danielezonca: Fine for me `Model` too, no need to change :)

drools-wb-screens/drools-wb-pmml-text-editor/drools-wb-pmml-text-editor-client/src/main/resources/META-INF/beans.xml
danielezonca: Ok, this means Errai doesn't consider for scanning J2EE CDI configurations (= beans.xml) but only its own file. Nice to know :)

manstis: Yes, Errai uses `ErraiApp.properties` see [here](http://docs.jboss.org/errai/latest/errai/reference/html_single/#sid-19398999) for "bean scanning".

PR #1145, opened 2019-05-21
[DROOLS-3745] Data objects are enabled in the right panel after selecting a non-header cell of an unpopulated column

yesamer:
@jomarko @gitgabrio @danielezonca Can you please review it?

https://issues.jboss.org/browse/DROOLS-3745


yesamer:
@jomarko I agree with you, I consider both unrelated with the changes of this PR.

yesamer:
@manstis Hi! Can you please merge it on master?

PR #1146, opened 2019-05-21
[DROOLS-3806] Improve error reporting messages + minor bug

danielezonca:
This PR cover what is described in the corresponding ticket ( https://issues.jboss.org/browse/DROOLS-3806 ) plus some other improvements in error reporting area:
- (DMN) now if FEEL has a parsing error in Given section the error is properly reported and the evaluation is stopped
- (RULE) when an unsupported operator is used in Given section (i.e. `> 3`) a proper message is provideda and the cell is highlighted
- (General) alert error messages improvement: i.e. if the specified KieSession doesn't exist the error is what it come from the KieContainer (` Impossible to find a KieSession with name test`)
- (General) migrate scenario activator Junit class for project created before backend/api refactoring (move some of them into `drools` repo). You just need to open and save one scesim file
- (DMN) minor performance improvement on model evaluation: moved to new `KieRuntimeFactory` API to avoid creation of an unnecessary `KieSession` just to obtain `DMNRuntime` 

@jomarko @gitgabrio @Rikkola 

PS: PR description is probably longer than code changes :)

PRs list:
- https://github.com/kiegroup/drools-wb/pull/1146
- https://github.com/kiegroup/drools/pull/2355


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
```diff
@@ -347,15 +347,15 @@ protected void createActivatorIfNotExist(Path context) {
         Package junitActivatorPackage = getOrCreateJunitActivatorPackage(kieModule);
         final org.uberfire.java.nio.file.Path activatorPath = getActivatorPath(junitActivatorPackage);
 
-        if (!ioService.exists(activatorPath)) {
+        boolean removeExistingActivator = ensureDependencies(kieModule);
+
+        if (removeExistingActivator || !ioService.exists(activatorPath)) {
+            removeOldActivatorIfExists(activatorPath, kieModule);
```
jomarko: I am missing something, method says if exists `removeOldActivatorIfExists` but the if that wrap, says if it doesn't exists `!ioService.exists(activatorPath)`, to me seems like contradiction.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
```diff
@@ -394,13 +396,17 @@ protected void ensureDependencies(KieModule module) {
             toSave |= removeFromPomIfNecessary(dependencies, oldDependency);
         }
 
+        boolean activatorToRecreate = toSave;
```
jomarko: Related to single responsibility of the method. I think in this method we should name the variable independently from `activator`. I think it is just side effect we remove `activator` according the change of dependencies. What do you think?

jomarko:
### Manual review
- Maybe not an issue. Just bringing to you attention as it brings change in behavior. As we know handle more exceptions in the backend, we lost dialog for fixing expected dmn strings with missing quotes. If there is expected column of dmn string and the cell value has no quotes, now there is dialog informing about unexpected variable - troubles with parsing, previously the dialog offered to fix value by adding quotes.
- [point one] Not sure if it is fixed, in the given section I have dmn number column, I put "abc" to cell, no compilation/parsing error shown
- [point two] :heavy_check_mark: 
- [point three] :heavy_check_mark: 
- [point four] not sure how to check
- [point five] not sure how to check

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
danielezonca: I added some comments to clarify. There are two reasons why the activator needs to be saved, if it doesn't exist or if it needs to be migrated (I changed `removeExistingActivator` variable to `needMigrateActivatorIfExists`)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
danielezonca: Correct, I renamed the variable to `oldDependenciesExist` and added a javadoc

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
jomarko: @danielezonca sorry, my mind is blank today, I can not understand this. In `if` we had statement `!ioService.exists(activatorPath)` that suggests me without any deep investigating, it is evaluated to `true` if `activatorPath`  does not exist and then `removeOldActivatorIfExists(activatorPath, kieModule)` tries to remove something at the same path, that we evaluated previously as it does not exist? I think I am missing something.

kkufova:
I'll take a look at this as well.

@kkufova, just so you know.

danielezonca:
This should solve also https://issues.jboss.org/browse/DROOLS-4074

danielezonca:
> ### Manual review
> * Maybe not an issue. Just bringing to you attention as it brings change in behavior. As we know handle more exceptions in the backend, we lost dialog for fixing expected dmn strings with missing quotes. If there is expected column of dmn string and the cell value has no quotes, now there is dialog informing about unexpected variable - troubles with parsing, previously the dialog offered to fix value by adding quotes.

Before this fix the internal value considered by DMN was `null` so it could be possible to have a successful scenario because of an _implicit_ `null` not visible to the user. Now if there is any parsing problem, it is visible
> * [point one] Not sure if it is fixed, in the given section I have dmn number column, I put "abc" to cell, no compilation/parsing error shown

Correct because by default DMN engine has type check disabled (see https://docs.jboss.org/drools/release/7.22.0.Final/drools-docs/html_single/index.html#dmn-properties-ref_dmn-models ) so only parsing errors are reported (and "abc" is a valid string). I tried to enable it with ticket https://issues.jboss.org/browse/DROOLS-3730 but I had to revert because the type check performed is too strict (it fails if a column is missing)

> * [point four] not sure how to check

Clarified via chat

> * [point five] not sure how to check

Clarified via chat

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
danielezonca: The method `removeOldActivatorIfExists` looks for existing activators in multiple paths, the idea is to be sure that after that method no existing activators are in the project.
`!ioService.exists(activatorPath)` verify only if an activator exists in the target position

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioSimulationServiceImpl.java
```diff
@@ -368,8 +370,17 @@ protected Package getOrCreateJunitActivatorPackage(KieModule kieModule) {
...
-    protected void removeOldActivatorIfExists(KieModule kieModule) {
+    /**
+     * This routine looks for existing activators to migrate
+     * @param activatorPath
+     * @param kieModule
+     */
+    protected void removeOldActivatorIfExists(org.uberfire.java.nio.file.Path activatorPath, KieModule kieModule) {
 
+        // migration step after Test Scenario runner modules split DROOLS-3389
+        ioService.deleteIfExists(activatorPath);
```
This is related somehow to @jomarko doubts: expanding the code from _createActivatorIfNotExist_ and _removeOldActivatorIfExists_ we get the following (pseudo) code:
` if ( !ioService.exists(activatorPath)) ->   ioService.deleteIfExists(activatorPath);`

Maybe is needed/usefull because of other usage, but it actually looks _strange_

danielezonca: I agree it could be a misleading but the logic is a bit different, the external condition is
`if (needMigrateActivatorIfExists || !ioService.exists(activatorPath))`
So the other method with the `deleteIfExists` can be invoked if the file doesn't exist (= useless call) but also if the variable `needMigrateActivatorIfExists` is true.
That variable is true if the project contains any "old dependencies".

The general problem is we should have a migration mechanism also for dependencies and activator but we have not. There are four possible statuses:
- empty project: nothing to migrate, just add dependencies and activator
- first version of activator: need to find where the activator is and remove it (now is in `testscenario` package)
- second version of activator: need to update dependencies and activator too
- current version: nothing to migrate

The general idea in the code is:
- first update/add/remove dependencies
- if I had to update the dependencies or the activator does not exist try to remove existing activator (it could be in different paths)
- create the new activator

danielezonca:
@manstis 
Can you please squash and merge these PRs?
- https://github.com/kiegroup/drools-wb/pull/1146
- https://github.com/kiegroup/drools/pull/2355

PR #1147, opened 2019-05-21
DROOLS-4061 : Guided Score card rules not executed via test scenario.

Rikkola:
https://issues.jboss.org/browse/DROOLS-4061

https://github.com/kiegroup/drools/pull/2356
https://github.com/kiegroup/kie-wb-common/pull/2657
https://github.com/kiegroup/drools-wb/pull/1147

### DIFF
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/main/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerService.java
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/main/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerService.java
@@ -193,6 +193,14 @@ public class ScenarioRunnerService
             } else {
                 ksession = sessionService.newKieSession(module,
                                                         ksessionName);
+
+                if (isPMMLScenario(module)) {
+                    /* There is an issue with Scorecards where the first run does not work.
+                    * This is because the tests only passes with clones, this forces the test run to use clones.
+                     */
+                    ksession = sessionService.newKieSession(module,
+                                                            ksessionName);
+                }
             }
         } catch (Exception e) {
             // If for one reason or another we can not load the ksession. Return null
@@ -200,5 +208,9 @@ public class ScenarioRunnerService
         }
 
         return ksession;
+    }
+
+    private boolean isPMMLScenario(KieModule module) {
+        return module.getModuleName() != null;
     }
 }
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/VerifyFactsPanel.java
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/VerifyFactsPanel.java
@@ -26,6 +26,7 @@ import org.drools.workbench.models.testscenarios.shared.Fixture;
 import org.drools.workbench.models.testscenarios.shared.FixtureList;
 import org.drools.workbench.models.testscenarios.shared.Scenario;
 import org.drools.workbench.models.testscenarios.shared.VerifyFact;
+import org.drools.workbench.models.testscenarios.shared.VerifyScorecardScore;
 import org.drools.workbench.screens.testscenario.client.resources.i18n.TestScenarioConstants;
 import org.gwtbootstrap3.client.ui.Button;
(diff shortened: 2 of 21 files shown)


Rikkola:
Not to myself. Hide the scorecard name selector if the experimental editor is hidden.

Rikkola:
@jomarko , @lanceleverich and all. This adds selector into the test scenario configuration tab. Only insert and assert for the inserted Facts are supported. There is still an issue that needs to be fixed related to broken asserts. I tried few things and could not get any data out of the scorecard run.

jomarko:
Unused dependencies Rikkola#1

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/page/settings/ScorecardSelector.java
```diff
@@ -0,0 +1,95 @@
...
+
+                    if (scenario.getModelName() != null) {
+                        view.setSelected(scenario.getModelName());
+                    }
+                }).find(path,
+                        scenario.getPackageName());
+    }
+
+    public void onValueSelected(final String modelName) {
+        if (NONE.equals(modelName)) {
```
jomarko: `Objects.equals(NONE, modelName)` would be more safe.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/page/settings/ScorecardSelectorViewImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    @Override
+    public void init(final ScorecardSelector presenter) {
+        listBox.setWidth("202px");
+        listBox.addChangeHandler(event -> presenter.onValueSelected(listBox.getSelectedValue()));
+    }
+
+    @Override
+    public void setSelected(final String modelName) {
+        for (int i = 0; i < listBox.getItemCount(); i++) {
+            if (listBox.getItemText(i).equals(modelName)) {
```
jomarko: Could we please use `Objects.equals`?

jomarko:
### Manual review
Passed

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-backend/src/main/java/org/drools/workbench/screens/testscenario/backend/server/ScenarioRunnerService.java
```diff
@@ -193,6 +193,14 @@ private KieSession loadKSession(KieModule module,
...
                 ksession = sessionService.newKieSession(module,
                                                         ksessionName);
+
+                if (isPMMLScenario(module)) {
+                    /* There is an issue with Scorecards where the first run does not work.
+                    * This is because the tests only passes with clones, this forces the test run to use clones.
+                     */
+                    ksession = sessionService.newKieSession(module,
+                                                            ksessionName);
+                }
```
Rikkola: Not the clean way to fix this, but the fix that works with the time we have.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/VerifyFactsPanel.java
```diff
@@ -61,27 +62,46 @@ public VerifyFactsPanel( FixtureList verifyFacts,
...
         }
-    }
 
+        private Object getName(Fixture fixture) {
+            if (fixture instanceof VerifyFact) {
+                return ((VerifyFact) fixture).getName();
+            } else if (fixture instanceof VerifyScorecardScore) {
+                return TestScenarioConstants.INSTANCE.ScoreCardScore();
+            } else {
+                return "";
```
danielezonca: This branch has been added by this PR, before that the code was `setText("'" + verifyFact.getName() + "'");`. Are you sure about that? If this should not happen I would prefer to throw an exception as `else`. Wdyt?

Rikkola: Yes here exception makes sense. This should always be either of the two types that are in the top if clauses.

Rikkola:
Might have to abandon the idea of getting a full downstream build. Looks like I start getting merge conflicts.

PR #1149, opened 2019-05-23
[DROOLS-3855] Unsupported DMN asset shows each item two times

yesamer:
@jomarko  @gitgabrio @danielezonca Can you please review it?

The error checking is perfmormed 2 times, so I simply replaced the List with a Set to avoid duplicates.

https://issues.jboss.org/browse/DROOLS-3855

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
@@ -16,11 +16,12 @@
 
 package org.drools.workbench.screens.scenariosimulation.backend.server;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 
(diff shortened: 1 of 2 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
```diff
@@ -61,7 +62,6 @@ public FactModelTuple retrieveFactModelTuple(Path path, String dmnPath) {
         for (InputDataNode input : dmnModel.getInputs()) {
             DMNType type = input.getType();
             checkTypeSupport(type, errorHolder, input.getName());
-            visitCompositeType(type, false, errorHolder, input.getName());
```
jomarko: @yesamer could you please explain why we remove this `visitCompositeType`? I thought in my review we just discuss `HashSet` vs `TreeSet`

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
```diff
@@ -71,7 +71,6 @@ public FactModelTuple retrieveFactModelTuple(Path path, String dmnPath) {
         for (DecisionNode decision : dmnModel.getDecisions()) {
             DMNType type = decision.getResultType();
             checkTypeSupport(type, errorHolder, decision.getName());
-            visitCompositeType(type, false, errorHolder, decision.getName());
```
jomarko: @yesamer could you please explain why we remove this `visitCompositeType`? I thought in my review we just discuss `HashSet` vs `TreeSet`

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNTypeServiceImpl.java
yesamer: @jomarko Because, after a discussion with Gabriele, we discovered that   `checkTypeSupport` method simply calls `visitCompositeType` with the same parameters, and this generate the issue: `visitCompositeType` is called 2 times.

yesamer:
@manstis Hi! Can you please merge it on master? Thanks!

PR #1160, opened 2019-05-31
[DROOLS-4060] CSS bug in Test Scenario setting panel when namespace is long

yesamer:
@kkufova @gitgabrio @danielezonca @jomarko Can you please review it?

https://issues.jboss.org/browse/DROOLS-4060


kkufova:
I agree with @danielezonca, I like this approach a lot, but I would prefer if **all** the properties that **cannot** be edited looked the same (like this).

yesamer:
@danielezonca @kkufova Done, can you please check it?

yesamer:
@manstis  Can you please merge it on master?

yesamer:
@manstis Can you please merge it on master?

PR #1161, opened 2019-06-04
[DROOLS-3744] "Value" is not highlighted in the right panel

yesamer:
@kkufova @gitgabrio @danielezonca Can you please review it?

I simply added the property title (`fieldName`) in the `propertyNameElements` list in case the baseClass is a simple object (this means it will be added in any case). `isSimple` method is removed because unused.

https://issues.jboss.org/browse/DROOLS-3744

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
@@ -133,10 +133,11 @@ public class ScenarioSimulationGridHeaderUtilities {
             List<String> propertyNameElements = null;
             if (scenarioGridColumn.isPropertyAssigned()) {
                 final Optional<Simulation> optionalSimulation = scenarioGrid.getModel().getSimulation();
-                if (ScenarioSimulationUtils.isSimpleJavaType(scenarioGridColumn.getFactIdentifier().getClassName()))
+                if (ScenarioSimulationUtils.isSimpleJavaType(scenarioGridColumn.getFactIdentifier().getClassName())) {
                     propertyNameElements = Arrays.asList(VALUE);
-                else
+                } else {
                     propertyNameElements = optionalSimulation.map(simulation -> getPropertyNameElements(simulation, uiColumnIndex)).orElse(null);
+                }
             }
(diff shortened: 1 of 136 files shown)


yesamer:
@gitgabrio  According to the bug description: in case of a simple data object added to to the scenario, the property title "value" is not highlighted when selected. The reason of this behaviour is in line 377: if a simple class is handled, only the class is taken. And this is wrong: the field name (value) should be added too.

yesamer:
@kkufova Please wait before test it

kkufova:
@yesamer, it is still not working:

bug1

yesamer:
@kkufova Basically I found the same bug on RULE scenario test. I fixed it in my previous commits. Now, I found the problem on DMN side and fixed it too. I kindly ask you to test the fixes on both RULE / DMN cases. I suggest to test nested facts too, just to be sure. 
Moreover, this PR fix https://issues.jboss.org/browse/DROOLS-4078 as well. Can you please verify it?
Thanks

kkufova:
@yesamer, is this ready for another review?

yesamer:
@kkufova Not yet :)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationGridHeaderUtilities.java
```diff
@@ -151,10 +152,18 @@ public static String getExistingInstances(final String group, final ScenarioGrid
...
 
-    public static List<String> getPropertyNameElements(final Simulation simulation, final int columnIndex) {
-        return Collections.unmodifiableList(simulation.getSimulationDescriptor().getFactMappingByIndex(columnIndex).getExpressionElementsWithoutClass()
-                                                    .stream()
-                                                    .map(ExpressionElement::getStep)
-                                                    .collect(Collectors.toList()));
+    public static List<String> getPropertyNameElements(final ScenarioGridModel scenarioGridModel, final int columnIndex) {
+        final Optional<Simulation> optionalSimulation = scenarioGridModel.getSimulation();
+        return optionalSimulation.map(simulation -> {
+            final FactMapping factMapping = scenarioGridModel.getSimulation().get().getSimulationDescriptor().getFactMappingByIndex(columnIndex);
```
danielezonca: You can use `simulation` here instead of access again via `scenarioGridModel.getSimulation()...`

yesamer: @danielezonca Done, thank you!

PR #1166, opened 2019-06-06
DROOLS-4042: [DMN Designer] Add support for importing and consuming PMML models

manstis:
See https://issues.jboss.org/browse/DROOLS-4042

Further too https://github.com/kiegroup/kie-wb-common/pull/2683 `drools-wb-webapp` needs a JAXB implementation in order for the PMML integration with DMN to function correctly. `kie-wb-distributions` has a JAXB implementation bundled with it by virtue of depending on `kie-pmml` too.

### DIFF
--- a/drools-wb-webapp/pom.xml
+++ b/drools-wb-webapp/pom.xml
@@ -1068,6 +1068,15 @@
       <groupId>org.kie.workbench</groupId>
       <artifactId>kie-wb-common-dmn-backend</artifactId>
     </dependency>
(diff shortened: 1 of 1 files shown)


manstis:
Jenkins please retest this.

manstis:
@jomarko AFAIK the JAXB libraries contain the implementation of the interfaces used at compile time. Resolution of the implementation is performed at runtime. IDK if `dependency:analyze` covers that? If you remove these libraries you'll get an error relating to being unable to load a class (in these dependencies). `kie-wb-distributions` includes these libraries as part of the `kie-pmml` dependency.

PR #1173, opened 2019-06-20
[DROOLS-4180] Invoking "duplicate instance" from a row causes unexpected error

yesamer:
@kkufova @gitgabrio @danielezonca please review and test it.

To solve the issue, we defined the following updated behaviour:
Columns actions are now present ONLY when you open the menu starting from a HEADER cell.
Therefore, When you right-click from a BODY cell, you will have ONLY items related to row actions.

https://issues.jboss.org/browse/DROOLS-4180


yesamer:

kkufova:
@kaldesai, FYI, this will require a change in the documentation.

kaldesai:
@kkufova, Ack. I will have to identify where it will impact the document.

PR #1175, opened 2019-06-21
[DROOLS-4072]: Search query is not empty but all data objects listed

yesamer:
@kkufova @gitgabrio @danielezonca Can you please review and test it?

https://issues.jboss.org/browse/DROOLS-4072


kkufova:
@danielezonca, I think this can be merged.

PR #1176, opened 2019-06-26
[DROOLS-4094] Replace DmnName input element with DropDown widget in SettingsDock Panel

yesamer:
@kkufova @gitgabrio @danielezonca please review and test it.

https://issues.jboss.org/browse/DROOLS-4094

@kkufova We can have a chat or a call tomorrow morning in order to describe you the changes I made and to suggest you how to cover the possible cases.

PR related: https://github.com/kiegroup/drools/pull/2416


yesamer:
@gitgabrio @kkufova done, you can proceed.

kkufova:
@yesamer, I noticed the following thing: if you change the DMN model, the table layout is not modified as it should be. See this [screencast](https://issues.jboss.org/secure/attachment/12452759/screencast-01-07-19-1.webm). I think this is a bug, as I don't see why else would we offer this feature if the layout cannot be changed.

yesamer:
@kkufova Thank you for your feedback. About  the issue you reported: consider that it is ALREADY present on master: if you try to manually change the DMN file in the input box, you will report the same behaviour. This means this PR didn't introduce it.
Regarding the issue you reported: In my understanding, it's correct that, when you change DMN file, the table layout remains the same, but only the facts list in left right bar should change. And this happens according your screencast.

kkufova:
@yesamer, OK, sorry about that, I haven't noticed that this issue is already on master. However, I am not sure about what exactly the correct behavior should be. Let's ask @danielezonca; after changing the source DMN model, should the generated table layout be generated again, or should it stay the same without changes?

My problem with the current approach is that once you change the DMN model, it doesn't really make sense to keep the old table, as you cannot use it anymore. The problematic part is that if it contains any data, they will be lost, even if you change the model and than change it back, which can happen. But if you don't want to change it back, you need to remove all the data from the table manually. What is the real use case for changing the DMN model? Do you want to change it to something completely different, or to something similar?

danielezonca:
@kkufova 
I confirm this is the current behavior because there is no "template migration" strategy in place. We have this other ticket to do that https://issues.jboss.org/browse/DROOLS-3570.
Current behavior is covering only these use cases:
- user moved DMN model to another place so he needs to update the path in the test scenario
- user has another DMN model with the same input/decisions and he wants to run the test scenario on it (i.e. new decision logic to test and see if equivalent)

We can add a small note next to the widget but I would like to keep this PR focused on replace the widget only

yesamer:
@kkufova I agree with your evaluation: we need to handle this situation in some way, we can use the ticket you opened to start a discussion about a possible solution.

kkufova:
@danielezonca, could you please review the PRs and merge them?

PR #1179, opened 2019-07-02
DROOLS-4285: Slow build due to inefficient EnumServiceImpl.doValidation() call

manstis:
See https://issues.jboss.org/browse/DROOLS-4285

The change uses an existing cache for the `ClassLoader` as opposed to retrieving it from `KieModuleMetaData` everytime it's needed. Existing CDI based tests - on `7.23.x` - continue to pass (i.e. the tests actually run the code as opposed to mocking it to death). Obviously this won't build or run on `master` until the `ip-bom` stuff is resolved.


manstis:
Jenkins please retest this.

PR #1184, opened 2019-07-08
Drools 4298 - DO_NOT_MERGE - 

gitgabrio:
@manstis @danielezonca @yesamer @Rikkola 

See https://issues.jboss.org/browse/DROOLS-4298

Scope of this pr should be to improve reliability/predictability of drools-wb-webapp behavior.

Namely, the basic idea is that when someone build/starts drools-wb-webapp: 

1. he/she must be sure of which dependency artifact (locally built? remotely downloaded?) is currently using, 
2. he/she must be able to specify the artifact origin (locally built? remotely downloaded?) to use

This is valid *only* for the drools-wb modules (all other artifacts keep following standard rules)

This one has to eventually be merged after 4294

### DIFF
--- a/drools-wb-webapp/pom.xml
+++ b/drools-wb-webapp/pom.xml
@@ -37,6 +37,8 @@
     <errai.jboss.home>${project.build.directory}/wildfly-${version.org.jboss.errai.wildfly}</errai.jboss.home>
     <gwt.helper.includes>Client,BusinessCentral</gwt.helper.includes>
     <gwt.helper.rootDirectories>${project.parent.basedir}/drools-wb-screens</gwt.helper.rootDirectories>
+    <!-- Horrible workaround used to deploy to local-maven-repo dir -->
+    <root-base-dir>${project.parent.basedir}</root-base-dir>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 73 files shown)


jomarko:
@gitgabrio Hi just two comments form me:
- could we please use the format of properties according the existing ones? I mean to use `root.base.dir` rather than `root-base-dir` just for the sake of consistency
- If I read `Horrible workaround` really doesn't make me trust the solution, I would omit the word `Horrible`

pom.xml
```diff
@@ -39,6 +39,10 @@
     <checkstyle.logViolationsToConsole>true</checkstyle.logViolationsToConsole>
     <jacoco.haltOnFailure>true</jacoco.haltOnFailure>
     <spotbugs.failOnViolation>true</spotbugs.failOnViolation>
+    <!-- This property has to be overridden by each module -->
+    <root-base-dir>${project.basedir}</root-base-dir>
```
manstis: Can you not consider using `${maven.multiModuleProjectDirectory}` instead - you'd probably not have to override for every module then either. See https://stackoverflow.com/questions/3084629/finding-the-root-directory-of-a-multi-module-maven-reactor-project for some different approaches. `multiModuleProjectDirectory` seems to work OK as of Maven [3.3.9](https://issues.apache.org/jira/browse/MNG-5786).

gitgabrio: @manstis 
Already tried, of course, but it does not works.

drools-wb-webapp/pom.xml
```diff
@@ -1862,6 +1869,30 @@
...
+<!--    <repository>-->
+<!--      <snapshots>-->
+<!--        <enabled>false</enabled>-->
+<!--      </snapshots>-->
+<!--      <id>central</id>-->
+<!--      <name>Central Repository</name>-->
+<!--      <url>https://repo.maven.apache.org/maven2</url>-->
+<!--    </repository>-->
+<!--  </repositories>-->
+
```
Rikkola: Is this something that will get uncommented at some situations?

pom.xml
```diff
@@ -225,7 +237,133 @@
...
+              <verbose>true</verbose>
+              <!--
+               <regexIgnoredClass>org\.jboss\.errai\.marshalling\.server\.impl\.ServerMarshallingFactory.*</regexIgnoredClass>
+                    <regexIgnoredClass>io\.fabric8\.kubernetes.*</regexIgnoredClass>
+                    <regexIgnoredClass>io\.fabric8\.openshift.*</regexIgnoredClass>
+                    <regexIgnoredClass>com\.sun\.net\.httpserver\..+</regexIgnoredClass>
+                    <regexIgnoredClass>javax\..+</regexIgnoredClass>
+                    <regexIgnoredClass>org\.w3c\.dom\..+</regexIgnoredClass>
+                    <regexIgnoredClass>org\.xml\.sax\..+</regexIgnoredClass>
+              -->
```
Rikkola: Or this?

gitgabrio:
I'll take care of them

Il lun 15 lug 2019, 13:08 Toni Rikkola <notifications@github.com> ha
scritto:

PR #1186, opened 2019-07-09
remove illegalTransitiveDependencyCheck override

danielezonca:
Remove override
@gitgabrio @mbiarnes @yesamer

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -203,28 +203,6 @@
           </execution>
         </executions>
       </plugin>
-
-      <plugin>
-        <artifactId>maven-enforcer-plugin</artifactId>
-        <version>3.0.0-M1</version>
-        <executions>
-          <execution>
-            <id>enforce-direct-dependencies</id>
-            <phase>verify</phase>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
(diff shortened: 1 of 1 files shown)


danielezonca:
@gitgabrio 
We discovered that even with this configuration this enforcer check has another side effect that for some reason prevent the artifact to be installed
`[INFO] --- maven-install-plugin:2.5.2:install (default-install) @ drools-wb-globals-editor-api ---`
`[INFO] No primary artifact to install, installing attached artifacts instead.`
While disabling it entirely you will see
`[INFO] --- maven-install-plugin:2.5.2:install (default-install) @ drools-wb-globals-editor-api ---`
`[INFO] Installing /home/dzonca/IdeaProjects/KieGroup/drools-wb/drools-wb-screens/drools-wb-globals-editor/drools-wb-globals-editor-api/target/drools-wb-globals-editor-api-7.24.0-SNAPSHOT.jar to /home/dzonca/.m2/repository/org/drools/drools-wb-globals-editor-api/7.24.0-SNAPSHOT/drools-wb-globals-editor-api-7.24.0-SNAPSHOT.jar`

PR #1187, opened 2019-07-10
[7.23.x] DROOLS-4061 : Guided Score card rules not executed via test scenario.

Rikkola:
(cherry picked from commit 9159a7b6d0f61bcb3abebc199110462fa44e0148)

https://issues.jboss.org/browse/DROOLS-4061

@danielezonca @jomarko Here is the 7.23.x version for reviews.

Bundle:
https://github.com/kiegroup/drools/pull/2445
https://github.com/kiegroup/kie-wb-common/pull/2743
https://github.com/kiegroup/drools-wb/pull/1187


jomarko:

jomarko:
@Rikkola maybe I miss something, but seems like not working on 7.23.x:
I created data object [1], guided score card [2] and test scenario [3], but it fail. Can you spot some issue from screenshots? I am attaching the project also to the jira if it helps.

[1]
Screenshot from 2019-07-11 15-41-45

[2]
Screenshot from 2019-07-11 15-42-23

[3]
Screenshot from 2019-07-11 15-43-03

Rikkola:
I did some debugging and the test run gives me "ERROR-2" and based on the source code this means the run could not find the rule unit. I think the fix is missing a commit, but what, since I think this bundle was the only one that was missing from 7.23.x. Asking around and looking at the commit histories.

Rikkola:
Also the war fails every time with this, while the unit test for drools that runs this same setup again works.

Rikkola:
@kiegroup/gatekeepers Can we merge this in? The same bundle for master is not yet in due what at this point looks to be randomly failing tests, not sure if we need that in first.

PR #1193, opened 2019-07-16
[DROOLS-4215] Single instance can be mapped multiple times using data object search

yesamer:
@kkufova @gitgabrio @danielezonca can you please review and test it?

https://issues.jboss.org/browse/DROOLS-4215

Basically, this fix disables the ADD button on left tool bar if the instance (fact) is already assigned. 
Additionally, even in the case the button is enabled (eg manually), a check validate if the action is permitted, and in particular:
- When the instance is still not present in the same group of the selected column;
- When the instance is present but the selected column is near (in the right or in the left) of the column with the same instance.

@kkufova This fix impacts on SetInstance, SetProperty and DuplicateInstance. I suggest you to test these functionalities.


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/ListGroupItemPresenter.java
```diff
@@ -42,18 +42,21 @@
 
     private AtomicBoolean disabled = new AtomicBoolean(true);
 
-    private String factName = null;
+    /* In case of a selected column with <b>assigned</b> Instance, it contains its related Instance title.
+       Otherwise, in case of a <b>NOT assigned</b> Instance, it contains all the titles assigned to other columns of
+       the same group with the following syntax: "TITLE1;TITLE2;.." */
+    private String filterTerm = null;
```
yesamer: Here, I just renamed 'factName' to 'filterTerm'. Considering the description I put in the comment, it fits better, in my opinion.

kkufova:
@yesamer, there are two relevant test failures, could you please take a look?

yesamer:
@kkufova done :)

kkufova:
We agreed with @yesamer to divide the problem into two parts:

In the situation where all the data objects are already mapped in the table:
1. once a data object is searched using the search button, it appears in the right panel. However, it cannot be used again, so the **Add** button is disabled.
2. Once the user presses the cross icon in the search box to cancel the search (or alternatively, if the user deletes the search box and presses Enter), the right panel needs to go back to the state _before_ the search (_i.e._ empty right panel). Currently, the right panel shows all the data objects in this situation.

kkufova:
Also, these is a bug in the PR: selecting a nested property in the right panel causes an unexpected error:

```
Uncaught exception: Exception caught: (TypeError) : Cannot read property 'Qwg' of undefined Caused by: (TypeError) : Cannot read property 'Qwg' of undefined
```

See [this](https://issues.jboss.org/secure/attachment/12453339/screencast-17-07-19-1.webm) screencast.

yesamer:
@kkufova Reported bug: FIXED. Point 2) implemented. :)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractSelectedColumnCommand.java
```diff
@@ -310,4 +314,55 @@ protected String getPropertyHeaderTitle(ScenarioSimulationContext context, List<
                 .findFirst()
                 .map(FactMapping::getExpressionAlias);
     }
+
+    /**
+     * It checks if the provided instance title is valid. A title is valid if not already present in his own column group
+     * (GIVEN/EXPECTED). The only exception is when the selected column is part (i.e on the left or on the right) of
```
danielezonca: Verify that in DMN test scenarios each object can be assigned only one time in the whole header.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractSelectedColumnCommand.java
```diff
@@ -310,4 +314,55 @@ protected String getPropertyHeaderTitle(ScenarioSimulationContext context, List<
...
+     * @param title It's the title to assign to selectedColumn
+     * @return
+     *
+     */
+    protected boolean isValidInstanceTitle(ScenarioSimulationContext context,
+                                           ScenarioGridColumn selectedColumn,
+                                           FactIdentifier factIdentifier,
+                                           String title) {
+        String columnGroup = selectedColumn.getInformationHeaderMetaData().getColumnGroup();
+        String groupName = columnGroup.contains("-") ? columnGroup.substring(0, columnGroup.indexOf("-")) : columnGroup;
```
danielezonca: Can you please explain the reason of this substring?

yesamer: I refactored this part as agreed, now it should be easy to handle this subColumnGroup name handling.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/actualcommands/AbstractSelectedColumnCommand.java
yesamer: @danielezonca check removed as agreed.

yesamer:
jenkins retest this please

yesamer:
@danielezonca I think we can merge it, wdyt?

yesamer:
@manstis @kiegroup/gatekeepers Hi, can you please merge it? Thanks!

PR #1194, opened 2019-07-17
DROOLS-2548 : java.lang.IllegalStateException: Unknown PartType: dialect

Rikkola:
https://issues.jboss.org/browse/DROOLS-2548

https://github.com/kiegroup/kie-wb-common/pull/2762
https://github.com/kiegroup/drools-wb/pull/1194

### DIFF
--- a/drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-backend/src/test/java/org/drools/workbench/screens/guided/template/server/indexing/IndexGuidedRuleTemplateAttributesTest.java
+++ b/drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-backend/src/test/java/org/drools/workbench/screens/guided/template/server/indexing/IndexGuidedRuleTemplateAttributesTest.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import org.apache.lucene.search.Query;
+import org.drools.workbench.models.datamodel.rule.RuleAttribute;
 import org.drools.workbench.models.guided.template.backend.RuleTemplateModelXMLPersistenceImpl;
 import org.drools.workbench.models.guided.template.shared.TemplateModel;
 import org.drools.workbench.screens.guided.template.type.GuidedRuleTemplateResourceTypeDefinition;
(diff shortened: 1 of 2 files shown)


drools-wb-screens/drools-wb-guided-template-editor/drools-wb-guided-template-editor-backend/src/test/java/org/drools/workbench/screens/guided/template/server/indexing/IndexGuidedRuleTemplateAttributesTest.java
```diff
@@ -49,6 +50,13 @@ public void testIndexGuidedRuleTemplateAttributes() throws IOException, Interrup
                                                                                           add(new Import("org.drools.workbench.screens.guided.template.server.indexing.classes.Mortgage"));
                                                                                       }},
                                                                                       "template1");
+
+
+        model.addAttribute(new RuleAttribute("salience",
```
manstis: Should we also assert that there is no index entries for these attributes? 

Otherwise I don't see any value in adding these lines to the test.

Rikkola: Right now they pick up if the test throws an exception, but I'll add assertion with comments saying these are not indexed.

Rikkola: Yes that was what I was meaning with the "assertion". I'll do that.

Rikkola: @manstis ok so can't search with "salience" since it is not a recognized PartType.

Bonuseto:

Bonuseto:

PR #1200, opened 2019-07-26
DROOLS-4369: Fix the Cancel button styling - Add Global Variable dialog

jomarko:
For more details see: https://issues.jboss.org/browse/DROOLS-4369

@Bonuseto hello, may I ask you for a review? Please check both Firefox and Chrome.

### DIFF
--- a/drools-wb-screens/drools-wb-globals-editor/drools-wb-globals-editor-client/src/main/java/org/drools/workbench/screens/globals/client/editor/AddGlobalPopupViewImpl.html
+++ b/drools-wb-screens/drools-wb-globals-editor/drools-wb-globals-editor-client/src/main/java/org/drools/workbench/screens/globals/client/editor/AddGlobalPopupViewImpl.html
@@ -29,7 +29,7 @@
             </div>
         </form>
         <div class="modal-footer">
-            <button id="cancelButton" data-i18n-key="Cancel" class="btn-cancel"></button>
+            <button id="cancelButton" data-i18n-key="Cancel" class="btn btn-cancel"></button>
             <button id="addButton" class="btn btn-primary" data-i18n-key="AddGlobal"></button>
         </div>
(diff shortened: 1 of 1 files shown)


Bonuseto:

jomarko:
There are unrelated failed tests, however artifacts (war files) are generated, so should be ready for review.

Bonuseto:
Hi @jomarko ,sorry I already did review on Friday, but totally forgot to leave comment here, also I'm not set as reviewer here, but see below the result I guess it's right one. Sooo  approved!
Screenshot from 2019-07-29 11-35-09

jomarko:
@kiegroup/gatekeepers could you please merge?

drools-wb-screens/drools-wb-globals-editor/drools-wb-globals-editor-client/src/main/java/org/drools/workbench/screens/globals/client/editor/AddGlobalPopupViewImpl.html
```diff
@@ -29,7 +29,7 @@
             </div>
         </form>
         <div class="modal-footer">
-            <button id="cancelButton" data-i18n-key="Cancel" class="btn-cancel"></button>
+            <button id="cancelButton" data-i18n-key="Cancel" class="btn btn-cancel"></button>
```
karreiro: I think we need to change this line to something like this.
```suggestion
            <button id="cancelButton" data-i18n-key="Cancel" class="btn btn-default"></button>
```

See the current result:
Screen Shot 2019-07-29 at 09 20 28

And see the result with the suggestion above applied:
Screen Shot 2019-07-29 at 09 20 45

jomarko:
@karreiro thank you, I committed your suggestion.

PR #1203, opened 2019-07-26
DROOLS-4306: [DMN Designer] Search feature - Apply UX and enable search feature

karreiro:
See:
- https://issues.jboss.org/browse/DROOLS-4306
- https://issues.jboss.org/browse/DROOLS-4322
- https://issues.jboss.org/browse/DROOLS-4323

This PR considerably increases the performance of the search in the Guided Decision Tables by making each search element thinner. now the search is 5x faster :-) This work supports the addition of the number of results and the number of the current result.

---

Part of an ensemble:
- https://github.com/kiegroup/kie-wb-common/pull/2785
- https://github.com/kiegroup/drools-wb/pull/1203

---

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/search/SearchableElementFactory.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/search/SearchableElementFactory.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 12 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/editor/search/SearchableElementFactory.java
```diff
@@ -0,0 +1,91 @@
...
+        searchableElement.setHighlightHelper(highlightHelper);
+        searchableElement.setModeller(modeller);
+        searchableElement.setValue(convertDTCellValueToString(cellValue52));
+        searchableElement.setRow(row);
+        searchableElement.setColumn(column);
+
+        return searchableElement;
+    }
+
+    String convertDTCellValueToString(final DTCellValue52 cellValue52) {
```
jomarko: I think we should use `org.drools.workbench.screens.guided.dtable.client.widget.table.utilities.CellUtilities.asString`

karreiro: Good catch. Fixed! :-)

karreiro:
@jomarko Thank you for the review. Changes applied.

PR #1212, opened 2019-08-13
RHDM-1036: Prevent Index Out Of Bounds Exception

jomarko:
For more details, see the https://issues.jboss.org/browse/RHDM-1036

@manstis may I ask you for a review?

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/BRLActionColumnSynchronizer.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/BRLActionColumnSynchronizer.java
@@ -84,13 +84,15 @@ public class BRLActionColumnSynchronizer extends BaseColumnSynchronizer<BaseColu
         final Map<String, List<DTCellValue52>> originalColumnsData = new HashMap<String, List<DTCellValue52>>();
         for (BRLActionVariableColumn variable : originalColumn.getChildColumns()) {
             int iColumnIndex = model.getExpandedColumns().indexOf(variable);
-            final List<DTCellValue52> originalColumnData = new ArrayList<DTCellValue52>();
-            final String key = makeUpdateBRLActionColumnKey(variable);
-            for (List<DTCellValue52> row : model.getData()) {
-                originalColumnData.add(row.get(iColumnIndex));
+            if (iColumnIndex != -1) {
+                final List<DTCellValue52> originalColumnData = new ArrayList<DTCellValue52>();
+                final String key = makeUpdateBRLActionColumnKey(variable);
+                for (List<DTCellValue52> row : model.getData()) {
+                    originalColumnData.add(row.get(iColumnIndex));
+                }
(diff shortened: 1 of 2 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/BRLActionColumnSynchronizer.java
```diff
@@ -84,13 +84,15 @@ public boolean handlesUpdate(final MetaData metaData) throws VetoException {
         final Map<String, List<DTCellValue52>> originalColumnsData = new HashMap<String, List<DTCellValue52>>();
         for (BRLActionVariableColumn variable : originalColumn.getChildColumns()) {
             int iColumnIndex = model.getExpandedColumns().indexOf(variable);
-            final List<DTCellValue52> originalColumnData = new ArrayList<DTCellValue52>();
-            final String key = makeUpdateBRLActionColumnKey(variable);
-            for (List<DTCellValue52> row : model.getData()) {
-                originalColumnData.add(row.get(iColumnIndex));
+            if (iColumnIndex != -1) {
```
jomarko: `iColumnIndex` is `-1` if column is edited second time without saving table in between. This change prevent the `NPE` on the row `originalColumnData.add(row.get(iColumnIndex));`

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/model/synchronizers/impl/BRLConditionColumnSynchronizer.java
```diff
@@ -101,13 +101,15 @@ public boolean handlesUpdate(final MetaData metaData) throws VetoException {
         final Map<String, List<DTCellValue52>> originalColumnsData = new HashMap<>();
         for (BRLConditionVariableColumn variable : originalColumn.getChildColumns()) {
             int iColumnIndex = model.getExpandedColumns().indexOf(variable);
-            final List<DTCellValue52> originalColumnData = new ArrayList<>();
-            final String key = makeUpdateBRLConditionColumnKey(variable);
-            for (List<DTCellValue52> row : model.getData()) {
-                originalColumnData.add(row.get(iColumnIndex));
+            if (iColumnIndex != -1) {
```
jomarko: `iColumnIndex` is `-1` if column is edited second time without saving table in between. This change prevent the `NPE` on the row `originalColumnData.add(row.get(iColumnIndex));`

manstis:
I think this could be fixing a symptom and not the cause.. IIRC when you edit a column the original is cloned and used in the popup editor. When you OK the changes the cloned column should be copied back into the Decision Table model. When you edit the column again the popup editor would again clone the column etc. If the `iColumnIndex` is `-1` then I suspect the edited column is not being written back to the Decision Table model correctly... this is a hunch but w/o stepping through it more I cannot guarantee this is the issue.

jomarko:
The issue is in second iteration of editing, we are searching something ` int iColumnIndex = model.getExpandedColumns().indexOf(variable);` what we deleted in first iteration of editing `delete(originalMetaData);`

PR #1213, opened 2019-08-13
[DROOLS-3570]  Add validation of simulation structure

danielezonca:
Implemented validation of structure of test scenario.

These are check on RULE test scenario:
-> empty column skip
-> instance type removed
-> navigation of bean still valid
-> field type changed

These are check on DMN test scenario
-> empty column skip
-> DMN node removed
-> simple type becomes complex type
-> navigation of data type still valid
-> field type changed

Validation is performed during the loading of the editor or via `Validate` button

https://issues.jboss.org/browse/DROOLS-3570

@kkufova @gitgabrio @Rikkola @yesamer


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioValidationService.java
```diff
@@ -0,0 +1,244 @@
...
+     * @param factMapping
+     * @return
+     */
+    private boolean isToSkip(FactMapping factMapping) {
+        return OTHER.equals(factMapping.getExpressionIdentifier().getType()) ||
+                EMPTY.equals(factMapping.getFactIdentifier()) ||
+                factMapping.getExpressionElements().size() == 0;
+    }
+
+    private DMNType navigateDMNType(DMNType rootType, List<String> steps) {
```
gitgabrio: @danielezonca 
I think it would be better to move this and the following method to a "utility" class - DMNTypeUtils for the moment being has only one method, but if it would be refactored as a common "DMNUtil", it would be the best place to put all this DMN-related helper methods

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
```diff
@@ -470,6 +499,8 @@ public void getModelSuccessCallbackMethod(DataManagementStrategy dataManagementS
         view.setContent(model.getSimulation());
         context.getStatus().setSimulation(model.getSimulation());
         CustomBusyPopup.close();
+        // check if structure is valid
+        getValidateCommand().execute();
```
gitgabrio: @danielezonca 
I think this specific back-and-forth may and should be avoided. We can add those "FactMappingError" inside "ScenarioSimulationModel", mark them to be ignored by Xstream (=not persisted on xml) and showing them at load, without having to issue another service call. Nothing would prevent, anyway, to let the user call the "validate" service at will.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioValidationService.java
```diff
@@ -0,0 +1,244 @@
...
+        for (FactMapping factMapping : simulation.getSimulationDescriptor().getFactMappings()) {
+            if (isToSkip(factMapping)) {
+                continue;
+            }
+
+            String nodeName = factMapping.getFactIdentifier().getName();
+
+            DMNType rootDMNType;
+            try {
+                rootDMNType = dmnModel.getDecisionByName(nodeName) != null ?
```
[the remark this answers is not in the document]
danielezonca: @gitgabrio 
No, the behavior is correct because user can edit the template after the creation and move a decision from EXPECT to GIVEN to test another decision and the opposite

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/editor/ScenarioSimulationEditorPresenter.java
danielezonca: I prefer to postpone this refactor so I have opened this ticket to track it
https://issues.jboss.org/browse/DROOLS-4435

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioValidationService.java
danielezonca: Done

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/ScenarioValidationService.java
```diff
@@ -0,0 +1,244 @@
...
+import static org.drools.scenariosimulation.backend.util.ScenarioBeanUtil.navigateToObject;
+import static org.drools.workbench.screens.scenariosimulation.backend.server.util.DMNTypeUtils.getRootType;
+import static org.drools.workbench.screens.scenariosimulation.model.FactMappingValidationError.createFieldChangedError;
+import static org.drools.workbench.screens.scenariosimulation.model.FactMappingValidationError.createGenericError;
+import static org.drools.workbench.screens.scenariosimulation.model.FactMappingValidationError.createNodeChangedError;
+import static org.kie.dmn.feel.lang.types.BuiltInType.CONTEXT;
+import static org.kie.dmn.feel.lang.types.BuiltInType.UNKNOWN;
+
+@ApplicationScoped
+public class ScenarioValidationService
```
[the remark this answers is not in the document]
danielezonca: Done

kkufova:
@danielezonca, so far, I found the following issues:

1. In rule-based test scenarios, if you remove a data object that is used in the test scenario table, the designer correctly shows the **Wrong column mapping** dialog and Business Central freezes afterwards. Moreover, if this data object is still used inside a DRL file, after clicking **Validate**, an error is thrown instead:

```
Unable to complete your request. The following exception occurred: java.lang.IllegalArgumentException:Retrieving KieContainer has failed. Fix all compilation errors within the project and build the project again..
```

2. In general, my problem with the validation is that it happens quite often that after validating, the **Loading...** message appears and Business Central freezes (and must be reloaded). It makes the testing almost impossible. I got into situation when the loading popup appears every time a test scenario asset is opened and it cannot be edited at all.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNScenarioValidation.java
```diff
@@ -0,0 +1,126 @@
...
+            return false;
+        }
+        String factMappingType = ScenarioSimulationSharedUtils.isList(typeName) ?
+                factMapping.getGenericTypes().get(0) :
+                typeName;
+
+        if (Objects.equals(factMappingType, dmnType.getName())) {
+            return true;
+        }
+        return false;
```
jomarko: I think [1] is same as [2]

[1]
```java
if (Objects.equals(factMappingType, dmnType.getName())) {
            return true;
        }
        return false;
```

[2]
```java
return Objects.equals(factMappingType, dmnType.getName());
```

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/AbstractScenarioValidation.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Skip descriptive columns (FactMappingType.OTHER), column with no instance (FactIdentifier.EMPTY)
+     * and with not expression elements
+     * @param factMapping
+     * @return
+     */
+    protected boolean isToSkip(FactMapping factMapping) {
+        return OTHER.equals(factMapping.getExpressionIdentifier().getType()) ||
+                EMPTY.equals(factMapping.getFactIdentifier()) ||
+                factMapping.getExpressionElements().size() == 0;
```
jomarko: Just for the sake of short codebase `.size() == 0` could be replaced by `.isEmpty()`

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNScenarioValidation.java
```diff
@@ -0,0 +1,126 @@
...
+                errors.add(createNodeChangedError(factMapping, "node not found"));
+                continue;
+            }
+
+            List<String> steps = expressionElementToString(factMapping);
+
+            // error if direct mapping (= simple type) but it is a composite
+            // NOTE: context is a special case so it is composite even if no fields are declared
+            Type rootType = getRootType((BaseDMNTypeImpl) rootDMNType);
+            if (!CONTEXT.equals(rootType) && steps.size() == 0 && rootDMNType.isComposite()) {
```
jomarko: We could use `steps.isEmpty()` as we do in `RULEScenarioValidation`

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNScenarioValidation.java
danielezonca: Yes, you are right, done :)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/AbstractScenarioValidation.java
danielezonca: Done

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/main/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNScenarioValidation.java
danielezonca: Done

danielezonca:
@jomarko 
Can you please test it?
I have fixed both issues raised by Klara:
1. A notification error appear if validation fail because the compilation of the project fails (please note that DMN test scenario has another error popup raised by right panel too in case of compilation failure and this has not been changed)
Screenshot from 2019-08-19 18-48-32
2. Removed loading indicator after validation

PR #1214, opened 2019-08-14
DROOLS-4407: [DMN Designer] Search feature - Highlight the current result with colors that follow the PF standard

danielzhe:
This PR is part of an ensemble and also covers [DROOLS-4406](https://issues.jboss.org/projects/DROOLS/issues/DROOLS-4406).

https://github.com/kiegroup/drools-wb/pull/1214
https://github.com/kiegroup/appformer/pull/767
https://github.com/kiegroup/kie-wb-common/pull/2815

### DIFF
--- a/drools-wb-webapp/src/main/webapp/WEB-INF/beans.xml
+++ b/drools-wb-webapp/src/main/webapp/WEB-INF/beans.xml
@@ -4,6 +4,9 @@
     <!-- These exclusions were added by Errai to avoid deploying client-side classes to the server -->
     <!-- End of Errai exclusions -->
 
+    <!-- These exclusions were added by Errai to avoid deploying client-side classes to the server -->
+    <!-- End of Errai exclusions -->
+
     <exclude name="org.drools.workbench.client.**"/>
     <exclude name="org.drools.workbench.screens.guided.dtable.client.**"/>
     <exclude name="org.drools.workbench.screens.guided.dtree.client.**"/>
--- a/drools-wb-webapp/src/main/webapp/WEB-INF/beans.xml
+++ b/drools-wb-webapp/src/main/webapp/WEB-INF/beans.xml
@@ -7,6 +7,9 @@
     <!-- These exclusions were added by Errai to avoid deploying client-side classes to the server -->
     <!-- End of Errai exclusions -->
 
+    <!-- These exclusions were added by Errai to avoid deploying client-side classes to the server -->
+    <!-- End of Errai exclusions -->
+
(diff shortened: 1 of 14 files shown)


danielzhe:
@karreiro @jomarko Can you check it, please?

jomarko:
@danielzhe Hi, I will provide final feedback here. I haven't commented code of first two parts of ensemble. However this last PR gave me some ideas about whole ensemble.

- One issue is that the highlight doesn't work in the DMN grids. 
- The second point in my mind is, we introduce and observe an event (`SearchPerformedEvent`) that is not needed from my point of view. I think we just need to update `GridHighlightHelper#select` method to apply also the background color. I think this because @karreiro introduced concept of searchable elements. Each searchable element has callback `onFound` that is invoked in same situation, when we observe `SearchPerformedEvent`. All our searchable elements use the `GridHighlightHelper#select`, that is why I think we could try to update this method. What do you think?

danielzhe:
@jomarko I would agree with you if we have more time, because to achieve this we would have to make changes in Lienzo level. About the DMN Grids, I'm checking.

danielzhe:
@jomarko Changes applied! Is working on DMN Grids too. @karreiro

jomarko:
@danielzhe thank you for updates, could you please rebase the ensemble? There are some conflicts now.

danielzhe:
@karreiro @jomarko Ready for your re-review.

drools-wb-webapp/src/main/webapp/WEB-INF/beans.xml
```diff
@@ -4,6 +4,12 @@
     <!-- These exclusions were added by Errai to avoid deploying client-side classes to the server -->
     <!-- End of Errai exclusions -->
 
+    <!-- These exclusions were added by Errai to avoid deploying client-side classes to the server -->
+    <!-- End of Errai exclusions -->
+
+    <!-- These exclusions were added by Errai to avoid deploying client-side classes to the server -->
+    <!-- End of Errai exclusions -->
+
```
jomarko: Please do not include these changes in PR.

jomarko: A reminder, we shouldn't these changes of `beans.xml` in this PR.

jomarko: @danielzhe please, ensure no changes are done in beans.xml as part of drools-wb#1214 PR

PR #1217, opened 2019-08-19
Drools 4434 Setup drools-wb for Sonar analysis

gitgabrio:
@danielezonca @rsynek 

See https://issues.jboss.org/browse/DROOLS-4434

For the moment being most of submodules have sonar analysis disabled, and the report on sonar mark this project as "Red" quality gate


pom.xml
```diff
@@ -281,4 +286,169 @@
...
+            <artifactId>jacoco-maven-plugin</artifactId>
+            <configuration>
+              <skip>${code.coverage.disabled}</skip>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <!-- Merges test coverage reports and uploads statistics via sonar scanner into the SonarCloud. -->
+    <profile>
```
rsynek: This profile is already defined in kie-parent

pom.xml
```diff
@@ -281,4 +286,169 @@
...
+                  <goal>sonar</goal>
+                </goals>
+                <phase>generate-resources</phase>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
```
rsynek: ditto

pom.xml
```diff
@@ -39,6 +39,11 @@
     <checkstyle.logViolationsToConsole>true</checkstyle.logViolationsToConsole>
     <jacoco.haltOnFailure>true</jacoco.haltOnFailure>
     <spotbugs.failOnViolation>true</spotbugs.failOnViolation>
+    <!-- JaCoCo coverage data file location. Using a single file for appending in the project's root directory makes it
+       possible to measure cross-module test coverage -->
+    <!--suppress UnresolvedMavenProperty -->
+    <jacoco.exec.file>${maven.multiModuleProjectDirectory}/target/jacoco.exec</jacoco.exec.file>
```
rsynek: The property is already defined here:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L542

pom.xml
gitgabrio: @rsynek 
Fixed!

pom.xml
gitgabrio: @rsynek 
Fixed!

gitgabrio:
@rsynek 
Hi. Could you verify the PR now? I've removed all the inherited stuff

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/pom.xml
```diff
@@ -65,6 +65,13 @@
       <artifactId>drools-scenario-simulation-api</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.jacoco</groupId>
```
rsynek: what is the reason for declaring this dependency here?

pom.xml
```diff
@@ -37,7 +37,8 @@
     </checkstyle.header.template>
     <checkstyle.failOnViolation>true</checkstyle.failOnViolation>
     <checkstyle.logViolationsToConsole>true</checkstyle.logViolationsToConsole>
-    <jacoco.haltOnFailure>true</jacoco.haltOnFailure>
+    <jacoco.haltOnFailure>false</jacoco.haltOnFailure>
+    <jacoco.line.coveredratio.minimum>0.6</jacoco.line.coveredratio.minimum>
```
rsynek: No need to override the threshold, no one cares as long as it doesn't fail the build, which is prevented by the property one line above.

gitgabrio: @rsynek 
Copied them from kie-parent - I know it is "ignored" due to the "haltOnFailure"

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/pom.xml
gitgabrio: @rsynek 
Without that I had classnotfound exception.

gitgabrio:
@rsynek @danielezonca 
I've setup code coverage (both jacoco execution and sonar analysis) only for scenario-simulation, and the reuslt is a 66% coverage. I've created another sonar-profile to allow such level of coverage, but anyway the bugs rate is "E" (like for other projects).

pom.xml
```diff
@@ -44,6 +44,7 @@
        possible to measure cross-module test coverage -->
     <!--suppress UnresolvedMavenProperty -->
     <code.coverage.disabled>false</code.coverage.disabled>
+    <sonar.coverage.exclusions></sonar.coverage.exclusions>
```
rsynek: If I understand it correctly, this line made the coverage working, right?

I wonder why it behaves like that, as e.g. for drools, there is no such declaration needed in the top-level pom.xml [1]. The situation is similar there, one of the sub-modules excludes every java file from coverage statistics [2].

[1] https://github.com/kiegroup/drools/blob/master/pom.xml
[2] https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-api/pom.xml#L68

rsynek:
> @rsynek @danielezonca
> I've setup code coverage (jacoco execution) and whole sonar analysis only for scenario-simulation, and the result is a 66% coverage with bug level of "C". I've created another sonar-profile to allow such level of coverage, but anyway the security rate is "B", worst then the accepted one.

First of all, 66% overall coverage is nice. There is no point in lowering the threshold via a Quality Gate just to see it passed. The gate is more useful when it comes to new PRs - and there you might want to keep the 80%. After some time (and amount of code), if every PR is above 80% test coverage, the entire project will get over 80% too.

Regarding the bug and security ratings. I suggest taking some time to look into these issues and either fix them or close them.

Thank you, nice job!

pom.xml
```diff
@@ -356,8 +356,24 @@
                 <systemPropertyVariables>
                   <jacoco-agent.destfile>${jacoco.exec.file}</jacoco-agent.destfile>
                 </systemPropertyVariables>
+                <skip>${code.coverage.disabled}</skip>
               </configuration>
             </plugin>
+            <plugin>
+              <artifactId>maven-compiler-plugin</artifactId>
```
rsynek: Would you please explain why skipping test compilation is necessary?

pom.xml
rsynek: As it's not used, I would recommend removing the property from here.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/pom.xml
rsynek: Have you found out why is the dependency required here? I should be already on classpath due to [1] and I can't see a reason why it needs to be declared here, especially outside any profile.

[1] https://github.com/kiegroup/drools-wb/pull/1217/files#diff-600376dffeb79835ede4a0b285078036R310

pom.xml
gitgabrio: @rsynek 
That profile is used only for sonar reports. So, all the time spent on "testCompile" modules that are ignored by jacoco and sonar is wasted.

gitgabrio:
> > @rsynek @danielezonca
> > I've setup code coverage (jacoco execution) and whole sonar analysis only for scenario-simulation, and the result is a 66% coverage with bug level of "C". I've created another sonar-profile to allow such level of coverage, but anyway the security rate is "B", worst then the accepted one.
> 
> First of all, 66% overall coverage is nice. There is no point in lowering the threshold via a Quality Gate just to see it passed. The gate is more useful when it comes to new PRs - and there you might want to keep the 80%. After some time (and amount of code), if every PR is above 80% test coverage, the entire project will get over 80% too.
> 
> Regarding the bug and security ratings. I suggest taking some time to look into these issues and either fix them or close them.
> 
> Thank you, nice job!

@rsynek 
This is not the scope of this PR, as we agreed upon since the very beginning.

pom.xml
rsynek: I think that's a false assumption - the sonarcloud analysis is bound to `generate-resources` phase, which is called from jenkins jobs:
https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/pr_jobs.groovy#L254

It's not ideal; I should have made it even more early - `validate`. So if you want to optimize the execution time even further, I would recommend binding the maven execution on Jenkins to the `validate` phase and of course making a corresponding change in kie-parent.

The described optimization is out of scope of this PR, though.

rsynek:
> > > @rsynek @danielezonca
> > > I've setup code coverage (jacoco execution) and whole sonar analysis only for scenario-simulation, and the result is a 66% coverage with bug level of "C". I've created another sonar-profile to allow such level of coverage, but anyway the security rate is "B", worst then the accepted one.
> > 
> > 
> > First of all, 66% overall coverage is nice. There is no point in lowering the threshold via a Quality Gate just to see it passed. The gate is more useful when it comes to new PRs - and there you might want to keep the 80%. After some time (and amount of code), if every PR is above 80% test coverage, the entire project will get over 80% too.
> > Regarding the bug and security ratings. I suggest taking some time to look into these issues and either fix them or close them.
> > Thank you, nice job!
> 
> @rsynek
> This is not the scope of this PR, as we agreed upon since the very beginning.

Absolutely, that's something which should be addressed after merging this PR - sorry if my comment made you think it was a required change. This is a follow-up work, I would just avoid lowering the thresholds as the first step.

pom.xml
gitgabrio: @rsynek 
I've see that while launching commands, that's why I've put that that way.
Please verify yourself.

rsynek: It depends on what command did you use to run the analysis. If I run it via:
`mvn generate-resources -Psonarcloud-analysis`, there is no invocation of the compiler plugin.

gitgabrio: @rsynek 
That exclusion is inside the run-code-coverage profile, not in the sonarcloud one.

rsynek: I see. So you want to skip compilation of tests in the original maven goal that is supposed to run tests? This leads me to a question - are these tests from modules where you skip their compilation executed?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-api/pom.xml
gitgabrio: @rsynek 
Removed

pom.xml
gitgabrio: @rsynek 

Needed to analyze/upload code-coverage. Moved to scenario-simulation

gitgabrio:
@danielezonca  @rsynek 
This is the last status.

rsynek:
@gitgabrio nice! So the last comment that remains to be resolved is the one about a compiler plugin and then we can merge.

PR #1220, opened 2019-08-21
DROOLS-4452 : [GDST] Wrong FQCN is selected if the end matches

Rikkola:
https://issues.jboss.org/browse/DROOLS-4452

@jomarko @karreiro if you two could please take a look

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/GuidedDecisionTableModelVisitor.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/GuidedDecisionTableModelVisitor.java
@@ -89,7 +89,7 @@ public class GuidedDecisionTableModelVisitor {
         }
         String fullyQualifiedClassName = null;
         for (Import imp : imports.getImports()) {
-            if (imp.getType().endsWith(factType)) {
+            if (imp.getType().endsWith("." + factType)) {
                 fullyQualifiedClassName = imp.getType();
                 break;
             }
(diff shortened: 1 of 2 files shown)


drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/GuidedDecisionTableModelVisitor.java
```diff
@@ -89,7 +89,7 @@ private String convertToFullyQualifiedClassName(final String factType) {
         }
         String fullyQualifiedClassName = null;
         for (Import imp : imports.getImports()) {
-            if (imp.getType().endsWith(factType)) {
+            if (imp.getType().endsWith("." + factType)) {
```
jomarko: Hmm I think this is still not proper fix, imagine a testcase:
```java
    @Test
    public void differentPackagesSameNames() {
        final GuidedDecisionTable52 model = new GuidedDecisionTable52();
        model.getImports().addImport(new Import("org.test.demo.Person"));
        model.getImports().addImport(new Import("org.test.Person"));
        final Pattern52 pattern52 = mock(Pattern52.class);
        doReturn("Person").when(pattern52).getFactType();

        model.getConditions().add(pattern52);

        final Set<String> consumedModelClasses = new GuidedDecisionTableModelVisitor(model).getConsumedModelClasses();

        assertTrue(consumedModelClasses.contains("org.test.Person"));
    }
```

Rikkola:
@karreiro yeah that is an issue. You actually can't use two Person fact types from different packages, but to make that work we would have to use the FQCN in the generated source and upate the fact selectors with the package data. A lot of work, but doable.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/GuidedDecisionTableModelVisitor.java
Rikkola: At the moment the guided rule editors do not support the same class name from two sources. This is because the generated DRL imports every import and then in the actual rule part you only have Person. The compiler does not know what to pick.

Rikkola:
@kiegroup/gatekeepers Can I please get a merge for this PR. Thank you.

PR #1226, opened 2019-09-04
[DROOLS-3969] Highlight selected cell when invoking contextual menu.

gitgabrio:
@jomarko  @yesamer  @danielezonca

https://issues.jboss.org/browse/DROOLS-3969

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/menu/ScenarioContextMenuRegistry.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/menu/ScenarioContextMenuRegistry.java
@@ -51,6 +51,9 @@ public class ScenarioContextMenuRegistry {
     protected GridContextMenu gridContextMenu;
     protected UnmodifiableColumnGridContextMenu unmodifiableColumnGridContextMenu;
     protected ErrorReportPopoverPresenter errorReportPopoverPresenter;
+    
+    private static final String GIVEN = "GIVEN";
+    private static final String EXPECT = "EXPECT";
 
     @Inject
     public ScenarioContextMenuRegistry(final OtherContextMenu otherContextMenu,
@@ -127,7 +130,7 @@ public class ScenarioContextMenuRegistry {
                                     final Integer uiRowIndex,
                                     final Integer uiColumnIndex,
                                     final boolean isHeader) {
-
+        scenarioGrid.clearSelections();
         ScenarioGridColumn scenarioGridColumn = (ScenarioGridColumn) scenarioGrid.getModel().getColumns().get(uiColumnIndex);
         if (scenarioGridColumn == null) {
             return false;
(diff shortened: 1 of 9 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/menu/ScenarioContextMenuRegistry.java
```diff
@@ -184,20 +187,20 @@ private boolean manageScenarioBodyRightClick(final ScenarioGrid scenarioGrid,
                                                  final int top,
                                                  final int uiRowIndex,
                                                  final int uiColumnIndex) {
-        scenarioGrid.deselect();
```
yesamer: @gitgabrio can you please explain why you removed this? I see you added `scenarioGrid.clearSelections();` to clear the selection @ line 133, but `scenarioGrid.deselect();` modifies a different status.

gitgabrio: @yesamer 
Good question. The behavior we have to implement is that when you right click on a cell **only that cell** has to be selected. But that "deselect" method (for what I get from documentation) change the status of the whole widget (the grid, in our case). I fail to find a reason to do that, and I'm also puzzled on the actual usage of it, but it seems wrong to me: why should be deselected the grid that has just been clicked on?
I fear it could be related somehow with the keyboard management stuff, but with the test I've done everything seems to work, so unless an actual reason comes up, I would prefer to remove that.

manstis: The `GridWidget` (used by _scesim_, DMN and Gudied Decision Tables) supports selection of the whole grid. The Guided Decision Table Editor can have multiple grids visible at once and a _selected_ grid may have no _selected_ cells but has a highlight around the whole grid to indicate it is _selected_ (has focus). The `gridWidget.deselect()` is the opposite of `gridWidget.select()` that _selects_ the whole grid. The `gridWidget.clearSelections()` method clears the selected cells - but perhaps calling it `clearSelectedCells()` would be nicer... feel free to submit PRs accordingly :-)

yesamer: @gitgabrio @manstis Thank you for your explanations. I agree to remove that deselect method call and to rename the gridWidget.clearSelections() as Michael suggested.

gitgabrio: @manstis 
Thanks, I thought it was so. My phrase "and I'm also puzzled on the actual usage of it" was related on how/when that status is used in that specific process and -overall - inside scenario editor.
About renaming, agree but not in this PR and not in an "anarchic" way. Working on the grid   for some time we have found some things that could be improved, but I think it is better to discuss them with you beforehand to clearly understand the original design and follow it, without messing the code with wrong interpretation/assumptions; and I do not think it is a great idea to make a PR just to change one/some method names.

danielezonca:
jenkins retest this please

gitgabrio:
jenkins please retest this

PR #1227, opened 2019-09-06
[DROOLS-4083] After asset renaming, column width is automatically changed

yesamer:
To resolve the issue, I changed the `SaveAndRenameCommandBuilder` to launch a command BEFORE the saveAndRename call is invoked. In this way, it's possibile to add a command witch synchronizes the columns widths from front-end model to back-end model.

@jomarko @gitgabrio @danielezonca

Related PR: https://github.com/kiegroup/appformer/pull/792


yesamer:
jenkins retest this please

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-businesscentral-client/src/main/java/org/drools/workbench/screens/scenariosimulation/businesscentral/client/editor/ScenarioSimulationEditorBusinessCentralWrapper.java
```diff
@@ -266,6 +270,27 @@ protected void save(final String commitMessage) {
                                                                                                                  commitMessage);
     }
 
+    @Override
+    protected Command getSaveAndRename() {
```
jomarko: I think we should not override `getSaveAndRename` in this way. It differs form `BaseEditor.getSaveAndRename` just in one line and it duplicates 10 lines. I think we should move the current `ScenarioSimulationEditorBusinessCentralWrapper.getSaveAndRename` into `BaseEditor`. Then In `BaseEditor` implement `beforeSaveAndRenameCommand` as no operation and here in `ScenarioSimulationEditorBusinessCentralWrapper` override just one method `getBeforeSaveAndRenameCommand`. Please let me know what do you think.

jomarko:
@yesamer manual review note, I can confirm the current code if you click **Save and Rename**. However the column width is discarded if you click just **Rename**, is it expected?

yesamer:
> ### Manual review note
> 
> @yesamer I can confirm the current code works if you click **Save and Rename**. However the column width is discarded if you click just **Rename**, is it expected please?
@jomarko thank you for your manual review. I confirm, that is the expected behaviour. **Any** change applied between last save and "Rename" action is lost.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-businesscentral-client/src/main/java/org/drools/workbench/screens/scenariosimulation/businesscentral/client/editor/ScenarioSimulationEditorBusinessCentralWrapper.java
yesamer: @jomarko I think your proposed change fits better with current code in `BaseEditor` class, then I applied it as you suggested. Thx!

jomarko:
@yesamer I approved already, hehe

yesamer:
> @yesamer I approved already, hehe

my bad :)

yesamer:
@kiegroup/gatekeepers Hi, can you please merge it in master? Thanks!

PR #1231, opened 2019-09-16
DROOLS-4536 : Test scenarios fails with default values

Rikkola:
https://issues.jboss.org/browse/DROOLS-4536

@jomarko @manstis if you two could please review.
This presets the value that the textbox has by default for the model. I changed the style to MVP so it is testable.

### DIFF
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/EditableTextBox.java
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/EditableTextBox.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.workbench.screens.testscenario.client;
+
+import com.google.gwt.user.client.ui.IsWidget;
+import com.google.gwt.user.client.ui.Widget;
+import org.gwtbootstrap3.client.ui.TextBox;
+import org.uberfire.client.callbacks.Callback;
+
+public class EditableTextBox
+        implements IsWidget {
+
+    private TextBox view;
+    private Callback<String> changed;
+
+    public EditableTextBox(final Callback<String> changed,
+                           final TextBox view,
+                           final String fieldName,
+                           final String initialValue) {
+        this.changed = changed;
+        this.view = view;
+        view.addValueChangeHandler(event -> this.onValueChange(event.getValue()));
+
+        view.setText(initialValue);
+        view.setTitle(fieldName);
+    }
+
+    @Override
+    public Widget asWidget() {
+        return view.asWidget();
+    }
+
+    public void onValueChange(final String value) {
+        changed.callback(value);
+    }
+}
--- a/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/test/java/org/drools/workbench/screens/testscenario/client/EditableTextBoxTest.java
+++ b/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/test/java/org/drools/workbench/screens/testscenario/client/EditableTextBoxTest.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.workbench.screens.testscenario.client;
+
+import com.google.gwt.event.logical.shared.ValueChangeEvent;
+import com.google.gwt.event.logical.shared.ValueChangeHandler;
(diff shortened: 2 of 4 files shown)


drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/EditableTextBox.java
```diff
@@ -0,0 +1,52 @@
...
+        this.changed = changed;
+        this.view = view;
+        view.setPresenter(this);
+
+        view.setText(initialValue);
+        view.setTitle(fieldName);
+
+        if (initialValue == null || initialValue.isEmpty()) {
+            changed.valueChanged(view.getText());
+        }
```
Rikkola: The actual fix is these 3 lines above this comment

manstis:
@Rikkola in light of @jomarko comment please let me know when this PR is ready for review.

I assume you have to investigate and possible make additional changes.

Rikkola:
@jomarko This works for me, but not sure if this bug is related:
https://issues.jboss.org/browse/DROOLS-4540

Rikkola:
@jomarko If that bug was not the casue for it not working for you. I could use the test scenario you used.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/EditableTextBox.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.gwt.user.client.ui.Widget;
+import org.uberfire.client.callbacks.Callback;
+
+public class EditableTextBox
+        implements IsWidget {
+
+    private final EditableTextBoxView view;
+    private Callback<String> changed;
+
+    public EditableTextBox(final Callback<String> changed,
```
Rikkola: @jomarko FYI I changed the deprecated listener to callback.

Rikkola:
@manstis All good now. Ready for review.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/EditableTextBox.java
jomarko: @Rikkola Ok, I expect this change is safe.

Rikkola:
So the `TextBox` you get from `TextBoxFactory.getTextBox(..)` has a preset value based on the text box type. When you create a new field check widget in the test scenario the `initialValue` has no value. So the UI and model are out of sync. The row that was linked syncs the model and UI when this happens. 

The complexity comes from me trying to move parts of this to MVP for testing purposes, while still leaving as much as it is so I don't break anything else.

Not sure if this explains it or causes more confusion ;)

manstis:
@Rikkola Thank-you for the explanation.

I ran `master` code and debugged a little.

It appears that if you move the `ValueChangeHandler` registration in `VerifyFieldConstraintEditor` to before the value is set the model is correctly updated with the default/preset value. Moving [this](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/VerifyFieldConstraintEditor.java#L220) to the line after [this](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/VerifyFieldConstraintEditor.java#L216) also works successfully.

Please advise whether you prefer to handle the existing support for default values (in the presence of providing an invalid value) or whether you prefer to work around it, as this PR does... IMO moving the registration of the `ValueChangeHandler` is the _correct_ fix; but I'm an awkward fellow ;-)

Rikkola:
Riiiight. That works. So setting the `TextBox` value to empty does not reset the `TextBox` value, but updates the model to what the `TextBox`default. Not sure why this was not the first thing I tried :)

This should be future proof too ( if this part of the code even has a future ) so that if the `TextBox` value someday can be set to empty, then it just updates to empty. I'll update the fix.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/test/java/org/drools/workbench/screens/testscenario/client/EditableTextBoxTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        new EditableTextBox(valueChanged,
+                            view,
+                            "age",
+                            "11");
+
+        final InOrder o = inOrder(view, valueChanged);
+
+        o.verify(valueChanged).callback(anyString());
+        o.verify(view).setText("11");
```
Rikkola: So now the actual fix is that these two happen in this order.

Rikkola:
There is now quite a lot of complexity to just swap the order of two rows. Do we keep this for testing purposes or move out of this? I would prefer what the PR now contains, but open to suggestions.

drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/main/java/org/drools/workbench/screens/testscenario/client/EditableTextBox.java
```diff
@@ -0,0 +1,50 @@
...
+                           final EditableTextBoxView view,
+                           final String fieldName,
+                           final String initialValue) {
+        this.changed = changed;
+        this.view = view;
+        view.setPresenter(this);
+
+        changed.callback(view.getText());
```
manstis: This should not be needed.

1. You add a `ValueChangeHandler` [here](https://github.com/kiegroup/drools-wb/pull/1231/files#diff-694563eaf1d207c6a6706ca0e0f9ab77R39).
2. When the value is set [here](https://github.com/kiegroup/drools-wb/pull/1231/files#diff-694563eaf1d207c6a6706ca0e0f9ab77R49), if it is invalid, the `ValueChangeHandler` attached above should be called.

So, provided, you call `view.setPresenter(this)` **before** `view.setText(initialValue)` (as you already do) then the existing mechanism to update the model with the default/valid value should execute.

Rikkola: Right I actually meant to do this, but now this gives me another puzzle on how to test this in a way that makes sense. I'll figure something out.

Rikkola:
@manstis Updated, let me know what you think.

@jomarko Not sure if you want to retry this or not.

Rikkola:
@manstis always happy to see people review my changes with some effort. The current implementation is a lot better than what I proposed.

Rikkola:
Jenkins please retest this.

Rikkola:
@kiegroup/gatekeepers Can I get a merge please.

PR #1240, opened 2019-10-11
Drools 4638 - Disable Save button on settings for user without write permission

gitgabrio:
@dupliaka  @yesamer @danielezonca

see https://issues.jboss.org/browse/DROOLS-4638

Logic behind the current implementation is:

1) currently, user "credential" as a concept exists only inside BC, not in Kogito
2) as such, I consider the "read/write" permission a BC-specific condition
3) so, main presenter and settings presenter should normally work without such knowledge, and their default editable status is always _true_
4) it is the BC specific component (**ScenarioSimulationEditorBusinessCentralWrapper**) - in case - to "disable" the editable status
5) I had to add the **ScenarioSimulationEditorPresenter.saveEnabled** flag because both the _SettingsPanel_ setup and the retrieval of user "read/write" permissions happen in different - more or less asynchronous - threads, so it is not sure which of the two happen before

- if credential retrieval complete before, only **ScenarioSimulationEditorPresenter.saveEnabled** is updated (since there is still no _SettingsPanel_ active) and, when _SettingsPanel_ setup is invoked, that status will be taken in consideraton
- if _SettingsPanel_ setup complete before, **ScenarioSimulationEditorPresenter** will update its **saveEnabled** flag and will update the _SettingsPanel_ presenter

6) I had to create **SettingsPresenter.saveEnabled** because - during the lifecycle of the editor - **SettingsPresenter.reset()** is invoked (that restore the settings view to the default status) and then this flag is used to know if the button has to be shown/removed

7) I choose to add/remove the "Save" button to be more consistent with the behavior of the main toolbar.

8) I had to add some indirection inside **ScenarioSimulationEditorBusinessCentralWrapper** for testing purposes.

/------/
A concern that crossed may mind working on this PR is: prevent unwanted user actions (e.g."Save" for a read-only user) only by hiding buttons is not extremely secure inside a web page.
Anyway it is a topic that - eventually - should be discussed in other place and involve the whole architecture, I think.

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
@@ -50,6 +50,8 @@ public class SettingsPresenter extends AbstractSubDockPresenter<SettingsView> im
 
     protected SettingsScenarioSimulationDropdown settingsScenarioSimulationDropdown;
 
+    private boolean saveEnabled = true;
+
     public SettingsPresenter() {
         //Zero argument constructor for CDI
         title = ScenarioSimulationEditorConstants.INSTANCE.settings();
@@ -93,22 +95,34 @@ public class SettingsPresenter extends AbstractSubDockPresenter<SettingsView> im
     }
 
     @Override
-    public void onSaveButton(String scenarioType) {
-        simulationDescriptor.setSkipFromBuild(view.getSkipFromBuild().isChecked());
-        switch (ScenarioSimulationModel.Type.valueOf(scenarioType)) {
-            case RULE:
-                saveRuleSettings();
-                break;
-            case DMN:
-                saveDMNSettings();
-                break;
+    public void setSaveEnabled(boolean toSet) {
+        saveEnabled = toSet;
+        view.getSaveButton().setDisabled(!saveEnabled);
+        if (!saveEnabled) {
+            saveCommand = null;
         }
-        saveCommand.execute();
+    }
+
+    @Override
+    public void onSaveButton(String scenarioType) {
+        if (saveCommand != null) {
+            simulationDescriptor.setSkipFromBuild(view.getSkipFromBuild().isChecked());
+            switch (ScenarioSimulationModel.Type.valueOf(scenarioType)) {
+                case RULE:
+                    saveRuleSettings();
+                    break;
+                case DMN:
+                    saveDMNSettings();
+                    break;
+            }
+            saveCommand.execute();
+        }
     }
 
     @Override
     public void reset() {
         view.reset();
+        view.getSaveButton().setDisabled(!saveEnabled);
         settingsScenarioSimulationDropdown.clear();
     }
 
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
@@ -50,7 +50,7 @@ public class SettingsPresenter extends AbstractSubDockPresenter<SettingsView> im
 
     protected SettingsScenarioSimulationDropdown settingsScenarioSimulationDropdown;
 
-    private boolean saveEnabled = true;
+    protected boolean saveEnabled = true;
 
     public SettingsPresenter() {
         //Zero argument constructor for CDI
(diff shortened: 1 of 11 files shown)


manstis:
Being a nosy blighter.. if your editor extends `KieEditor` (for Business Central) the save menu item should already be disabled if the User lacks permissions. See https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-widgets/kie-wb-metadata-widget/src/main/java/org/kie/workbench/common/widgets/metadata/client/KieEditor.java#L310...

gitgabrio:
> Being a nosy blighter.. if your editor extends `KieEditor` (for Business Central) the save menu item should already be disabled if the User lacks permissions. See https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-widgets/kie-wb-metadata-widget/src/main/java/org/kie/workbench/common/widgets/metadata/client/KieEditor.java#L310...

@manstis 
You are, but I'm getting used to it.... :) 
Anyway, the issue here is with the "Save" button on the "Settings" panel (you know, we love to add buttons here and there :) )

gitgabrio:
> @gitgabrio Fine for me. Considering 2 asynchronous flows are present, the only suggestion I have is to add a test in `ScenarioPresenterTest` which assures that the calls of `setSaveEnabled()` and `reset()` methods lead to an expected behaviour in _any_ order they are called. I agree it's not an unit test, it's a little integration tests, but in my opinion it could be very useful, even in case someone try to change the code. (a similar test could be introduced for `ScenarioSimulationEditorPresenter.java` methods `setSaveEnabled(boolean toSet)` and `setSettings(SettingsView.Presenter presenter)` )

@yesamer 
I'll think to that, thanks!

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
```diff
@@ -92,23 +94,38 @@ public void setSaveCommand(Command saveCommand) {
...
-                break;
+        if (saveCommand != null) {
+            simulationDescriptor.setSkipFromBuild(view.getSkipFromBuild().isChecked());
+            switch (ScenarioSimulationModel.Type.valueOf(scenarioType)) {
+                case RULE:
+                    saveRuleSettings();
+                    break;
+                case DMN:
+                    saveDMNSettings();
+                    break;
```
dupliaka: Just  a small suggestion to add a default with new IlligalStateException in case if we would add some new scenario type.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
```diff
@@ -92,23 +94,38 @@ public void setSaveCommand(Command saveCommand) {
...
+            }
+            saveCommand.execute();
         }
-        saveCommand.execute();
     }
 
     @Override
     public void reset() {
         view.reset();
+        if (!saveEnabled) {
```
dupliaka: A small suggestion: if you use !saveEnabled every time, probably it could worth to use saveDisabled.

dupliaka:

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
gitgabrio: Hi @dupliaka 
Yes, I know I could do the other way around. But as a matter of "comprehensibility" - i.e. to explicit the meaning of this piece of code - that way it is clear that _!saveEnabled_ is a sort of _specific_ situation that "normally" should not happen (unless in very specific cases).

dupliaka: Oh, I didn't realize it this way.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
gitgabrio: Hi @dupliaka 
in other contexts I would do exactly as per your suggestion.
Here, throwing an exception could lead to an unexpected (or difficult to manage) ui behavior, so for the moment being I'm adding an empty default.
What should be actually tested (but it is not doable in this current PR) is that this method is _never_ invoked with a not-existing type of scenario.

gitgabrio:
jenkins please retest this

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/SettingsPresenter.java
dupliaka: Thanks for an explanation, I will keep it in mind

gitgabrio:
jenkins please retest this

PR #1245, opened 2019-10-21
DROOLS-4668: Show placeholder for dmn date cells

jomarko:
For more details, see https://issues.jboss.org/browse/DROOLS-4668

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ConstantHolder.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ConstantHolder.java
@@ -31,4 +31,6 @@ public class ConstantHolder {
     // GWT doesn't support Java 8 LocalDate
     public static final String LOCALDATE_SIMPLE_NAME = "LocalDate";
     public static final String LOCALDATE_CANONICAL_NAME = "java.time.LocalDate";
+
+    public static final String DMN_DATE = "date";
 }
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
@@ -17,6 +17,7 @@ package org.drools.workbench.screens.scenariosimulation.client.utils;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import com.ait.lienzo.client.core.types.Point2D;
@@ -273,15 +274,17 @@ public class ScenarioSimulationUtils {
         return new Point2D(cellXMiddle, cellYMiddle);
     }
 
-    public static String getPlaceholder(String canonicalClassName) {
-        switch (canonicalClassName) {
-            case LOCALDATE_CANONICAL_NAME:
-                return ScenarioSimulationEditorConstants.INSTANCE.dateFormatPlaceholder();
-            case DMN_DATE:
-                return ScenarioSimulationEditorConstants.INSTANCE.dmnDateFormatPlaceholder();
-            default:
-                return ScenarioSimulationEditorConstants.INSTANCE.insertValue();
+    public static String getPlaceholder(final String canonicalClassName) {
+
+        if (Objects.equals(LOCALDATE_CANONICAL_NAME, canonicalClassName)) {
+            return ScenarioSimulationEditorConstants.INSTANCE.dateFormatPlaceholder();
         }
+
+        if (Objects.equals(DMN_DATE, canonicalClassName)) {
+            return ScenarioSimulationEditorConstants.INSTANCE.dmnDateFormatPlaceholder();
+        }
+
+        return ScenarioSimulationEditorConstants.INSTANCE.insertValue();
     }
 
     protected static double getColumnWidth(String columnId) {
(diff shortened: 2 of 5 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
```diff
@@ -273,15 +275,23 @@ public static Point2D getMiddleXYCell(final GridWidget gridWidget, final GridCol
     }
 
     public static String getPlaceholder(String canonicalClassName) {
-        return isLocalDate(canonicalClassName) ?
-                ScenarioSimulationEditorConstants.INSTANCE.dateFormatPlaceholder() :
-                ScenarioSimulationEditorConstants.INSTANCE.insertValue();
+        if (isLocalDate(canonicalClassName)) {
```
gitgabrio: @jomarko 
Hi - Maybe this could be replaced with switch - it was not so because there where just two possibilities, but if adding more is needed, the switch is more easier to read and extend in future (the called methods just compares Strings, and since for each comparison one of the compared term is a final static, I think they can be used in the case statement)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
```diff
@@ -158,6 +158,7 @@ apply=Apply
 errorPopoverMessageFailedWithError=The expected value is ''{0}'' but the actual one is ''{1}''.
 errorPopoverMessageFailedWithException=Error message: {0}
 dateFormatPlaceholder=Insert date as yyyy-mm-dd
+dmnDateFormatPlaceholder=Insert date as date("yyyy-mm-dd")
```
gitgabrio: @jomarko 
Maybe  _Insert date as "yyyy-mm-dd"_  ?

jomarko: The thing is for drools, you need to put in _yyyy-mm-dd_, for DMN you need to put in _date("yyyy-mm-dd")_. The message you propose is already there, however it is true just for drools, not for DMN.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ScenarioSimulationUtils.java
jomarko: ok, will do

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/utils/ConstantHolder.java
```diff
@@ -31,4 +31,6 @@
     // GWT doesn't support Java 8 LocalDate
     public static final String LOCALDATE_SIMPLE_NAME = "LocalDate";
     public static final String LOCALDATE_CANONICAL_NAME = "java.time.LocalDate";
+
+    public static final String DMN_DATE = "date";
```
jomarko: @gitgabrio @manstis me and @dupliaka are not happy I reintroduced an existing constant **date**. It is already for example in the [BuiltInType](https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-api/src/main/java/org/kie/workbench/common/dmn/api/property/dmn/types/BuiltInType.java). Would be possible from your point of view to add a dependency to **kie-wb-common-dmn-api** module? If that is not preferable, is it possible to move **BuiltInType** to some shared module, for example **drools-workbench-models**? Or is it really best to introduce the date string constant here?

gitgabrio: @jomarko 
Hi. One thing that I would like to avoid as the pest - as general rule - is to add other dependencies. 
Beside that, I surely won't do that for a single string (as this seems the case).
Having said that, if there is not already a constant "date" to reuse in all our current classpath, I would not see any issue in using this one. 
The only issue I see is that - in general - we should name objects based on what they _are_, and not on what they _are used for_. In that line, `DMN_DATE` is just _date_ - but "DMN" is only the way you currently use it.
So, if there is not already a  _date_ String somewhere, please keep using this (maybe rename it as "DATE" or something more generally re-usable).
About moving **BuiltInType**, it brings with it _org.kie.workbench.common.dmn.api.property.dmn.QName_, and who knows what in cascade - and all of that only for a single string: again, I do not see a valid reason to do that.

( Out of curiosity - please look that https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-dmn/kie-wb-common-dmn-api/pom.xml: my point of view about dependencies  should be more clear)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
gitgabrio: @jomarko 
Ok, gotcha

jomarko:
@gitgabrio thank you for feedback. I will keep the string constant without adding new dependencies. However I prefer to keep name **DMN_DATE**, because in that case we return **dmnDateFormatPlaceholder**. I think it is better to keep dmn suffix in both places.
Unfortunately I needed to revert switch back to if statements and use **Objects.equals**, otherwise **ScenarioGridModelTest** throws an NPE.

gitgabrio:
> @gitgabrio thank you for feedback. I will keep the string constant without adding new dependencies. However I prefer to keep name **DMN_DATE**, because in that case we return **dmnDateFormatPlaceholder**. I think it is better to keep dmn suffix in both places.
> Unfortunately I needed to revert switch back to if statements and use **Objects.equals**, otherwise **ScenarioGridModelTest** throws an NPE.

gitgabrio:
@jomarko 
Sorry I inadvertently
 clicked the wrong button - PR should be still open anyway

jomarko:
@dupliaka just a reminder, please review when you have a moment.

jomarko:

PR #1252, opened 2019-10-25
DROOLS-4650: Fix failing tests

jomarko:
As part of DROOLS-4650 we implemented equals method of columns. Some tests needed to be updated to differentiate once compared via equals.

Ensemble together with:
https://github.com/kiegroup/drools/pull/2624


jomarko:
@manstis may I ask you for review and merge? together with kiegroup/drools#2624 ?

jomarko:

PR #1260, opened 2019-11-07
RHDM-329: Tests cleanup

jomarko:
The jira was closed as won't fix, so removing the test. For more details see https://issues.jboss.org/browse/RHDM-329


jomarko:
@manstis please merge

PR #1257, opened 2019-11-05
[DROOLS-4647][DROOLS-4701][DROOLS-4732] -Update History (undo & redo actions) to support Background grid + import/export background

gitgabrio:
@yesamer @danielezonca 

See https://issues.jboss.org/browse/DROOLS-4647

PR #1261, opened 2019-11-11
[DROOLS-4732] Removed save button from Settings panel

danielezonca:
Now the only save button is in the toolbar

@gitgabrio @yesamer @dupliaka 

https://issues.jboss.org/browse/DROOLS-4732


danielezonca:
Merged with https://github.com/kiegroup/drools-wb/pull/1257

PR #1268, opened 2019-11-28
[DROOLS-4804] Disabling TestTools when switching grids

yesamer:
@dupliaka @gitgabrio @danielezonca Can you please test and review it?

Very simple, to solve the reported issues, the tool panel it's now disabled after a grid switching.
We agreed to improve the search functionality in a separate ticket.

https://issues.jboss.org/browse/DROOLS-4804


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/rightpanel/TestToolsPresenter.java
```diff
@@ -200,15 +200,9 @@ public void setEventBus(EventBus eventBus) {
     @Override
     public void setGridWidget(GridWidget gridWidget) {
         this.gridWidget = gridWidget;
-        switch (gridWidget) {
```
gitgabrio: @yesamer 
Just double check: this method invoked only on tab switch, correct?

yesamer:
@danielezonca Can you please merge it in master?

PR #1271, opened 2019-12-03
Drools 4767 Fix error-popover position on simulation/background grids

gitgabrio:
@dupliaka @yesamer @danielezonca

See https://issues.jboss.org/browse/DROOLS-4767


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/popover/ErrorReportPopoverPresenter.java
```diff
@@ -16,51 +16,62 @@
...
 
-import org.drools.workbench.screens.scenariosimulation.client.utils.ViewsProvider;
 import org.uberfire.mvp.Command;
 
 @ApplicationScoped
 public class ErrorReportPopoverPresenter implements ErrorReportPopover.Presenter {
 
     @Inject
-    protected ViewsProvider viewsProvider;
+    protected ErrorReportPopoverView errorReportPopupView;
```
yesamer: @gitgabrio With this change, we could remove `errorReportPopupView` from `ViewsProvider` class, because no more used.
I was wondering about that change, probably you did it because it's marked as "@ApplicaitonScoped", right? Or is it just to avoid multiple `.getErrorReportPopupView()` calls?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/popover/AbstractPopoverView.java
```diff
@@ -66,48 +66,61 @@ public AbstractPopoverView(final Div popoverElement,
         this.popoverContainerElement = popoverContainerElement;
         this.popoverContentElement = popoverContentElement;
         this.jQueryPopover = jQueryPopover;
+        options.setContent(element -> popoverContentElement);
+        options.setAnimation(false);
+        options.setHtml(true);
+    }
+
+    public void init() {
+        // no op
```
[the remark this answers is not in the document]
gitgabrio: @yesamer 
Actually it is just a left over - I'll remove that

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/popover/ErrorReportPopoverPresenter.java
gitgabrio: @yesamer 
You're right. I'm not sure why ViewsProvider contains `ErrorReportPopoverView getErrorReportPopoverView()`: since `ErrorReportPopoverView` is ApplicationScoped, there is no need. I'll remove.

gitgabrio:
> Fine for me.
> Considering these changes doesn't support the popup in the BOTTOM position of the cell, I suggest to evaluate to open a new ticket (with minor priority) to implement that, because I'm pretty sure someone will ask for it. Wdyt?

@yesamer 
We will consider it when/if someone will ask. It would require a certain amount of effort, and from my point of view it does not add a big value.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/popover/PopoverView.java
```diff
@@ -22,10 +22,26 @@
 
     enum Position {
         LEFT,
-        RIGHT
+        RIGHT,
+        BOTTOM,
```
jomarko: As we do not use `BOTTOM` anywhere, not sure if we should add it.

gitgabrio: @jomarko 
Done.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/handlers/ScenarioSimulationMainGridPanelMouseMoveHandler.java
```diff
@@ -85,70 +85,92 @@ protected boolean manageBodyCoordinates(Integer uiRowIndex, Integer uiColumnInde
...
+        if (xPosition <= scenarioGrid.getLayer().getElement().getAbsoluteLeft()) {
+            xPosition = (int) cellXYMiddleCoordinates.getX();
+            yPosition = (int) cellXYMiddleCoordinates.getY() - cellHeight/2;
+            position = PopoverView.Position.TOP;
+        }
+        setupPopupPresenter(toManage, uiRowIndex, uiColumnIndex, xPosition, yPosition, position);
+        errorReportPopupPresenter.show();
+    }
+
+    protected void setupPopupPresenter(FactMappingValue toManage, Integer uiRowIndex, Integer uiColumnIndex, int xPosition, int yPosition, PopoverView.Position position) {
```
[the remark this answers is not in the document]
gitgabrio: > * One comment to code.
> * Brief manual check, one question, is this just about Background tab? The jira says ordering of preferences is Right, Left, Top. What seems working on background tab, but on Model tabl I would expect Right in the attached screenshot.
> 
> Screenshot from 2019-12-05 11-44-05

@jomarko 
Unfortunately it is a bug in master. Behavior is broken also if you extend/collapse the right panel. Could you please open a specific Jira ticket? Thx

PR #1273, opened 2019-12-13
DROOLS-4871 : Complex value list doesn't work in guided rule editor

Rikkola:
https://issues.redhat.com/browse/DROOLS-4871

Part two of the fix. This fixes from editor to file changes.

https://github.com/kiegroup/drools/pull/2689
https://github.com/kiegroup/drools-wb/pull/1273

### DIFF
--- a/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/ConstraintValueChangeHandler.java
+++ b/drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/ConstraintValueChangeHandler.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


drools-wb-screens/drools-wb-guided-rule-editor/drools-wb-guided-rule-editor-client/src/main/java/org/drools/workbench/screens/guided/rule/client/editor/ConstraintValueChangeHandler.java
```diff
@@ -0,0 +1,54 @@
...
+    public ConstraintValueChangeHandler(final BaseSingleFieldConstraint constraint,
+                                        final Command onChangeCommand) {
+        this.constraint = constraint;
+        this.onChangeCommand = onChangeCommand;
+    }
+
+    @Override
+    public void onValueChange(ValueChangeEvent<String> event) {
+
+        if (Arrays.asList(in, not_in).contains(constraint.getOperator())) {
```
jomarko: Please use `Arrays.asList(OperatorsOracle.EXPLICIT_LIST_OPERATORS)` instead of `Arrays.asList(in, not_in)`

manstis: Thanks @jomarko that was something I was meaning to check!

Rikkola:
@jomarko The templates continues to have redundant code for this, but now fixed.

jomarko:
@manstis I think we can merge this ensemble.

PR #1274, opened 2019-12-17
DROOLS-4876: GDST: Aliases only work for Strings

Rikkola:
https://issues.redhat.com/browse/DROOLS-4876

The `isValueValidForType `tests with cast if the value is boolean, interger or anything, but String. For String everything goes. This change trims the checked value from the first = character. This means only the actual value is checked and not the alias.

### DIFF
--- a/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/utilities/ColumnUtilities.java
+++ b/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/utilities/ColumnUtilities.java
@@ -145,13 +145,22 @@ public class ColumnUtilities
 
         final String[] values = ListSplitter.split("'", true, valueList);
 
-        return Stream.of(values).filter(value -> isValueValidForType(value,
+        return Stream.of(values).filter(value -> isValueValidForType(parseActualValue(value),
                                                                      convertToTypeSafeType(fieldType)))
                 .toArray(String[]::new);
     }
 
+    private String parseActualValue(final String value) {
+        if (value != null && value.indexOf("=") > 0) {
+            return value.substring(0, value.indexOf("="));
+        } else {
+            return value;
+        }
+    }
+
     private boolean isValueValidForType(String value,
                                         DataType.DataTypes type) {
+
         switch (type) {
             case STRING:
                 return true;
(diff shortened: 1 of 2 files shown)


Rikkola:
I keep asking the same people, but @jomarko and @manstis if you could again take a look thank you.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/java/org/drools/workbench/screens/guided/dtable/client/widget/table/utilities/ColumnUtilities.java
```diff
@@ -145,13 +145,22 @@ public static boolean canAcceptOtherwiseValues(final BaseColumn column) {
...
         final String[] values = ListSplitter.split("'", true, valueList);
 
-        return Stream.of(values).filter(value -> isValueValidForType(value,
+        return Stream.of(values).filter(value -> isValueValidForType(parseActualValue(value),
                                                                      convertToTypeSafeType(fieldType)))
                 .toArray(String[]::new);
     }
 
+    private String parseActualValue(final String value) {
+        if (value != null && value.indexOf("=") > 1) {
```
manstis: I think this should be `> 0` since a value list item of `1=One` does not work.

Rikkola: @manstis Well spotted and now fixed.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/widget/table/utilities/ColumnUtilitiesTest.java
```diff
@@ -146,6 +146,32 @@ public void testGetValueConditionColumn() throws Exception {
...
+        final String[] valueList = utilities.getValueList(column);
+        Assertions.assertThat(valueList).containsExactly("true=Yes", "false=No");
+    }
+
+    @Test
+    public void testBooleanAliasesInvalidValues() {
+        column.setFactField(FIELD_NAME);
+        column.setFieldType(DataType.TYPE_BOOLEAN);
+        column.setValueList("yes=Yes, no=No");
+        assertEquals(0, utilities.getValueList(column).length);
```
jomarko: Can be rewritten as `Assertions.assertThat(utilities.getValueList(column)).isEmpty()`

Rikkola:
@kiegroup/gatekeepers Can I have a merge please.

PR #1279, opened 2020-01-08
Added handling an end key for selecting bottom right cell in scenario…

dupliaka:
… editor for testing purposes.
Based on:
https://issues.redhat.com/browse/DROOLS-4923


dupliaka:
jenkins please retest this

dupliaka:
jenkins please retest this

dupliaka:
@kiegroup/gatekeepers  please merge

manstis:
Don't we need a QE review too @dupliaka / @yesamer?

yesamer:
@manstis Honestly in this case, where a QE opened the PR, I don't know exactly how to proceed. Probably we can ask to @jomarko to provide a QE review?

jomarko:
However I realized one thing. We should create a jira in DROOLS project about this change, otherwise we have no record about this new feature. In that jira we can set appropriate "fix version". So if anyone reading release notes will wind out this new feature.what do you think @dupliaka @yesamer ?

dupliaka:
@jomarko - I am fully agree with you that this is a new feature and it could be lost, but I am not happy with the process when
1) You need to change production code to test the feature 
2) You need to pass the full cycle of approve  for the feature that is Only For Testing and not regards to the application functionality
3) If that takes so much effort I would find another way to do that.

jomarko:
@dupliaka not sure I get you point, but I would like to say, that I do not see this situation in some negative way.

You came to that feature in the middle of process as new team member and realized you need this shortcut for testing with some delay.

I do not see as negative thing that we are adding some code to to product to facilitate testing. Even more, the added code is completely valid for any human user.

If you do not want to invest more time into this PR, fine for me, take my note just a note for future.

dupliaka:
@jomarko I created an jira ench:https://issues.redhat.com/browse/DROOLS-4923

jomarko:
@dupliaka thank you, please include DROOLS-4923 in the PR summary/title and we can merge

dupliaka:
@jomarko Done, thanks!

dupliaka:

PR #1283, opened 2020-01-16
DROOLS-4945 : Deficient Rows warning shows when Decision Table references items in a list

Rikkola:
https://issues.redhat.com/browse/DROOLS-4945

Bundle:
https://github.com/kiegroup/drools/pull/2727
https://github.com/kiegroup/drools-wb/pull/1283


jomarko:

Rikkola:
@jomarko This PR should now build the webworker js.

Rikkola:
But this needs to be retriggered 

drools-wb-services/drools-wb-verifier/drools-wb-verifier-webworker/pom.xml
```diff
@@ -77,6 +77,7 @@
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>gwt-maven-plugin</artifactId>
         <configuration>
+          <skip>false</skip>
```
Rikkola: This does do the trick, but not sure if this is the best way. For some reason drools-wb-showcase runs by GWT by default, but there is no easy ( for me ) to spot reason why the webworker should not do the same. Spending some more time on investigating this.

Rikkola:
@manstis if you could please review.

Rikkola:

PR #1284, opened 2020-01-17
RHPAM-2492: Cannot validate list in legacy Test Scenario

karreiro:
See https://issues.redhat.com/browse/RHPAM-2492

Before:

After:


manstis:
@karreiro I've assigned @Rikkola instead of me as he knows (knew!) the area better.

PR #1285, opened 2020-01-17
[DROOLS-4949]: Blue rectangle above test scenario collection editor

yesamer:
@danielezonca @jomarko 
It seems that useless styles properties generate the problem, and in particular the `Padding` one.

https://issues.redhat.com/browse/DROOLS-4949


jomarko:

PR #1288, opened 2020-01-20
RHDM-1198: Change use date settings don't be persisted on Test Scenario Legacy

karreiro:
See https://issues.redhat.com/browse/RHDM-1198

Before:
before

After:
after


jomarko:

PR #1294, opened 2020-01-27
DROOLS-4976 : Guided Decision Table Graphs do not work

Rikkola:
https://issues.redhat.com/browse/DROOLS-4976


Rikkola:
@jomarko @manstis I was missing a setup for the graphs editor. If you could review the change please.

PR #1299, opened 2020-01-30
DROOLS-4972: Buttons stay disabled when switching to textual collection editor

yesamer:
@danielezonca @jomarko @dupliaka 

https://issues.redhat.com/browse/DROOLS-4972


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/collectioneditor/CollectionView.java
```diff
@@ -147,13 +151,11 @@
...
-    ButtonElement getCancelButton();
-
-    ButtonElement getRemoveButton();
-
-    ButtonElement getSaveButton();
+    /**
+     * It enables/disables buttons group which manage the Collection Editor
+     * @param enable
+     */
+    void disableCollectionEditorButtons(boolean enable);
```
dupliaka: I feel that confusing-parameter name should be disable according to the function name.
Enable of disableness is hard to imagine)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/collectioneditor/CollectionPresenter.java
```diff
@@ -70,6 +70,11 @@
      */
     protected Map<String, Map<String, Map<String, String>>> expandablePropertiesMap = new HashMap<>();
 
+    /**
+     * It defines if <b>Create collection</b> guided editor is in editing status or not.
+     */
+    protected boolean createCollectionIsEditing = false;
```
danielezonca: What about define an enum to keep the status of the whole panel? Something like `GUIDED`, `EXPRESSION` and `GUIDED_ADDING_ELEMENT`. I think it is easier to understand

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/collectioneditor/CollectionView.java
yesamer: @dupliaka You're right, I forgot to update the name.

yesamer:
@danielezonca @dupliaka I changed the fix, now It simply disables the "Define Collection" radio button when Editing mode is active. This simplify the logic and the change (a status is no longer required). Thank you for your suggestions.

yesamer:
@kiegroup/gatekeepers Can your please merge it?

PR #1311, opened 2020-02-12
DROOLS-5059 : Verifier: GDST columns empty enumeration should be ignored

Rikkola:
https://issues.redhat.com/browse/DROOLS-5059


drools-wb-services/drools-wb-verifier/drools-wb-verifier-guided-decision-table-adapter/src/main/java/org/drools/workbench/services/verifier/plugin/client/builders/ValuesResolver.java
```diff
@@ -377,6 +377,6 @@ private Integer getInteger() {
     private boolean isTypeGuvnorEnum() {
         return conditionColumn != null &&
                 conditionColumn.getValueList() != null &&
-                conditionColumn.getValueList().split(",").length != 0;
+                conditionColumn.getValueList().split(",").length > 1;
```
Rikkola: Problem was that "" value list would return true and the Verifier would then think this was an enumeration column.

drools-wb-services/drools-wb-verifier/drools-wb-verifier-guided-decision-table-adapter/src/test/java/org/drools/workbench/services/verifier/plugin/client/fromfile/DecisionTableAnalyzerFromFileTest.java
```diff
@@ -127,6 +129,29 @@ public void testFileLargeFileGDST() throws
                            CheckType.EMPTY_RULE);
     }
 
+    @Test
+    public void emptyValueListOnColumnShouldNotCountAsAnEnum() throws
+            Exception {
+
+        analyzerProvider.getFactTypes().add(new FactTypes.FactType("Data",
+                                                                   new HashSet<FactTypes.Field>() {{
```
jomarko: could we please use https://github.com/kiegroup/kie-soup/blob/master/kie-soup-commons/src/main/java/org/kie/soup/commons/util/Sets.java ? Those builders were created to eliminate double brace static field initialization - it is marked as bad pattern by sonar.

yesamer:
jenkins retest this please

drools-wb-services/drools-wb-verifier/drools-wb-verifier-guided-decision-table-adapter/src/test/java/org/drools/workbench/services/verifier/plugin/client/fromfile/DecisionTableAnalyzerFromFileTest.java
Rikkola: Done

Rikkola:
Build stuck so 
jenkins please retest

jomarko:
Will re-review once this is converted to standard PR and build passed.

Rikkola:
@yesamer would you like to review? Not much to look at, one line change :)

drools-wb-services/drools-wb-verifier/drools-wb-verifier-guided-decision-table-adapter/src/main/java/org/drools/workbench/services/verifier/plugin/client/builders/ValuesResolver.java
jomarko: Shouldn't we handle case when single enum is specified?
```java
private boolean isTypeGuvnorEnum() {
        if (conditionColumn != null && conditionColumn.getValueList() != null) {
            if (conditionColumn.getValueList().split(",").length > 1) {
                return true;
            } else {
                return conditionColumn.getValueList().trim().length() > 0;
            }
        } else {
            return false;
        }
    }
```

Rikkola: I'll try this out. Might be enough to just check if value list empty without splitting it.

Rikkola: I changed it to:
```
    private boolean isTypeGuvnorEnum() {
        return conditionColumn != null &&
                conditionColumn.getValueList() != null &&
                !conditionColumn.getValueList().trim().isEmpty();
    }
```

Rikkola:
@kiegroup/gatekeepers Can I have a merge please.

PR #1329, opened 2020-03-10
DROOLS-5134: Allow code coverage measurement for whole project

jomarko:


jomarko:
@rsynek thank you, I ahve incorporated point [1]. However not sure about [2]. Do you think I should add system property in `kie-parent` for that will be passed to configuration of `sonarcloud-analysis` profile directly in `kie-parent` and in `drools-wb` I will set this new variable? I am not sure because this way there is also risk of renaming the variable just on one place.

rsynek:
@jomarko I think we misunderstood each other.
`<sonar.coverage.exclusions>**/editor/ScenarioSimulationEditorWrapper.java,**/events/**.java</sonar.coverage.exclusions>`
is fine to stay in the same file where it is right now. It's just unnecessary and, in fact, error-prone, to override the `sonarcloud-analysis` profile there. It should just move to the `<properties/>` section of the pom.xml.

yesamer:
@jomarko I'm not totally agree to enable code coverage on all these projects. I'm my understanding most of these project are no more maintained and with low test coverage. As a result I see a total low test coverage of the overall project. I think we should enable it for the current active projects only.

jomarko:
@rsynek thank you, I pushed changes in new commit.

@yesamer I do not see it as a big issue. We assess each PR individually. If we see (occasionally it happens) PR for the obsolete editor, we can use sonar analysis just as check new code is covered. I do not think someone will ask to cover whole obsolete editor.

yesamer:
@jomarko ok, for evaluating a single PR I agree with you, in that case is useful and correct. My concern is more on the coverage of the overall drools-wb project, please check here: https://sonarcloud.io/dashboard?id=org.drools%3Adrools-wb
Currently we have 71% of overall coverage, because it covers scesim project only. If we include all other projects, we will have a low coverage as a result. This because, as I said, some of these projects are poor tested and no more maintained. For this reason, I suggest to add only the active projects, otherwise we will have a low coverage of the overall project forever, witout any chance to increase it (I think there is no sense to increase test coverage of a no-maintained project)

jomarko:
I have no strong preference and I am fine with both approaches (enable just for maintained vs all). @rsynek as sonar kiegroup expert asked me to have a look on all modules, so what is your preference please?

rsynek:
@yesamer @jomarko It's true that the project's coverage will dramatically drop. However, you will get:
- overview of the real coverage of the entire project
- coverage on PR targetting any part of the codebase, including the less active parts

Please note that coverage on PRs is calculated only for the changed codebase, thus the default, non-mandatory, threshold of 80% on the PR check will be perfectly reachable.

I would argue that any part of the codebase which, in the end, lands in a release, is relevant and we should be at least aware of its coverage rather than pretending it doesn't exist.

jomarko:

PR #1338, opened 2020-03-18
KOGITO-1496: Removing ACE and monaco-editor form kogito-runtime module

yesamer:
@danielezonca @jomarko Can you please review and test?

Goal of this PR is to reduce generated war file of `kogito-rutime` module.
Analyzing it, I noticed `ace` (9,4 MB Unzipped) and `monaco-editor` (6.7 MB Unzipped) were added in the WAR even if not currently used by SCESIM. 
Removing it, the new size of WAR is now **4.8 MB** (it was 7.9 MB).
To do that, I excluded the following resources with `maven-war-plugin`:
- All `ace` folder and sub-folders. `js` link in `index.html` file was removed as well
- All `monaco-editor` folder and sub-folders with the exception of `/monaco-editor/dev/vs/editor/editor.main.css`, which is required inside `PatternFly.gwt.xml` file.

Unfortunately, It's not easily possible to remove these 2 module from the source code, because they are inside a complex dependency tree.


danielezonca:
@yesamer 
I agree about monaco-editor but not sure ACE. From what I remember with DMN/BPMN editors, if there is any error during the loading of a file, you will get an XML text editor that is ACE-based. Can you try with a malformed scesim file to see if I'm right?

manstis:
@danielezonca @yesamer Fallback to ACE for DMN and BPMN is "a Stunner thing".

_scesim_ is not based on Stunner so does not have similar fallback and hence does not need ACE.

yesamer:
> Please remove also
> 
> ```
>     
>     <dependency>
>       <groupId>org.uberfire</groupId>
>       <artifactId>uberfire-widgets-core-client</artifactId>
>       <scope>provided</scope>
>     </dependency>
> ```

@jomarko Done, thank you !

manstis:
@yesamer Please be aware of this https://github.com/kiegroup/appformer/pull/927

IDK if it'll lead to you having to change the paths you exclude on this PR...

jomarko:
@manstis thank you for heads-up. I think we are fine due to :
```
**/monaco-editor/dev/vs/base/**,
**/monaco-editor/dev/vs/basic-languages/**,
```

PR #1341, opened 2020-03-25
KOGITO-1568: Show an error if referred DMN file is missing

yesamer:
@danielezonca @dupliaka Can you please test and review?

Very small change, in case a Scesim file refers to a missing / no more present DMN File, an error should be shown to the user.

Screenshot from 2020-03-25 14-38-21

https://issues.redhat.com/browse/KOGITO-1568


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/resources/org/drools/workbench/screens/scenariosimulation/client/resources/i18n/ScenarioSimulationEditorConstants.properties
```diff
@@ -135,6 +135,7 @@ removeCollectionWarningText=This action cannot be undone.
 #
 settings=Settings
 dmnPathErrorLabel=Impossible to find ''{0}'' file. Please select one.
+dmnPathErrorDetailedLabel=Impossible to find referred DMN file: ''{0}''. Please provide it and update setting panel.
```
dupliaka: Could we please change it to be mor concrete on what should be an update
Ex: "Please provide it and link in Settings"

yesamer: @dupliaka Done!

PR #1333, opened 2020-03-16
DROOLS-5038: Create testing webapp module

yesamer:
@danielezonca @dupliaka Can you please test and review it?

Testing module for kogito scesim. This is very useful for testing and debugging purposes. 

https://issues.redhat.com/browse/DROOLS-5038

PR #1345, opened 2020-04-02
KOGITO-1566: [SCESIM Editor] DMN Type with nested objects doesn’t works correctly

danielzhe:
It may be a little time expensive to test before [this other PR](https://github.com/kiegroup/drools-wb/pull/1333) get merged. You can cherry-pick the commit from this PR in [this branch](https://github.com/yesamer/drools-wb/tree/DROOLS-5038) if you want and use the scesim testing webapp.

The idea behind this change is:
1. Create all the DMN types in a recursive way ignoring its fields
2. Resolve the DMN types fields

Using this approach you solve [the issue reported in the JIRA](https://issues.redhat.com/browse/KOGITO-1566) and also the following situation:

tPerson [structure]
-name[string]
-address[tAdress]

tAddress[structure]
-street[String]
-country[tCountry]

tCountry
-name[String]
-president[tPerson]

Notice that tPerson has a tAddress which have a tCountry which have a tPerson and so on.
This PR handles this situation because it creates all types first and then populate the fields, which will work because in this point all types is already present.

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/AbstractKogitoDMNService.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/AbstractKogitoDMNService.java
@@ -16,8 +16,8 @@
 package org.drools.workbench.screens.scenariosimulation.kogito.client.dmn;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -31,7 +31,6 @@ import java.util.TreeSet;
 import javax.xml.namespace.QName;
 
 import jsinterop.base.Js;
-
 import org.drools.workbench.screens.scenariosimulation.kogito.client.dmn.feel.BuiltInType;
 import org.drools.workbench.screens.scenariosimulation.model.typedescriptor.FactModelTree;
 import org.drools.workbench.screens.scenariosimulation.model.typedescriptor.FactModelTuple;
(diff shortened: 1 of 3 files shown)


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/test/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/AbstractKogitoDMNServiceTest.java
```diff
@@ -171,6 +294,74 @@ public void getFactModelTupleSimpleInputAndDecisionData() {
         assertTrue(decisionDataNameFact.getSimpleProperties().values().contains("string"));
     }
 
+    @Test
+    public void testGetRecursiveDMNTypes() {
```
dupliaka, line 298: Could we also add a test for nested structures of the same type like
        tPeople[structure]
            nextGroup[tPeople]

yesamer:
@danielezonca can you please rebase with master?

danielzhe:
> 
> @danielezonca can you please rebase with master?

I think you marked wrong Daniel. Rebased!

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/test/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/AbstractKogitoDMNServiceTest.java
dupliaka: Well apparently that is not working in BC as well. The Ui in that case makes it impossible to use except by using expressions, so no need for changes here

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/AbstractKogitoDMNService.java
```diff
@@ -107,60 +105,74 @@ protected ClientDMNType getDMNTypeFromMaps(final Map<String, ClientDMNType> dmnT
...
-         * referred by typeRef field, BEFORE their referred items.  */
-        jsitItemDefinitions.sort(getItemDefinitionComparator());
+        final Map<String, JSITItemDefinition> indexed = indexDefinitionsByName(jsitItemDefinitions);
         for (int i = 0; i < jsitItemDefinitions.size(); i++) {
             final JSITItemDefinition jsitItemDefinition = Js.uncheckedCast(jsitItemDefinitions.get(i));
-            toReturn.put(jsitItemDefinition.getName(), getDMNType(jsitItemDefinition, nameSpace, toReturn));
+            toReturn.put(jsitItemDefinition.getName(), getDMNType(indexed, jsitItemDefinition, nameSpace, toReturn));
         }
+
+        toReturn.values().forEach((v) -> {
```
yesamer, line 114: @danielzhe Here, a `ConcurrentModificationException` can be thrown, because it iterates over  `toReturn` list which can be modified inside`populateItemFields(...)` method. I found it using the dmn file posted here by @dupliaka  https://issues.redhat.com/browse/KOGITO-1685.

danielzhe:
Thank you! Changes applied. @yesamer

PR #1352, opened 2020-04-20
DROOLS-5255: Wait with field page setup until model is loaded

jomarko:
The issue is in FieldPage#setupListFieldView method. It is invoked twice. Unfortunately was not able to investigate why.

The issue is it clears the field select box immediately when invoked and do not wait for new fields are loaded.

So basically when setupListFieldView is invoked it clears select box of fields, however first load/callback was not finished that time.

For more details see: https://issues.redhat.com/browse/DROOLS-5255


jomarko:

PR #1353, opened 2020-04-22
DROOLS-5250 (ex KOGITO-1685): Wrong label name for multiple nested elements

yesamer:
@danielezonca @dupliaka Can you please review and test?

Basically, the label  of nested properties was wrong in the case of multiple nested element. 
This is because, multiple nested properties shared the same list used to determine the label and the `expressionElement` list. 
To solve the issue, I simply created a new list of previous step for any nested property.

The fix is applied for both KOGITO and BC. Considering they are not sharing this part of code (in BC is on server side), the fix is applied in 2 different points

https://issues.redhat.com/browse/DROOLS-5250

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/KogitoScenarioSimulationBuilder.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/KogitoScenarioSimulationBuilder.java
@@ -359,7 +359,7 @@ public class KogitoScenarioSimulationBuilder {
         }
     }
 
-    private void addFactMapping(final FactMappingExtractor factMappingExtractor,
+    protected void addFactMapping(final FactMappingExtractor factMappingExtractor,
                                 final FactModelTree factModelTree,
                                 final List<String> previousSteps,
                                 final Map<String, FactModelTree> hiddenValues) {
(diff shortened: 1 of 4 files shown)


dupliaka:

dupliaka:
Unfortunately I could not find out a variable "nested" in tNested of  Testing Tools
See: 
image
 Please take a look

dupliaka:
For context type after this update i noticed that i can create two facts of the same types:
image

dupliaka:
Types are non connected 
image
tNested does not contains string parameter nested
And contextType contains value but should not.

yesamer:
@dupliaka  I'll manage these issue in https://issues.redhat.com/browse/KOGITO-1747, where I'm fixing several issues (not only the collection one).

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/KogitoScenarioSimulationBuilder.java
```diff
@@ -359,7 +359,7 @@ protected static double getColumnWidth(String expressionIdentifierName) {
         }
     }
 
-    private void addFactMapping(final FactMappingExtractor factMappingExtractor,
+    protected void addFactMapping(final FactMappingExtractor factMappingExtractor,
                                 final FactModelTree factModelTree,
                                 final List<String> previousSteps,
                                 final Map<String, FactModelTree> hiddenValues) {
```
jomarko, line 365: nitpicking - formatting is broken

yesamer: @jomarko Done :)

PR #1358, opened 2020-05-06
KOGITO-1722: Remove "Run scenario" from context menu in Kogito

yesamer:
@danielezonca @dupliaka Can you please review and test it?

This PR removes "Run Scenario" item present into Context Menu for Kogito channels.
Beside this, I removed `UnmodifiableColumnGridContextMenu` because it acts as `GridContextMenu` considering the items are the same

https://issues.redhat.com/browse/KOGITO-1722


yesamer:

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/editor/menu/GridContextMenuTest.java
```diff
@@ -111,65 +103,43 @@ public void show(GridWidget gridWidget, int mx, int my) {
...
-        verify(gridContextMenuSpy, times(1)).mapEvent(eq(duplicateRowLIElementMock), isA(DuplicateRowEvent.class));
-        verify(gridContextMenuSpy, times(1)).mapEvent(eq(deleteRowLIElementMock), isA(DeleteRowEvent.class));
-        verify(gridContextMenuSpy, never()).addExecutableMenuItem(eq(GRIDCONTEXTMENU_RUN_SINGLE_SCENARIO), eq(ScenarioSimulationEditorConstants.INSTANCE.runSingleScenario()), eq("runSingleScenario"));
-        verify(gridContextMenuSpy, never()).mapEvent(eq(createdElementMock), isA(RunSingleScenarioEvent.class));
-        verify(gridContextMenuSpy, times(1)).updateMenuItemAttributes(eq(gridTitleElementMock), eq(GRIDCONTEXTMENU_GRID_TITLE), eq(ScenarioSimulationEditorConstants.INSTANCE.background()), eq("background"));
+    protected void show(GridContextMenu gridContextMenu, GridWidget gridWidget, int mx, int my, int rowIndex) {
+        String expectedLabel;
+        String expectedI18n;
+        if (GridWidget.BACKGROUND.equals(gridWidget)) {
+            expectedLabel = ScenarioSimulationEditorConstants.INSTANCE.background();
```
dupliaka, line 131: @yesamer 
I feel that the cause in test complicates the test logic. Could you please rewrite it as a separate tests, moving all verification statements to a separate method, please.

Also can we have here a check for notNullRunScenarioElement and nullRunscenarioElement? I can see only first case. The same for background

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/menu/ScenarioContextMenuRegistry.java
```diff
@@ -49,7 +48,6 @@
     protected GivenContextMenu givenContextMenu;
     protected ExpectedContextMenu expectedContextMenu;
     protected GridContextMenu gridContextMenu;
-    protected UnmodifiableColumnGridContextMenu unmodifiableColumnGridContextMenu;
```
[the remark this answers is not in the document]
yesamer: @danielezonca @dupliaka Yes, I removed it because its behavior is a copy of `GridContextMenu` in my opinion. 
Please consider, `UnmodifiableColumnGridContextMenu` is shown on grid data cell of the first and second column with this context menu:
Screenshot from 2020-05-06 16-34-13
And `GridContextMenu` is show in the remaining columns, with this context menu:
Screenshot from 2020-05-06 16-34-09
As you can see, the only difference is between the missing "Scenario" title. The items are the same. I think i's correct to have the "Scenario" title for the first columns as well.

yesamer:

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/menu/ScenarioContextMenuRegistry.java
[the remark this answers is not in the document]
danielezonca: Ok thanks for the clarification. Probably at the beginning it wasn't like that

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/editor/menu/GridContextMenuTest.java
yesamer: @dupliaka 
1. I'm not totally sure that splitting the method as you suggested will improve the logic readability. That change will require to refer to 2 methods, in case someone need to modify / change it. I currently see it as a part of atomic test.
2. Those tests are present on `BusinessCentralGridContextMenuTest`, because `runScenario` item was removed from `GridContextMenu` and moved in `BusinessCentralGridContextMenu`

dupliaka: 1. So you do belief that it is atomic even  if you got bias inside? 
One of the definitions:
"Atomic tests are order-independent, relying on and causing no side effects.  They neither rely on nor alter shared state; they do not mutate their environment. If a test run involves 2, 10, 100 or 1,000,000 tests, the tests should be able to be executed in any order.  If one legitimately fails because a bug has been introduced in the production code, it shouldn’t have a domino effect. "
2. Ok, sorry missed that.

yesamer: @dupliaka The aim of the logic I wrote is just to define the expected values based on the given input, `GridWidget` type in particular. With this solution, I'm able to avoid to duplicate the same `verify` over all test cases, managing different expected values based on the input.
This logic doesn't affect shared state and doesn't mutate the environment, can be executed in any order. Doesn't affect production code at all, because it just used to define the expected conditions values.
The only possible issue I see here is to introduce a bug in this logic. Even in this case, the issue will affect the expected values, and not the production code. It would be the same if I put a wrong expected value in a test eg:
`int a = 2+2; 
assertEquals(5, a)`
On the other side, let assume the test is not atomic. I don't think that splitting it will solve the problem. WDYT?

dupliaka: I see your point, but what if 
1) At a certain time we add new GridWidget and it would be not a BACKGROUND and not a SIMULATION? Will that test still test scenario case or rather something different? 
2) If there would be more people who would admit that the if-else inside tests take more time to figure out what the test doing an how to fix that - would it change your belief?

yesamer: @dupliaka I got your point and I applied a change witch removes the `if else` and moves the expected values in the input fields. Does this make sense?

yesamer:
jenkins retest this please

yesamer:
@kiegroup/gatekeepers can you please merge it?

PR #1364, opened 2020-05-13
DROOLS-5317: Scenario Simulation shows misleading data type if DMN applies a constraint

yesamer:
@danielezonca @dupliaka Can you please review and test?

The reason on the issue relies on how a a DMNType for a property with `allowedType` is handled in `DMNCompilerImpl:562` logic. In details, if a property has allowedValues (eg Enum of specif values), its `DMNType` fields are populated in this way, according to DMN Specs (In this case is managed as an Anonymous inner type)
- `DMNType.name` contains the property name and not the property type
- `DMNType.baseType.name` cointans the property type.
Considering this, I adapted the code in Scesim handling the specific case for a property with allowedValues. This is two places: `FactModelTree` creation and scesim file validation.

It reproduces on BC only.

https://issues.redhat.com/browse/DROOLS-5317
https://issues.redhat.com/browse/RHDM-1153


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-backend/src/test/java/org/drools/workbench/screens/scenariosimulation/backend/server/DMNScenarioValidationTest.java
```diff
@@ -205,6 +218,21 @@ private void createDMNType(String decisionName, String rootType, String... steps
         mapOfMockDecisions.put(decisionName, decisionNodeMock);
     }
 
+    private void createDMNTypeWithAllowedValues(String decisionName, String rootType, String step, List<DMNUnaryTest> allowedTypes) {
```
dupliaka, line 221: Could we also add a test where  dmnType.getAllowedValues() == null || dmnType.getAllowedValues().isEmpty()? to check that    when(currentType.getBaseType()).thenReturn(baseType);
I mean to check SImple type and custom type.

dupliaka:
Got interesting issue
image
countryCode.value 
CountryCodeMapping.value 
are simple strings with constraints. could you please look a bit ?
So the bug reproduces for non-nested constraints fields
.
[Traffic Violation.zip](https://github.com/kiegroup/drools-wb/files/4650579/Traffic.Violation.zip)

dupliaka:
- [x] Type of parameter of structure with constraints in Test scenario equals DMN  
- [x] Kogito nested constraint fields scenarios generate with the same type as DMN
- [ ] Type of parameter of Simple Type with constraints in Test scenario equals DMN 
- [x] Type without constraints equals DMN
- [x] Type with nested constraints are the same for imported scesim and for created one

yesamer:
@dupliaka I changed the logic to include Simple types and I updated test according to these changes.

dupliaka:

yesamer:
@kiegroup/gatekeepers Can you please merge it?

PR #1365, opened 2020-05-14
DROOLS-5210 : BRL Column support for GDST to XLS export

Rikkola:
https://issues.redhat.com/browse/DROOLS-5210


Rikkola:
@jomarko It looks like the GDST BRL Columns are buggy. Take the BRL.gdst I'm using here for example and check the source it produces. The conversion however should convert the GDST correctly even if the original file has issues.

jomarko:
@Rikkola hello, thank you for PR could you please have a look on the vulnerability sonarcloud reports? Seems worth and easy to fix

https://sonarcloud.io/project/issues?id=org.drools%3Adrools-wb&open=AXISOP5IW8-XSEm0kOWQ&pullRequest=1365&resolved=false&types=VULNERABILITY

Rikkola:
@jomarko Yes that should be removed, I just used that when debuggin.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/SubHeaderBuilder.java
```diff
@@ -105,6 +128,99 @@ public void build() {
...
+
+    private String subString(String marshal,
+                             final String from,
+                             final String to) {
+        marshal = marshal.substring(marshal.indexOf(from) + from.length());
+        marshal = marshal.substring(0, marshal.lastIndexOf(to));
+
+        marshal = Arrays.stream(marshal.split("\\r?\\n")).map(String::trim).collect(Collectors.joining("\n"));
+        return marshal;
+    }
```
jomarko, line 184: We shouldn't combine two approaches
- updating computed value directly into method argument
- returning computed value as return value of function

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/SubHeaderBuilder.java
```diff
@@ -105,6 +128,99 @@ public void build() {
...
+
+        for (int i = 0; i < dt.getExpandedColumns().size(); i++) {
+            list.add(dtable.getData().get(0).get(i));
+        }
+
+        for (int i = 0; i < brlColumn.getChildColumns().size(); i++) {
+            if (brlColumn.getChildColumns().get(i).getVarName().isEmpty()) {
+                list.add(new DTCellValue52(true));
+            } else {
+                list.add(new DTCellValue52("$" + (i + 1)));
```
jomarko, line 209: Shouldn't we do same as for conditions?
```
  if (!varName.startsWith("$")) {
                    key = varName;
                    varName = "@{" + varName + "}";
                } else {
                    key = Integer.toString(i);
                    varName = "@{" + i + "}";
                }
```

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/DataBuilder.java
```diff
@@ -99,39 +105,106 @@ public void build() {
...
+                                }
+                                break;
+                            } else {
+                                result.append(value);
+                            }
+                            if (i < size - 1) {
+                                result.append(", ");
+                                sourceColumnIndex++;
+                            }
+                        }
```
jomarko, line 150: This is equivalent to:
```
if (baseColumn instanceof BRLVariableColumn) {

                        final StringBuilder result = new StringBuilder();
                        final BRLColumn brlColumn = dtable.getBRLColumn((BRLVariableColumn) baseColumn);
                        final int size = brlColumn.getChildColumns().size();
                        for (int i = 0; i < size; i++) {
                            final String value = getValue(row.get(sourceColumnIndex));
                            if (isMissingVariable(brlColumn)) {
                                if (Objects.equals("true", value.toLowerCase())) {
                                    result.append("X");
                                }
                                break;
                            } else {
                                result.append(value);
                            }
                            if (i < size - 1) {
                                result.append(", ");
                                sourceColumnIndex++;
                            }
                        }

                        xlsRow.createCell(targetColumnIndex)
                                .setCellValue(result.toString());
                    } 
```

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/SubHeaderBuilder.java
```diff
@@ -105,6 +128,99 @@ public void build() {
...
+            }
+        }
+        dt.getData().add(list);
+
+        dt.getActionCols().add(brlColumn);
+
+        final String marshal1 = GuidedDTDRLPersistence.getInstance().marshal(dt);
+        final String marshal = subString(marshal1, "then", "end");
```
jomarko, line 217: what if data objects/fields contains these substrings?

Rikkola: True this can be made a bit smarter. I think since this comes from a generated source it would be safe to check for lines that contain these values.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/DataBuilder.java
```diff
@@ -99,39 +105,106 @@ public void build() {
...
+                            }
+                        } else if (dtable.getExpandedColumns().get(sourceColumnIndex) instanceof ActionRetractFactCol52) {
+
+                            if (!StringUtils.isEmpty(cell.getStringValue())) {
+                                xlsRow.createCell(targetColumnIndex)
+                                        .setCellValue(cell.getStringValue());
+                            }
+                        } else {
+
+                            setValue(cell, getColumnDataType(sourceColumnIndex));
```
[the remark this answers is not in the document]
Rikkola: I can try to unify the approach, but if I remember correctly there were some trouble in detecting the actual value in each scenario. It is possible to create broken cell values by editing and changing data types, but I have the same problem in GDST validation so I should find a way to do this in a clean way.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/SubHeaderBuilder.java
Rikkola: I'll try with a more complex test file.

jomarko:
Screenshot from 2020-05-19 10-48-07
Screenshot from 2020-05-19 10-47-43


@Rikkola the screenshots shows that the table header start with new line. It means when user open downloaded XLS it seems like no header but if he resize row width it is there.

The issue is not resizing needed but I think user should see immediately see at least starting character of the header. Would it be possible?

Rikkola:
@jomarko I think it should be easy to add new rows for the cell.

jomarko:
@Rikkola sorry, confused, my comment https://github.com/kiegroup/drools-wb/pull/1365#issuecomment-630695364 is not about adding new row, but to trim white characters form the beginning of the header cells

Rikkola:
@jomarko Sure we can do that.

jomarko:
@Rikkola please ping me when changes are incorporated

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/DataBuilder.java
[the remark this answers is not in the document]
Rikkola: There were two methods due to the fact that the BRL column with `@` template manages the values in the GDST grid a bit differently than the Guided Rule variables. I did my best to merge them.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/SubHeaderBuilder.java
Rikkola: It did need this. I improved the test file to cover this and added the feature.

Rikkola:
@jomarko Ok. I think I covered everything and this is ready for review.

Rikkola:
Failing due to some dublicate classes in -kogito.
Jenkins please retest this.

jomarko:
Manual review
I found issue during converting, for more details see [comment in jira ](https://issues.redhat.com/browse/DROOLS-5210?focusedCommentId=14113959&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14113959)

Rikkola:
Looks like I accidentally pushed this. Still improving import for the RHS and cleaning up.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/DataBuilder.java
```diff
@@ -99,39 +105,118 @@ public void build() {
...
+                                    && cell.getBooleanValue()) {
+                                xlsRow.createCell(targetColumnIndex)
+                                        .setCellValue("X");
+                            }
+                        } else if (dtable.getExpandedColumns().get(sourceColumnIndex) instanceof ActionRetractFactCol52) {
+
+                            if (!StringUtils.isEmpty(cell.getStringValue())) {
+                                xlsRow.createCell(targetColumnIndex)
+                                        .setCellValue(cell.getStringValue());
+                            }
```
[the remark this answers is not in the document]
Rikkola: Not sure if I edited out of this or not, but if I'm not sure what you mean.

Rikkola:
@jomarko Ok. Pushed the latest with additional test coverage.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnSubHeaderBuilderDefault.java
```diff
@@ -0,0 +1,169 @@
...
+                    builder.append("\n");
+                }
+            }
+            if (!found && Objects.equals(row.trim(), from)) {
+                found = true;
+            }
+        }
+
+        return builder.toString();
+    }
```
jomarko, line 163: This can be rewritten as:
```java
private String subString(final String marshal,
                             final String from,
                             final String to) {

        Pattern regex = Pattern.compile("\\s*" + from + "\\s*(.*)" + to, Pattern.DOTALL);
        Matcher regexMatcher = regex.matcher(marshal);
        if (regexMatcher.find()) {
            return regexMatcher.group(1);
        } else {
            throw new IllegalStateException(String.format("No substring between '%s' and '%s' found",
                                                          from,
                                                          to));
        }
    }
```

jomarko:
### Manual review 01
We miss the acceptance criteria of original requirement:
"Explicitly mention HIT policies are not supported as a message in conversion results dialogue"

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnDataBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.poi.ss.usermodel.Row;
+import org.drools.workbench.models.guided.dtable.shared.model.BRLActionColumn;
+import org.drools.workbench.models.guided.dtable.shared.model.BRLColumn;
+import org.drools.workbench.models.guided.dtable.shared.model.BRLConditionColumn;
+import org.drools.workbench.models.guided.dtable.shared.model.DTCellValue52;
+import org.drools.workbench.screens.guided.dtable.backend.server.conversion.util.BRLColumnContext;
+
+import static org.drools.workbench.screens.guided.dtable.backend.server.conversion.util.Util.canSplit;
+
+public interface BRLColumnDataBuilder {
```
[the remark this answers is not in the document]
Rikkola: I added some notes about the functionality.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnSubHeaderBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.drools.workbench.screens.guided.dtable.backend.server.conversion;
+
+import org.drools.workbench.models.guided.dtable.shared.model.BRLActionColumn;
+import org.drools.workbench.models.guided.dtable.shared.model.BRLConditionColumn;
+
+public interface BRLColumnSubHeaderBuilder {
```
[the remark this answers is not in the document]
Rikkola: I moved this info into the column context. Makes sense to keep it there.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnSubHeaderBuilderByPatterns.java
```diff
@@ -0,0 +1,134 @@
...
+    public BRLColumnSubHeaderBuilderByPatterns(final SubHeaderBuilder subHeaderBuilder,
+                                               final ColumnContext columnContext,
+                                               final GuidedDecisionTable52 dtable) {
+        this.subHeaderBuilder = subHeaderBuilder;
+        this.columnContext = columnContext;
+        this.dtable = dtable;
+    }
+
+    @Override
+    public void brlActions(final BRLActionColumn brlColumn) {
```
vpellegrino, line 56: It seems that this method is doing too many things. Could you please modularize it in multiple methods with proper names?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnSubHeaderBuilderDefault.java
```diff
@@ -0,0 +1,178 @@
...
+    public BRLColumnSubHeaderBuilderDefault(final SubHeaderBuilder subHeaderBuilder,
+                                            final ColumnContext columnContext,
+                                            final GuidedDecisionTable52 dtable) {
+        this.subHeaderBuilder = subHeaderBuilder;
+        this.columnContext = columnContext;
+        this.dtable = dtable;
+    }
+
+    @Override
+    public void brlActions(final BRLActionColumn brlColumn) {
```
vpellegrino, line 59: It seems that this method is doing too many things. Could you please modularize it in multiple methods with proper names?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnSubHeaderBuilderDefault.java
```diff
@@ -0,0 +1,178 @@
...
+    private String getBoundName(final ActionInsertFact insertFact) {
+        if (hasContent(insertFact.getBoundName())) {
+            return insertFact.getBoundName();
+        } else {
+            return columnContext.getNextFreeColumnFactName();
+        }
+    }
+
+    @Override
+    public void brlConditions(final BRLConditionColumn brlColumn) {
```
vpellegrino, line 100: Proper name can help to understand what this method is doing

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnSubHeaderBuilderDefault.java
```diff
@@ -0,0 +1,178 @@
...
+        for (IPattern iPattern : brlColumn.getDefinition()) {
+            if (iPattern instanceof FactPattern) {
+                columnContext.addBoundName(((FactPattern) iPattern).getBoundName());
+            }
+        }
+
+        final GuidedDecisionTable52 dt = new GuidedDecisionTable52();
+        dt.getConditions().add(brlColumn);
+        final ArrayList<DTCellValue52> list = new ArrayList<>();
+        list.add(new DTCellValue52(1));
```
vpellegrino, line 111: Is 1 a specific value? Maybe, using a constant can help to understand its purpose

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/SubHeaderBuilder.java
```diff
@@ -101,37 +132,52 @@ public void build() {
...
                 throw new IllegalArgumentException("TODO REMOTE THIS");
             }
-            targetColumnIndex++;
+            upTargetIndex();
         }
     }
 
-    private void makeAction(final int sourceColumnIndex,
-                            final BaseColumn baseColumn) {
+    public void upTargetIndex() {
```
vpellegrino, line 139: The name of this method it does not suggest me what it is really doing.
What do you think about `incrementTargetIndex` or `incrementTargetColumnIndex`?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/util/Util.java
```diff
@@ -0,0 +1,75 @@
...
+import org.drools.workbench.models.datamodel.rule.FactPattern;
+import org.drools.workbench.models.datamodel.rule.FieldConstraint;
+import org.drools.workbench.models.datamodel.rule.FieldNatureType;
+import org.drools.workbench.models.datamodel.rule.IAction;
+import org.drools.workbench.models.datamodel.rule.IPattern;
+import org.drools.workbench.models.datamodel.rule.SingleFieldConstraint;
+import org.drools.workbench.models.guided.dtable.shared.model.BRLActionColumn;
+import org.drools.workbench.models.guided.dtable.shared.model.BRLConditionColumn;
+
+public class Util {
```
vpellegrino, line 30: If I understand correctly, the method inside this class are used as utilities for `BRLActionColumn` and `BRLConditionColumn`, right? Why not moving those methods directly in such classes? Otherwise, we should use a different (and less generic) name.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/util/Util.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public static boolean hasContent(final String value) {
```
vpellegrino, line 32: This method can be removed and `StringUtils.isNotEmpty` could be used: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isNotEmpty-java.lang.CharSequence-

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/util/Util.java
```diff
@@ -0,0 +1,75 @@
...
+                    }
+                }
+            } else if (!(iAction instanceof ActionFieldList)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public static boolean canSplit(final BRLConditionColumn brlColumn) {
```
vpellegrino, line 52: This method has a lot of return statements. Maybe intermediate boolean variables can help to combinate some of them and make the code more readable. WDYT?

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnDataBuilderDefault.java
```diff
@@ -0,0 +1,123 @@
...
+        final int amountOfChildColumns = brlColumn.getChildColumns().size();
+
+        final StringBuilder result = new StringBuilder();
+        boolean hasContent = false;
+        for (int i = 0; i < amountOfChildColumns; i++) {
+
+            final String value = dataRowBuilder.getValue(row,
+                                                         dataRowBuilder.getSourceColumnIndex());
+            if (value != null && missingVariable) {
+                if (Objects.equals("true", value.toLowerCase())) {
```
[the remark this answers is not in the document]
Rikkola: What do you know, it was available.

Rikkola:
TODO for myself. Need to double check the `update()` is working.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/BRLColumnDataBuilderByPatterns.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public void build(final BRLActionColumn baseColumn,
+                      final List<DTCellValue52> row,
+                      final Row xlsRow) {
+        setCellValues(row, xlsRow, columnContext.getCols(baseColumn));
+        upSourceColumnIndex(baseColumn);
+    }
+
+    private void setCellValues(final List<DTCellValue52> row,
+                               final Row xlsRow,
```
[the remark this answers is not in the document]
Rikkola: I tried to make some sense into these set* methods.

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/main/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/util/Util.java
Rikkola: The methods are used in several locations. I renamed the `Util` class and refactored to a more rule like format for the reason why we return `false` in each situation.

Rikkola:
@jomarko @vpellegrino The latest changes are in a separate commit. I still need to double check if the `update()` needs to be added as a column or not. It would add a new column like the `insert()` now does ( for the RHS ).

drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-backend/src/test/java/org/drools/workbench/screens/guided/dtable/backend/server/conversion/XLSBuilderHitPolicyTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Test
+    public void warning() throws Exception {
+        final String xml = loadResource(XLSBuilderAttributesNegateTest.class.getResourceAsStream("HitPolicy.gdst"));
+
+        final GuidedDecisionTable52 dtable = GuidedDTXMLPersistence.getInstance().unmarshal(xml);
+
+        final XLSBuilder.BuildResult buildResult = new XLSBuilder(dtable, makeDMO()).build();
+        assertFalse(buildResult.getConversionResult().isConverted());
+        assertEquals("Migrating Hit Policies is supported.", buildResult.getConversionResult().getMessage());
```
jomarko, line 37: will need update according a comment above.

Rikkola:
Update done. Fixes done by Jozef's comments and based on Sonar.

jomarko:
@Rikkola I am sorry, we can not merge this PR. See my added test https://github.com/Rikkola/drools-wb/pull/10.

There is issue if we try to set single string field via BRL column. We produce header like `setField( $param )` but in data we do not have quotes. We should either fix data and add quotes or fix header and add quotes around `$param`.

The issue is in one of these places:
-  SubHeaderBuilder#addSetMethod - where wo do not assume data type of column
- CombinedValueProvider where we inappropriately add and trim quotes
- DataBuilder#getValue - we somtimes call with adding quotes to string types sometimes not

jomarko:
jenkins please retest this

Rikkola:
This will need an update before merging, but the update depends on https://github.com/kiegroup/drools-wb/pull/1370 getting merged.

jomarko:
@Rikkola is that order the better? I mean this is already approved. If there will be massive rebase after #1370 I think this will need a re*review. Just asking, if this way is needed, then ok.

Rikkola:
@jomarko Either order is fine by me. For this PR the change will be just updating the test coordinates and the starting source index step count in the Builders. I expect this to be easier that way, but I could be wrong.

jomarko:
ok, take this way then

Rikkola:
@jomarko @vpellegrino I rebased this PR, the latest commit in the PR shows the one line I had to change on top of the rebase to get this working. The test data comes from GDST files and the tests test the XLS end result is filled correctly. I think we are good. I also tested locally.

Rikkola:
jenkins please retest now

Rikkola:
jenkins please retest this

Rikkola:
Worked locally both on CLI and in Idea

Rikkola:
Jenkins please retest this.

jomarko:
@Rikkola the fail on jenkins is due to different timezone on jenkins machine, I updated the existing test and added one more before merge https://github.com/Rikkola/drools-wb/pull/13

Rikkola:
@jomarko Thank you for the PR.
Jenkins please retest this.

Rikkola:
Same test still failing. Need to figure out what is going on here.

Rikkola:
@kiegroup/gatekeepers Can I get a merge please.

PR #1366, opened 2020-05-14
KOGITO-2189: Columns widths are not correctly saved in Scesim Editor KOGITO

yesamer:
@danielezonca @dupliaka can you please review and test it?

I noticed that `columnWidth` changes were not saved in VSCode Kogito Editor. Scope of this PR is to let this mechanism works.

`wrappedSave` method was removed because not called anywhere.

https://issues.redhat.com/browse/KOGITO-2189

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-businesscentral-client/src/main/java/org/drools/workbench/screens/scenariosimulation/businesscentral/client/editor/ScenarioSimulationEditorBusinessCentralWrapper.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-businesscentral-client/src/main/java/org/drools/workbench/screens/scenariosimulation/businesscentral/client/editor/ScenarioSimulationEditorBusinessCentralWrapper.java
@@ -241,11 +241,6 @@ public class ScenarioSimulationEditorBusinessCentralWrapper extends KieEditor<Sc
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


yesamer:
jenkins retest this please

yesamer:
jenkins retest this please

dupliaka:
@yesamer two questions:
-is that expected behavior: when I drag some of the bounder it does not stops following the mouse pointer even if I release the right mouse button
- is that ok that sometimes scroll is getting hidden and I could not move to the rightest columns
see https://issues.redhat.com/secure/attachment/12472633/scroll.webm

yesamer:
@dupliaka I think those are problems on appformer side

dupliaka:
> @dupliaka I think those are issues on appformer side

Are you sure about it? 
I checked 7.37 version of BC master and drag and drop is working there but after your PR that changes dramatically. Could you please provide me some details why do you thing so?

yesamer:
@dupliaka I meant appformer **Kogito** version. The widget which contains the grid is managed on appformer code. Please try to reproduce the same issues on **Kogito** VSCode without my changes, they should be present as well

dupliaka:
> @dupliaka I meant appformer **Kogito** version. The widget which contains the grid is managed on appformer code. Please try to reproduce the same issues on **Kogito** VSCode without my changes, they should be present as well
> 
> Anyway, I agree with your choice to open a new ticket for those issues, we can analyze it with the support of appformer team.

Doesn't matter, I updated all of the dependency, and it seems like it is finally works for me, thanks! 
Approved

PR #1384, opened 2020-06-05
DROOLS-5415: fixed beans.xml schemaLocation

mareknovotny:


Rikkola:
@jomarko ping just in case you are interested.

jomarko:
thank you, seem fine to me

PR #1386, opened 2020-06-10
DROOLS-4582: Cell selection doesn't span the entire cell if you Insert RightMost Column

yesamer:
@danielezonca @dupliaka Can you please review and test it?

After an `AppendColumn` or a `PrependColumn` command, the current selected header cell is re-selected to updated the full header cells selection.

Screencast-from-06-10-2020-04_10_48-PM

https://issues.redhat.com/browse/DROOLS-4582


dupliaka:
Cannot see the change on my local machine could you please check it locally if that changes have been applied, or share the link to the other module that has been changed?

https://issues.redhat.com/secure/attachment/12475061/span.webm

yesamer:
@dupliaka I just checked, it works for me, exactly how you can see on the video. Can you please restest? @jstastny-cz if you want, you can check it as well. Thanks!

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/models/AbstractScesimGridModel.java
```diff
@@ -603,6 +603,17 @@ public void selectRow(int rowIndex) {
         IntStream.range(0, columns).forEach(columnIndex -> selectCell(rowIndex, columnIndex));
     }
 
+    /**
+     * It selects the current selected header cells group
+     */
+    public void reselectCurrentHeaderCell() {
```
danielezonca, line 609: To be honest I don't think the fix should be done here. The header of the columns is merged so I expect to have the selection updated out of the box by grid code.
If this is not working the fix should be done directly on [uberfire-wires-core-grids](https://github.com/kiegroup/appformer/tree/master/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids).
Wdyt?

@manstis 
Do you have any advice?

yesamer: @danielezonca Ideally, it would be the best. The problem I see is that the current behavior of selecting spanned cells, is already defined in scesim module. If we choice to move the logic on uberfire-wires-core-grids, fine for me but it should be done moving all of the behavior, not only this PR.

yesamer:
jenkins retest this please

yesamer:
Jenkins please retest this.

manstis (approved):
LGTM.

Rikkola (approved):

yesamer:
@kiegroup/gatekeepers Can you please merge it?

PR #1390, opened 2020-06-25
KOGITO-2137 & KOGITO-2337: Integrate StateControlAPI in TestScenario

yesamer:
@danielezonca @jstastny-cz Can you please review and test it?
/cc @pefernan 
This PR enables StateControlAPI on TestScenario Editor in Kogito (VSCode channel)
Functionalities enabled are:
- If user modifies grid content without saving and tries to close the file, a popup will be shown to notify the file was not saved;
- Undo & Redo actions (Pressing CTRL+Z and CTRL+SHIFT+Z)

Screencast from 06-25-2020 02_59_11 PM

Related PR on `kie-wb-commons` repo: https://github.com/kiegroup/kie-wb-common/pull/3344

https://issues.redhat.com/browse/BAPL-1677
https://issues.redhat.com/browse/KOGITO-2137
https://issues.redhat.com/browse/KOGITO-2337


drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/pom.xml
```diff
@@ -130,7 +130,11 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
-    <!-- END TEST -->
+      <dependency>
+          <groupId>org.uberfire</groupId>
+          <artifactId>appformer-kogito-bridge</artifactId>
```
danielezonca, line 135: Scope test?
If not it is in the wrong dependency section :)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/producer/ScenarioSimulationKogitoProducer.java
```diff
@@ -0,0 +1,40 @@
...
+
+import javax.enterprise.context.Dependent;
+import javax.enterprise.inject.Specializes;
+import javax.inject.Inject;
+
+import org.drools.workbench.screens.scenariosimulation.client.producers.ScenarioSimulationProducer;
+import org.drools.workbench.screens.scenariosimulation.kogito.client.commands.ScenarioSimulationKogitoStateControlManager;
+
+
+@Specializes
```
danielezonca, line 26: Why this?
If now there is a command manager that works in Kogito only we should use interface to decouple and then provide proper bean based on environment.
Wdyt?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/domelements/ScenarioCellTextAreaDOMElement.java
```diff
@@ -115,7 +115,9 @@ protected void internalFlush(final String value) {
 
     @Override
     public void detach() {
-        flush(getValue());
         super.detach();
+        if (scenarioGridCell != null) {
+            scenarioGridCell.setEditingMode(false);
```
danielezonca, line 120: What is the reason of this change?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/ScenarioSimulationContext.java
```diff
@@ -251,6 +251,14 @@ public ScenarioExpressionCellTextAreaSingletonDOMElementFactory getScenarioExpre
         return getAbstractScesimGridModelByGridWidget(gridWidget).getScenarioExpressionCellTextAreaSingletonDOMElementFactory();
     }
 
+    public void setUndoButtonEnabledStatus(boolean enabled) {
```
danielezonca, line 254: Why these setters?

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/commands/ScenarioCommandRegistryManagerTest.java
```diff
@@ -0,0 +1,171 @@
...
+        doneCommandsRegistrySpy.register(abstractScenarioGridCommandMock);
+        doReturn(CommandResultBuilder.SUCCESS).when(scenarioCommandRegistryManagerSpy).commonUndoRedoOperation(any(), any(), anyBoolean());
+        doReturn(Optional.empty()).when(scenarioCommandRegistryManagerSpy).commonUndoRedoPreexecution(any(), any());
+        final CommandResult<ScenarioSimulationViolation> retrieved = scenarioCommandRegistryManagerSpy.undo(scenarioSimulationContextLocal);
+        assertEquals(CommandResult.Type.INFO, retrieved.getType());
+        verify(scenarioCommandRegistryManagerSpy, times(1)).commonUndoRedoOperation(eq(scenarioSimulationContextLocal), eq(abstractScenarioGridCommandMock), eq(true));
+        verify(scenarioCommandRegistryManagerSpy, times(1)).setUndoRedoButtonStatus(eq(scenarioSimulationContextLocal));
+    }
+
+    @Test //TODO Recheck
```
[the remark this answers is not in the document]
yesamer: @danielezonca Removed in the last commit.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/pom.xml
yesamer: @danielezonca Wrong section, moved

pefernan:
@yesamer you did it!

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/commands/ScenarioSimulationContext.java
yesamer: @danielezonca Because of this https://github.com/kiegroup/drools-wb/pull/1390/files#diff-62f87d136d61c90564bc41cf80af7a36L134 . I don't like direct field access :)

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/domelements/ScenarioCellTextAreaDOMElement.java
yesamer: @danielezonca Because I found a bug during my manual testing. Basically, when a cell value is edited, a double command is created (one calling `flush()` and the second one calling `detach()` . I aware the best way to manage it should be to create a new PR, but I preferred to include it in this PR in order to include it to DevPreview cut off (1th July).

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/producer/ScenarioSimulationKogitoProducer.java
yesamer: @danielezonca That was my first implementation (https://github.com/kiegroup/drools-wb/pull/1390/commits/a93ac60420e379262f5347102944a78179a3895e). What I don't like of that solution, is that it will result in a empty bean in case of BusinessCentral. The scope of this change is to define a bean which manages `StateControlAPI` mechanism, which is present in Kogito environment only. Therefore, I think that kind of solution should be used when either BusinessCentral and Kogito define a logic for that manager. In this case I'm sure this manager is required on Kogito only, so I want to define it in Kogito scope only.

yesamer:
> What about isDirty?
> I remember in BC version isDirty value is (was?) calculated using a hash function with the model while now in Kogito is based on command manager status if I have understood correctly.
> Is it possible to change BC code to use the same approach? Otherwise the risk is to have different behaviours.
> Wdyt?

This is my same understanding. As a evidence of that, consider that `isDirty()` method inside KogitoEditor is not implemented (it returns always the same int values), so I'm pretty sure different approaches are used.
Maybe @pefernan can give us more info about it

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/producer/ScenarioSimulationKogitoProducer.java
yesamer: @danielezonca Agreed changes applied, thank you!

pefernan:
@yesamer @danielezonca I've been away from this development for a while, but in VSCode isDirty was originally handled by VSCode itself. At the time I did this part, it was keeping a stack with all changes and when it was detecting an undo/redo it was adding/removing changes and notifying the extension to do his part of the work. So it was aware of the dirty state by checking that command stack.

yesamer:
@jstastny-cz Thank you for your review!
1. Not related to this PR. In my understanding, this is not working for any editor. Please use only keyboard shortcut (CTRL+Z, CTRL+SHIFT+Z ) to test UNDO / REDO.
2. Good catch. The problem here is the switch between tabs. That transition acts like a command, but it is not a real command and this broke the mechanism. A solution could be to merge the UNDO command and the tab switch in the same command (like excel do, for example)

jstastny-cz:

yesamer:
failing tests are not related to this PR @kiegroup/gatekeepers can you please merge it?

PR #1397, opened 2020-07-10
DROOLS-5317: Scenario Simulation shows misleading data type if DMN applies a constraint

yesamer:
@danielezonca @karreiro @Rikkola @jstastny-cz can you please test and review?
To handle the issue, I defined a nested DTO class `PropertyNameType` inside `FactModelTree` class.
The new class contains the `typeName` and `baseTypeName` string values. The first one represents the real Fact type, while `baseTypeName` is the base type to show into UI. 
`simpleProperties` now holds this new class as value: from `<String, String>` to `<String, PropertyNameType>`
Behind the scenes `typeName` is always used, eg. Performing the validation and launching the runner.

The change is all on that class, unfortunately it requires to change several other classes ant tests, which use the `simpleProperties` field.

Screenshot from 2020-07-13 16-20-02

https://issues.redhat.com/browse/DROOLS-5317
https://issues.redhat.com/browse/RHDM-1153


yesamer:
jenkins retest this please

yesamer:
jenkins retest this please

yesamer:
@all I'll add additonal tests

danielezonca:
@yesamer 
As far as I remember we rollbacked original impl of this ticket because it caused a validation error with most of existing scesim files. Am I correct?

yesamer:
@danielezonca you're right

danielezonca:
Ok and is it solves now?

yesamer:
@danielezonca The original issue doesn't have any validation problem. That one was introduced with my previous solution, because I substituted the `typedName` with `baseTypeName` in some conditions. Reverting it, we already solved the validation issue. 
This PR changes the label in test tools only, therefore I don't aspect changes on the current logic flow (validation included).

yesamer:
@karreiro I'm fine with tests :). Thx u

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/dmn/AbstractKogitoDMNService.java
```diff
@@ -534,13 +534,13 @@ protected FactModelTree createFactModelTreeForComposite(final Map<String, List<S
         if (!type.isComposite() && !isToBeManagedAsComposite(type)) {
             throw new IllegalStateException(WRONG_DMN_MESSAGE);
         }
-        Map<String, String> simpleFields = new HashMap<>();
-        FactModelTree toReturn = new FactModelTree(name, "", simpleFields, genericTypeInfoMap, fmType);
+        Map<String, FactModelTree.PropertyTypeName> simpleFields = new HashMap<>();
+        FactModelTree toReturn = new FactModelTree(name, "", simpleFields,  genericTypeInfoMap, fmType);
```
jstastny-cz, line 538: added extra space.

yesamer:
@kiegroup/gatekeepers can you please merge it?

PR #1419, opened 2020-09-16
BXMSPROD-990 drools-wb-scenario-simulation-editor-kogito-webapp-common artifact added

Ginxo:
**Thank you for submitting this pull request**

**JIRA**: https://issues.redhat.com/browse/BXMSPROD-990

<pre>
How to retest a PR or trigger a specific build:

</pre>

### DIFF
--- a/drools-wb-webapp/pom.xml
+++ b/drools-wb-webapp/pom.xml
@@ -922,24 +922,6 @@
       <artifactId>drools-scenario-simulation-api</artifactId>
     </dependency>
 
-    <dependency>
-      <groupId>org.drools</groupId>
-      <artifactId>drools-wb-scenario-simulation-editor-kogito-webapp-common</artifactId>
-      <scope>provided</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.drools</groupId>
-      <artifactId>drools-wb-scenario-simulation-editor-kogito-client</artifactId>
-      <scope>provided</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.kie.workbench</groupId>
-      <artifactId>kie-wb-common-kogito-webapp-base</artifactId>
-      <scope>provided</scope>
-    </dependency>
-
     <!-- Stunner -->
     <dependency>
       <groupId>org.kie.workbench.stunner</groupId>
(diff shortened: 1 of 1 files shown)


Ginxo:
let's for GA `Build Chain / Maven Build (8) (pull_request)` execution but locally works

yesamer:
@Ginxo @mareknovotny Most probably this change can fix the issue, but I think add a `kogito` dependency in drools-webapp should be avoided.  drools-webapp doesn't need any kogito module, so we should try to understand why during drools-webapp compilation that module is compiled.

danielezonca:
I agree with @yesamer , these dependencies should not be needed to compile. @Ginxo are you able to reproduce the error locally?

Ginxo:
> I agree with @yesamer , these dependencies should not be needed to compile. @Ginxo are you able to reproduce the error locally?

Yes I can.
I agree with you guys, can we merge while we find the real reason?

drools-wb-webapp/pom.xml
```diff
@@ -922,6 +922,24 @@
       <artifactId>drools-scenario-simulation-api</artifactId>
     </dependency>
 
+    <dependency>
```
yesamer, line 925: @Ginxo Can you please move the dependencies on block which starts in line 1096? Here the are others kogito dependencies which are temporary (probably for the same reason you need to add these new ones?) and planned to be removed with https://issues.jboss.org/browse/KOGITO-459 ? My suggestion is to link all these kogito dependencies in other to be managed (and removed) togheter.

yesamer (commented):
@Ginxo Considering this issue blocks the builds and we haven't any clue of the reason, fine for me to proceed. However, I underline this should be considered a temporary solution, in the same way we did for other kogito dependencies described here. I understand the difficulties to find the root cause, I tried to analyze it but I didn't have any clue at the moment, so please continue your analysis after the merge, I can support you if you find a clue.

drools-wb-webapp/pom.xml
yesamer: @Ginxo Or just put the same comment in this place, your choice :)

Ginxo: @yesamer moved, thanks

yesamer (approved):
Thank you @Ginxo.
 In my opinion the reason we need to add these new dependencies and the reason we added `kie-wb-common-stunner-kogito-api` (et others) it's the same.

danielezonca (approved):
Approving because it is blocking CI setup

mareknovotny (approved):
approving and @Ginxo will create a new jira for followup to fix this "unnecessary" linkage of kogito modules into Business Central components

PR #1446, opened 2020-11-23
DROOLS-5844: Errors when executing models using imported inputs and/or decisions nodes - BC

yesamer:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5844

**referenced Pull Requests**: 

* https://github.com/kiegroup/drools/pull/3460
* https://github.com/kiegroup/drools-wb/pull/1446

These changes affect BUSINESS CENTRAL only. Kogito will be managed in https://issues.redhat.com/browse/KOGITO-4765

A DMN file can refer to another "external" DMN file. The current version of Scesim doesn't support this feature, in particular, is not able to manage Decisions or Inputs inherited from the Imported external DMN file.
An example of a DMN file that uses an Input node inherited from an external file (Traffic Violation):
Screenshot from 2021-03-23 15-16-16

With its related scesim file, which now supports it with the introduced changes in this PR
Screenshot from 2021-03-23 15-15-14

An example of a DMN file that uses a Decision node inherited from an external file (Traffic Violation):
Screenshot from 2021-03-23 15-14-06
and related scesim generated file
Screenshot from 2021-03-23 15-13-22

<details>
<summary>
</summary>

</details>


yesamer:
jenkins run fdb

yesamer:
jenkins retest this please

karreiro (approved):
Thank you, @yesamer.

yesamer:
jenkins run fdb

jomarko (commented):
# Manual Review Questions/Comments

## 01
The Test Scenario offers all nodes from included model.
Imagine we are going to test model
Screenshot from 2021-03-30 11-21-37

Is it expected Test Scenario offers also other nodes from cc?
Screenshot from 2021-03-30 11-22-47

## 02
User do not see all fields of structures.
This is how cc.Customer is displayed in Test Scenario
Screenshot from 2021-03-30 11-25-55
And this is how the real definition looks like
Screenshot from 2021-03-30 11-27-38

As no expert, I understand it is maybe not possible to parse fields cc.Customer. However we should at least guide user how to put a value in Test Scenario designer in such case. What do you think? I expect it can be done somehow via context? `{}` or an expression `#` ?

## 03
Test Scenario does not offer Decision Services Nodes. Imagine Model:
Screenshot from 2021-03-30 11-44-16

If such is included into a DMN model, and the DMN model is tested by Test Scenario, then just 'Fine Points' and 'Decrease Points' are offered. 'Reclassified Points' are missing. Is this intentional, will we address it separately?

## 04
Expression can not be changed to value
Imagine context like this
Screenshot from 2021-03-30 11-50-30

Then try to change expression to value
Screenshot from 2021-03-30 11-50-41

It will not change after you click 'Insert Data Object'. For non imported nodes that works without an issue.

## 05
In code I noticed we assume more '.' in fact identifier as issue. However I am not sure if that is compatible with DMN tooling. Because that allows to include another DMN model and use prefix like demonstrated below.
Screenshot from 2021-03-30 11-55-27

Now it is question if DMN standard restrict somehow names, that can be used as prefix of included model. If yous. We should address this in DMN editor side. If not, I think we should address this on Test Scenario side, probably separately from this PR. hat is your opinion?

yesamer:
@jomarko 

1. Yes, that's expected. Currently we shows all avialble node in Test Tolls, even if not actually used in logic.  The same is true for RULE test scenario, we load all DataObjects avialable in the project, even if not used into project.
2. Need to check 
3. Yes, that is not currently supported. I agree to manage it separately.
4. I agree, that is a bug. I think is not a regression introduced with this PR, so I propose to manage it in a seperate JIRA.
5. That is a good point. I think we need to discuss it together with @karreiro. It is totally legal in DMN to have Types with multiple dots as a Name and this doesn't generate issues on their side I can think how to translate it in scesim only. Thit will require a more complex logic indeed.

jomarko:
@yesamer ok
01 - no action needed then
02 - waiting for your investigation
03 - reported separately
04 - reported separately
05 - waiting for @karreiro advice

karreiro:
@jomarko, @yesamer talked to me in the chat about the 05, and more than one `.` is allowed indeed. I believe he's working in the fix.

yesamer:
@jomarko @karreiro I confirm, I'm working to manage that use case. Thank you, Jozef for raising that.

jomarko:
@yesamer thank you, waiting for your green flag to restart my review. Just tell me then if you incorporated both 02 and 05, or just one of them.

yesamer:
@jomarko @karreiro 

 Ready to be reviewed again. In details:

2: The reason is `Customer` in `checkCustomer.dmn` has `undefined` type. Please simpy assign it to `tPerson` and everthing will work as expected. 
Screenshot from 2021-04-02 10-04-59
4: I added a fix for that issue as well. Tested it for an imported node with multiple dots as your example. 
5: Now multiple dots are supported. I changed the way to retrieve the prefix. In the previous version, I checked if the fact name contains a ".", retrieving the name before the ".". Now, I created a dedicated attribute in scesim file, which defines the prefix
`<factIdentifier importPrefix="imp">`
With this solution, we can support fact with multiple dots.

yesamer:
jenkins run fdb

jomarko (approved):
A recap:
01 - no action needed then
02 - works if the data types are set correctly in the original model
03 - reported separately
04 - closed, addressed as part of this PR
05 - aliases with multiple dots works

thank you @yesamer

karreiro (approved):
Thank you, @yesamer!

PR #1447, opened 2020-12-02
DROOLS-5876: Display actual test results instead of a generic message

yesamer:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5876  and https://issues.redhat.com/browse/DROOLS-5402

Client-side PR contains minor changes only on UI (popover is larger and error message shorter).
Server-side PR introduced a new DTO, `ExpressionEvaluatorResult` which holds detailed information related to the result of an Expression Evaluation. In particular, it contains the original boolean which represents the computation result (true = successful, false = failure), and adds the wrong value and the path to reach it. 
Please note, currently is not possible to provide a correct value to the user, because a List evaluation doesn't consider the list order. 

Examples:
- DMN List
Screenshot from 2021-02-03 10-03-02

- Rule Map
Screenshot from 2021-02-03 10-26-14

- Rule List<String> (Simple list type)
Screenshot from 2021-02-03 10-28-08

- Rule List<Author> 
Screenshot from 2021-02-03 10-30-35

**referenced Pull Requests**: 

* https://github.com/kiegroup/drools-wb/pull/1447
* https://github.com/kiegroup/drools/pull/3276

<details>
<summary>
</summary>

</details>


yesamer:
jenkins retest this please

yesamer:
jenkins retest this please

yesamer:
jenkins run fdb

yesamer:
jenkins run fdb

yesamer:
jenkins run fdb

Rikkola (approved):

yesamer:
jenkins run fdb

yesamer:
jenkins run fdb

yesamer:
jenkins run fdb

yesamer:
@jstastny-cz @Rikkola @karreiro 
Screenshot from 2021-02-17 16-08-14
Screenshot from 2021-02-17 16-06-35
Screenshot from 2021-02-17 16-05-18
Snapshots of the final version. Can you please finalize the review? Thank you!

yesamer:
jenkins run fdb

jstastny-cz:
Does not seem to work correctly. Reporting a different field than the problematic one.
Given
Screenshot from 2021-02-23 10-45-09
Expect
Screenshot from 2021-02-23 10-45-20
Error
Screenshot from 2021-02-23 10-45-28

jstastny-cz:
More detailed description:
Given:
Screenshot from 2021-02-23 10-55-51
Expect:
Screenshot from 2021-02-23 10-55-40
Error:
Screenshot from 2021-02-23 10-55-59

yesamer:
@jstastny-cz To explain that behavior I need to describe how the search logic works in this case.
As I previously said, the algorithm doesn't compare the items considering the list sorting.
The algorithm takes the first EXPECT item, which in this case contains `age = 18` and `name = name2`
Said so, the first EXPECT item will be compared with the first GIVEN item `age = 25` and `name = name2`, which doesn't match.
Then, it will be compared with the second GIVEN item `age = 18` and `name = name1` which don't match as well.
That second GIVEN item represents the last item of the GIVEN list, so an error message will be produced here and it will consider this last element and the related cause (which is name attribute is different). For this reason, the message correctly reports that the first EXPECT item is wrong, basically because it's not possible to match the same item in the GIVEN list, and as a suggestion, we provide the last not matching field in the last GIVEN list element. (the name).
The reason for this behavior is exactly the same we discussed previously: the comparison logic doesn't consider the ordering of the list. Introducing that constraint will give us the opportunity to provide the precise wrong value and the precise suggestion to give to the user.

jstastny-cz:
So it reports that the item no.1 in the expect list has not found a match. Great, but the hint there is misleading as you can see.
We should maybe just report the items that were not matched - and either print them fully, or just provide the item locator (offset or key into a map).
Cause the error hint we see above is wrong given the context - it's a list, so all already matched items should be excluded from comparisons with the rest.

yesamer:
@jstastny-cz Think changing the current comparison algorithm is not the best option: In our examples, we know the correct expecting order because defined in GIVEN columns, but if Collections are generated as a result DRL o DMN task, is not possible for the user to determine the expected order. 
Therefore, the only feasible change here is to modify again the message. 
Providing the full item description seems useless to me: it's just copying and pasting what the user can see opening the Collection Editor. The key information is: which items have failed. 
For this reason, I think the only way we have is to show only the first row of the label (eg. `Failed Item # 1 `). 
On the other side, this means providing a poor level of info to the user. 
Not easy to choose :)

yesamer:
@jstastny-cz Referring to your example, what do you expect to see as a correct message?

jstastny-cz:
I would expect to see
```
Failed in : Item #1
Field
* author
  * age
Value: 18
```

Reasoning behind is, that Item 2 is equal in given and expect, thus Item 1 in expect should not be compared to Item 2 in given. But this would probably require complex processing.

yesamer:
@jstastny-cz Extending the algorithm to exclude matching items it's possible, with medium-high effort. 
But even implement it I'm not sure it totally solves our problem. Let's assume the algorithm excludes the matching items. In your case, it will work well, because the error will point to the single not-matching item. 
Consider instead the case there are multiple not-matching cases. In this case, I need to choose to which element the error should point, and this means to have again the same misleading situation you reported. 
For this reason, I would exclude this solution - It will only partially solve the issue with a medium-high effort required.
The easiest solution is to show only info related to the wrong item: 
> Failed in : Item #1
I'm trying to figure out if the are valuable alternatives.

jstastny-cz:
@yesamer I completely agree. I mean, it's already a great help to see which exact item from expect has not matched, so it's not like we were not improving the experience at all.

yesamer:
jenkins run fdb

yesamer:
> @yesamer I completely agree. I mean, it's already a great help to see which exact item from expect has not matched, so it's not like we were not improving the experience at all.

I think it's our best choice currently. I applied the change. Lists of complex objects will show only the wrong item indication. Can you please continue your testing analysis? Thanks.

yesamer:
jenkins run fdb

yesamer:
jenkins run fdb

yesamer:
jenkins run fdb

karreiro (approved):
Thank you, @yesamer LGTM!

yesamer:
@jstastny-cz Can you please review it? Thanks!

jstastny-cz (approved):
LGTM, verified functionality of maps and lists. Lists are reporting just item offset information, maps report problematic key + information of wrong value.

yesamer:
@karreiro Can you please merge it?

PR #1455, opened 2021-01-15
DROOLS-5230: Managing a List of structures in Test Scenario editor produces an Exception

yesamer:
The issue occurs when a "user defined" java class is used as type parameter of a List eg. `List<Clazz>` and that class contains a field which is a "user defined" java class as well. Screenshot show an example:

Screenshot from 2021-01-20 10-02-49

Author
+ books (List<Book>)
   . isAvailable (boolean)
   . name (string) 
   . ....
   + **author (Author)**    _<----  Issue occurs with this nested class field_
       . name
       . ...

With current version there are several issues:
- Not possible to correctly delete / modify a list item 
- the nested **author (Author)** reference is THE SAME for all items (eg. try to create multiple items of the list. If you try to change a field of Author, this will be propagated to all items!)

**JIRA**: https://issues.redhat.com/browse/DROOLS-5230

<details>
<summary>
</summary>

</details>

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/collectioneditor/ElementView.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/collectioneditor/ElementView.java
@@ -106,6 +106,8 @@ public interface ElementView<T extends ElementView.Presenter> extends HasPresent
      */
     String getItemId();
 
+    void setItemSeparatorText(String itemSeparatorText);
+
     /**
      * @return the <code>LIElement</code> containing all the item properties
      */
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/collectioneditor/ElementViewImpl.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/main/java/org/drools/workbench/screens/scenariosimulation/client/collectioneditor/ElementViewImpl.java
@@ -89,6 +89,11 @@ public abstract class ElementViewImpl<T extends ElementView.Presenter> implement
     }
 
     @Override
+    public void setItemSeparatorText(String itemSeparatorText) {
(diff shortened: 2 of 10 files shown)


yesamer:
jenkins retest this please

yesamer:
jenkins run fdb

Rikkola (approved):

jstastny-cz:
I have `Library -> books:List<Book>` and `Book->author:Person` getting unexpected error when inserted fact Library.books and using Collection editor adding into the input field for Person->name.

error details:
```
Uncaught exception: Exception caught: (TypeError) : Cannot read property 'Vc' of undefined Caused by: (TypeError) : Cannot read property 'Vc' of undefined
```

Screenshot from 2021-01-26 16-06-26

yesamer:
jenkins run fdb

yesamer:
@jstastny-cz Thank you for reporting. I fixed the wrong behavior.

yesamer:
jenkins run fdb

yesamer:
jenkins run fdb

jstastny-cz:
I've been able to test with the artifacts from the FDB - the build failed, but the artifacts were there (probably failed at later stage). Not blocked by the failure in testing.

yesamer:
@jstastny-cz Ok, I launched it again because I merged the PR with master (now with version `7.50`). But you can test with previously generated artifacts.

jstastny-cz:
I just want to clarify one thing, consider following case (ignoring some unimportant things in the domain):
```
public class Library {
    private List<Book> books;
}
public class Book {
    private String title;
    private Author author;
}
public class Author {
    private String name;
    private Classification classification;
}
public class Classification {
    private String era;
}
```
i.e. adding one more level to the depth of custom object navigation.
Screenshot from 2021-01-28 17-17-35


I can't fill in Classification for the Author then, the field is not there - is it because we support just limited level of recursion now?

So I would like to insert fact Library->books. I would expect I would add Book instances. I can see author field there, it has all fields shown as input fields BUT the field `classification`.

yesamer:
@jstastny-cz Yes, these are two known limitation of Collection Editor. 

1. For the first limitation, when we created the Collection Editor we decided to support only 1 level of nesting object. This is because it's wasn't easy to manage a potential `n` levels of nesting inside a popup. A ticket which track the intention to analyze this limitation and found a better solution in order to remove this limitation is https://issues.redhat.com/browse/DROOLS-4290;
2. About this, basically we don't support nested collection. My intention is to implement it with https://issues.redhat.com/browse/BAPL-1841.

yesamer:
jenkins retest this please

yesamer:
jenkins run fdb

jstastny-cz (approved):
LGTM, thanks

PR #1458, opened 2021-01-26
[DROOLS-5964] 

gitgabrio:
@danielezonca @yesamer  @jiripetrlik

see https://issues.redhat.com/browse/DROOLS-5964

Minimal required modifications to xsd/converters/tests.

Depends on https://github.com/kiegroup/drools/pull/3344


gitgabrio:
@danielezonca @yesamer 
Created as draft because editor modification is partial.
Please let me know if you prefer to keep as draft until someone may complete the client-side part.

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-client/src/test/java/org/drools/workbench/screens/scenariosimulation/client/handlers/SourceTypeSelectorTest.java
```diff
@@ -78,7 +78,7 @@ public void validateValidDMN() {
     public void addRadioButtons() {
         reset(uploadWidgetMock);
         sourceTypeSelector.addRadioButtons();
-        assertEquals(2, sourceTypeSelector.radioButtonList.size());
+        assertEquals(3, sourceTypeSelector.radioButtonList.size());
```
jomarko, line 81: @gitgabrio according this change, do I understand correctly from now on TestScenario asset will serve for testing of:
- DRL rules- old
- DMN models - old
- PMML models - **new**

?

gitgabrio: @jomarko 
yes, that's the plan. The backend is ready, unfortunately frontend would need other modifications that no-one can do right now, so both PRs are on "stand-by"

PR #1522, opened 2021-10-14
GWT:run missing dependencies

ederign:
Fixing gwt:run


Rikkola:
@ederign Thank you

jomarko:
what was the exception thrown @ederign please? asking because I had troubles to start also webapps in kie-wb-common-dmn

